PR #71, opened 2022-10-09
Add GitHub Actions CI job

### LINKED ISSUE #68: CI pipeline for PRs (formatting, checks, etc)
I want to add a simple CI pipeline that effectively mimics the pre-commit hook:

- run cargo checks (clippy, fmt, etc)
- run tsc checks on `common`

In the future I'll add the tests to this pipeline, but since those aren't done / functional that will have to wait. I am very unfamiliar with CI pipelines with GitHub, but I want to have some sort of cache layer so CI minutes are not wasted (I am not rich lol)

Celeo:
This setup will run on every push to your `main` branch. The two jobs (one each for the Rust and TypeScript code) run in parallel, and both make use of caching to speed up their runs, as requested.

I've commend-out the step to run clippy, as it's reporting warnings that fail the job. If you'd rather the violation report not cause the job to fail, let me know.

I explicitly use the TS type check commands in the two directories I see `tsconfig.json` files rather than running `pnpm run checks` at the top level, as that'd require the Rust setup in both jobs. That'd work, but I'd recommend collapsing the two jobs into one and forgoing the concurrency in favor of command simplicity.

Example run: https://github.com/Celeo/stump/actions/runs/3212770920

Fixes #68

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,9 +26,9 @@ jobs:
       with:
         path: target
         key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('**/Cargo.lock') }}
-    - name: Install clippy
-      run: rustup component add clippy
     # run checks
+    - name: Run cargo check
+      run: cargo check
     - name: Run clippy
       run: cargo clippy
     - name: Run formatter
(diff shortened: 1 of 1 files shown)


aaronleopold:
Hey, thanks so much! Normally, I ask for PRs to go into `develop` ([see here](https://github.com/aaronleopold/stump/blob/main/CONTRIBUTING.md)) but in this instance it would probably be better to go into `gh-actions`.

Would you mind rebasing with that branch to merge what you have with the [check-rust](https://github.com/aaronleopold/stump/blob/gh-actions/.github/workflows/ci.yaml#L10) job, and then your typescript job can just be placed as is underneath/above I assume. Then update the PR to point to `gh-actions` and we'll be golden 🙂

Celeo:
Will do!

PR #72, opened 2022-10-09
Update GHA CI build

### LINKED ISSUE #68: CI pipeline for PRs (formatting, checks, etc)

Celeo:
This updates the existing build on push. The two jobs (one each for the Rust and TypeScript code) run in parallel, and both make use of caching to speed up their runs, as requested.

I explicitly use the TS type check commands in the two directories I see `tsconfig.json` files rather than running `pnpm run checks` at the top level, as that'd require the Rust setup in both jobs. That'd work, but I'd recommend collapsing the two jobs into one and forgoing the concurrency in favor of command simplicity.

Fixes #68

re: https://github.com/aaronleopold/stump/pull/71

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,9 +11,9 @@ jobs:
     name: Run Rust checks
     runs-on: ubuntu-latest
     steps:
+      # setup
       - name: Checkout repository
         uses: actions/checkout@v3
-
       - name: Install Rust
         uses: actions-rs/toolchain@v1
         with:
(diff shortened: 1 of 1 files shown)


aaronleopold:
The rust part needs some TLC, but that's mostly because of the dependencies I think. Overall, this looks good to me, it provides a good foundation. Thanks again!

Celeo:
Great! You're very welcome; happy Hacktober!

PR #75, opened 2022-10-16
Added SystemD Service File

KrautByte:
Signed-off-by: Lars Erhardt <lars@krautbyte.com>

Added a small and simple SystemD Service file that people might use to run Stump as a Service.


aaronleopold (approved):

aaronleopold:
LGTM, thanks for the example service file!

PR #81, opened 2022-10-31
user api functions

berkingurcan:
- User password or name or both can be updated by id
- get user by id
- admin can delete user by id but user cannot for now

### DIFF
--- a/apps/server/src/routers/api/user.rs
+++ b/apps/server/src/routers/api/user.rs
@@ -1,19 +1,17 @@
-use std::hash;
-
 use axum::{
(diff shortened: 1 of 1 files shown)


aaronleopold:
Looks like the format checks failed on the rust side of things, but I'm okay with that for now. I know hacktoberfest ends soon in your timezone so I'll merge this before I finish my lunch break and then correct the format in develop afterwards. Thanks again @berkingurcan for your contributions!!

aaronleopold (approved):

Issue #70, opened 2022-10-06
GH Actions for releases


aaronleopold:
If I ever want to make a release, I'll need to have some sort of automated system for doing it. This would honestly be super useful for testing/nightly releases or something, as well. Like #68, I am not very knowledgable in this area.

In general, when a release is made the following needs to happen:
- `server` build
    - linux, macos, windows binaries (this can probably be done at once thanks to [this](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds)
- `web` build
- `desktop` build
    - linux, macos, windows (AppImage, dmg, msi - I think? Unsure what best is for Linux broad support)
- docker `server` build
    - arm, amd64

Some constraints that I have no idea how to go about yet:
- prisma client needs to be generated, cached and shared between rust actions
    - this might be tricky with docker? `include_str!` macros might mess things up but also 🤷
- rust dependencies should be cached (https://github.com/Swatinem/rust-cache)

f8thl3ss:
Hello! I'll give it a try :)

aaronleopold:
Just assigned you, thanks! If you have any questions please let me know 🙂

aaronleopold:
Hey there @f8thl3ss, just merged a few additional changes to `gh-actions` (see #72). Those involved changes to the PR checks, not the release actions for this issue but wanted to let you know in case there are any conflicts 🙂

f8thl3ss:
Thanks for letting me know! :) I've pulled the latest changes.

PR #83, opened 2022-11-09
Nightly build for web, desktop and server

f8thl3ss:
This PR should provide the building blocks needed for a release pipeline. Once a day, it will pull the code, build it and upload as an artifact.

### DIFF
--- a/.github/actions/build-server/action.yml
+++ b/.github/actions/build-server/action.yml
@@ -1,0 +1,33 @@
+name: 'Compile rust'
+description: 'Compile stump rust server'
+
+inputs:
+  platform:
+    description: 'runner plaform'
+    required: true
+
+runs:
+  using: composite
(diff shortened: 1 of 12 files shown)


aaronleopold (approved):
> Once a day, it will pull the code, build it and upload as an artifact.

That's honestly pretty cool, didn't know that could be a thing! For the time being, I'll disable that workflow, but I'm super excited to have it defined for future use at the very least.

The rust check failure is my fault, re me not doing [this](https://github.com/aaronleopold/stump/pull/81#issuecomment-1297563689), and the typescript one looks like the pnpm cache action just needs to run install if no cache is hit. I don't mind fixing those after merging this in a little later in the night, thanks again for all your work!! Much appreciated

aaronleopold:
#70

PR #88, opened 2023-02-12
fix: cookie not being set causing infinite loop

aaronleopold:
- Logic added to append the local IP addresses to allowed origins by default
- Removed the `secure` flag being set for the session config, which caused the browser to block the cookie being set
    - I'd like to see if I can make this configurable, but that will have to be tackled once I test HTTPS configs

I'll probably try tackling this [small but really annoying bug](https://github.com/aaronleopold/stump/commit/cbf9277c1e3ff173d1b6ab7d4112801469e231d3#diff-420c0a97234bcf99a6e9134b58be8d57975161c20ce2b4e0bc2b4b797c2791d5R74-R77) I found while at the airport before merging this in.


aaronleopold:
> I'll probably try tackling this [small but really annoying bug](https://github.com/aaronleopold/stump/commit/cbf9277c1e3ff173d1b6ab7d4112801469e231d3#diff-420c0a97234bcf99a6e9134b58be8d57975161c20ce2b4e0bc2b4b797c2791d5R74-R77) I found while at the airport before merging this in.

Fixed 🚀  see [here](https://github.com/aaronleopold/stump/pull/88/files#diff-420c0a97234bcf99a6e9134b58be8d57975161c20ce2b4e0bc2b4b797c2791d5R75-R100)

Issue #56, opened 2022-09-25
Docker permissions issue


aaronleopold:
Docker seems to create stump's config directory, on the host machine, with root privilege, making root the owner. This becomes a big problem during runtime, as stump won't have permissions to modify/access the config folder, which causes a ton of errors.

Changing the permissions manually fixes this, but this is really not ideal. 

I have only been able to reproduce this issue on Linux (both a dedicated install, as well as WSL). Config folder must not already be present (`~/.stump`), otherwise issue won't occur.

oliviacarino:
Hey! Just to clear things up, the goal is to give stump only the permissions to the config directory and not create it within the host machine? 

Can I take on this issue?

aaronleopold:
Hi! So to clarify, when the config folder (e.g. `~/.stump`) does not exist on the host machine, Docker creates it with the root user as the owner. So if I had this compose file:

```yml
version: '3.3'
services:
  stump:
    image: aaronleopold/stump-preview
    ports:
      - '10801:10801'
    user: '1000:1000'
    volumes:
      - /Users/aaronleopold/Documents/Stump:/data
      - /Users/aaronleopold/.stump:/config
```

When I run `docker compose up`, docker creates `/Users/aaronleopold/.stump` with root and Stump internally throws a bunch of permissions errors, e.g. cannot create database or cannot create log file, etc. So the goal is to figure out how to get docker to create that config folder on the host machine **not** owned by root. Does that make sense?

If you still would like it, feel free to take it on! Keep in mind, if it is for hacktoberfest **don't make the PR until it actually starts**. Otherwise, I am not sure it will be counted towards your total.

oliviacarino:
That makes sense, thank you :) I will still take it on!

tuxuser:
That docker's standard behaviour and is not really fixable afaik.

The alternative would be to use a docker volume instead of a host-path-binding. Would that be an acceptable solution?

aaronleopold:
Resolved from https://github.com/aaronleopold/stump/pull/89

PR #89, opened 2023-02-25
docker: Allowing to downgrade to regular user via environment variables

tuxuser:
This change implements an entrypoint-script for the docker container.

By passing environment variables `PUID=<numeric user id>` and `PGID=<numeric group id>` to the container, which is started as root by the docker daemon, can downgrade to the respective user and run the stump server daemon with lower privileges.

This also ensures that mounted directories/volumes are assigned and accessed with proper permissions.

**NOTE**: This essentially breaks the invocation via `--user` argument, currently documented on [Docker hub](https://hub.docker.com/r/aaronleopold/stump-preview)

- Fixes issue #56

Example docker compose

```yaml
version: '3.3'
services:
  stump:
    image: aaronleopold/stump-preview:latest
    container_name: stump
    volumes:
      - /home/user/.stump-config:/config
      - /media/books:/data
    ports:
      - 10801:10801
    environment:
      - PUID=1000
      - PGID=1000
    restart: unless-stopped
```

or via CLI

```
docker create \
  --name=stump \
  -e "PUID=1000" \
  -e "PGID=1000" \
  -p 6969:6969 \
  --mount type=bind,source=/Users/aaronleopold/.stump,target=/config \
  --mount type=bind,source=/Users/aaronleopold/Documents/Stump,target=/data \
  --restart unless-stopped \
  aaronleopold/stump-preview
```

### DIFF
--- a/scripts/release/Dockerfile
+++ b/scripts/release/Dockerfile
@@ -120,7 +120,7 @@ COPY --from=frontend /app/build ./app/client
 
 # Copy docker entrypoint
 # This will take care of starting the service daemon as a regular user, if desired
-COPY entrypoint.sh /entrypoint.sh
+COPY scripts/release/entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
 
 # TODO: replace this with something more elegant lol maybe a bash case statement
--- a/scripts/release/Dockerfile
+++ b/scripts/release/Dockerfile
@@ -105,10 +105,6 @@ FROM alpine:latest
 # libc6-compat
 RUN apk add --no-cache libstdc++ binutils
 
-# Create the user/group for stump
-RUN addgroup -g 1000 stump
-RUN adduser -D -s /bin/sh -u 1000 -G stump stump
-
 WORKDIR /
 
 # create the config, data and app directories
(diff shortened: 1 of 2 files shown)


scripts/release/Dockerfile
```diff
@@ -105,31 +105,24 @@ FROM alpine:latest
...
 # copy the binary
-COPY --chown=stump:stump --from=core-builder /app/stump ./app/stump
+COPY --from=core-builder /app/stump ./app/stump
 
 # copy the react build
 COPY --from=frontend /app/build ./app/client
 
+# Copy docker entrypoint
+# This will take care of starting the service daemon as a regular user, if desired
+COPY entrypoint.sh /entrypoint.sh
```
aaronleopold, line 123:
```suggestion
COPY scripts/release/entrypoint.sh /entrypoint.sh
```

It isn't ideal, but since this will be running from the project root, and not the `scripts/release` sub directory, I think we have to specify the entire relative path like that

aaronleopold:
Thanks so much, @tuxuser! I only had one comment, it looks great otherwise! I'm building an image now and can report back one I test it out

aaronleopold:
All looks good! I'll merge this tomorrow 🚀 

image

aaronleopold (approved):

PR #108, opened 2023-03-25
Navigation fixes

### LINKED ISSUE #45: Implement appropriate application navigation system
Navigation is really not ideal right now. The nav icons mimic the native browser navigation, but that is incorrect.

If I am at `/books/:id/pages/:page` and I go backward, it should take me to `/books/:id`, not the literal previous history (which might just be the previous page).

If I am on `/books/:id`, going backward should take me to `/series/:id`, etc

dancamdev:
Fixes #45 

- buttons now mimic native browser navigation buttons.
- disabled in the browser, native navigation buttons should be used instead.

A couple of extra fixes that felt natural at that point:
- I removed all references to `backwardUrl` and `forwardUrl` because it looked like they weren't needed anymore.
- EpubControls also uses the browser history when going back, instead of forcing it to go to `books/id`
- Fixed some tedious lint errors in `AppRouter`


packages/interface/src/AppRouter.tsx
```diff
@@ -13,23 +13,23 @@ type LazyComponent = Promise<{
...
-const SettingsLayout = lazily(() => import('./components/settings/SettingsLayout'))
-const GeneralSettings = lazily(() => import('./pages/settings/GeneralSettings'))
-const JobSettings = lazily(() => import('./pages/settings/JobSettings'))
-const ServerSettings = lazily(() => import('./pages/settings/ServerSettings'))
-const UserSettings = lazily(() => import('./pages/settings/UserSettings'))
-const FourOhFour = lazily(() => import('./pages/FourOhFour'))
-const ServerConnectionError = lazily(() => import('./pages/ServerConnectionError'))
-const LoginOrClaim = lazily(() => import('./pages/LoginOrClaim'))
-const OnBoarding = lazily(() => import('./pages/OnBoarding'))
+const Home = lazily(() => import('./pages/Home.js'))
```
aaronleopold, line 16: > Fixed some tedious lint errors in AppRouter

I assume adding `.js` was part of this? What was the lint?

dancamdev: Yes it was, also for some reason anything but 'js' would trigger an error as well.

Can't remember what the linter was, I'll let you know in the morning it's 11:30pm here and I'm done with code for the day 😅

aaronleopold: Huh, interesting - alright. I was worried those changes wouldn't bundle properly since those files with `.js` don't technically exist, but locally building it seems to have properly picked it up so 🤷 

> Can't remember what the linter was, I'll let you know in the morning it's 11:30pm here and I'm done with code for the day 😅

Haha no worries, have a good night!

packages/client/src/stores/useTopBarStore.ts
aaronleopold, line 1: Nice, with this I can finally get rid of this store, as it won't be used after `al/bye-chakra` gets merged

packages/interface/src/components/readers/utils/EpubControls.tsx
```diff
@@ -91,7 +91,10 @@ function EpubHeaderControls({
 				align="flex-start"
 			>
 				<ButtonGroup isAttached>
-					<IconButton variant="ghost" onClick={() => navigate(`/books/${epub.media_entity.id}`)}>
+					{/* FIXME: This navigate(-3) is effectively going back three times.
```
aaronleopold, line 94: Thanks for noting this!

aaronleopold (approved):
LGTM, thanks for working on this! I'll wait to merge until CI completes and to find out what those lints were for `AppRouter`, mostly out of curiosity 😄

PR #138, opened 2023-06-03
feat: RAR support (fix Windows)

aaronleopold:
Moved off of my old (bad) in-place byte reading implementation and using the WIP `0.5` feature branch from the upstream unrar.rs. I had to make a few adjustments to get things compiling, but you can see the fork [here](https://github.com/stumpapp/unrar.rs/tree/feature/typestate)

So far, tested on Windows and macOS. I will test Linux and Docker over the weekend 👏


aaronleopold:
Soooo it seems to be working on macOS, Windows, and Linux when running from source. Docker still throws a seg fault when inline-reading bytes, however 😭 

I'm inclined to merge this AS-IS, since it is better than it was as far as RAR support goes, but I don't want to dive too hard into C++ world right now to try and debug what is going on in Docker with `unrar.rs`. Conversions (which requires extracting the  RAR files FULLY) _also_ throws a seg-fault. So, I am not necessarily feeling this is necessarily an inline-reading problem anymore.  I think the next steps once I have time might be to create an isolated Rust app with `unrar.rs` and run it in a docker container to see when and how it breaks.

PR #148, opened 2023-07-27
separate thumbnail generation and other job manager improvements

aaronleopold:
I've restructured the job manager a little bit to make queuing jobs a bit more intuitive. It should also set it up to be a little bit easier to make jobs more stateful at some point down the road.

Thumbnail jobs are now separate from scans, and are queued once a scan is finishing up (assuming a thumbnail configuration is present). This is somewhat _dumb_ at the moment in that you cannot regenerate thumbnails yet, but I will add that in either before or after this PR merges.


aaronleopold:
A lot was added since making this PR:

- A good first pass jobs table was added to show a brief history of jobs
    - The pagination utils leave some to be desired, but is fine for now
- Future job scheduling section added
- Moved back off of SSE and trying websockets again for core->server->client events
- Fixed a bug with the avatar picker
- Added some additional checks in the `entrypoint.sh` to prevent  `adduser` and `su` from failing
- Added endpoint to delete all thumbnails associated with media in a library, should probably be made into a job some day
    - This is in a new `Quick Actions` section of the edit library page, which is now more accurately named manage library. I will try and experiment colocating other useful library actions here, but for now just the thumbnail removal.
- Fixed an issue with metadata parsing in the instance where `number` was a float. I imagine a few more metadata inconsistencies are bound to pop up and cause breakages. If it winds up being to inconsistent, we might just need to use strings.

One thing that I've made a note to fix is when doing a batch scan it seems to hang on the last book, but in reality the updates just stop while the server generates the massive SQL operations. I need to add a simple event to tell the client 'hey, I'm inserting stuff now'

Issue #67, opened 2022-10-04
Support markdown descriptions


aaronleopold:
One of my test files has markdown and I never noticed. I think it would be a really nice feature to allow for rendering and editing media descriptions as markdown. 

Will consider this for all entities with description fields.

dancamdev:
Let's do it - is the test file in the source code already? If not, mind sharing it?

aaronleopold:
This will be added alongside the filter and sort changes I am currently working on: https://github.com/stumpapp/stump/tree/al/filter-and-sort

PR #152, opened 2023-09-06
✨ support filtering and basic search

aaronleopold:
Finally added a good foundation for filtering and searching within the UI! The search is really naive at the moment, I can't implement a full text search until Prisma supports it for SQLite, but I think for the time being this is more than enough. It also doesn't look great on mobile ATM, but OK for now. 

- Build upon existing filter system I started in the server
- Started age restriction support, most of the work left for this will be adding the config options when creating a user and tightening up some of the query conditions
- Quickly added support for markdown descriptions (#67)
- Updated docs accordingly

<details>
<summary>
Screenshots
</summary>
Screenshot 2023-09-05 at 6 38 31 PM
Screenshot 2023-09-05 at 6 38 39 PM
</details>


aaronleopold:
Also to note, I forgot about sorting 🙈 the server supports it but I got so wrapped up in filtering that sorting slipped my mind. I'll make a follow-up PR sometime in the coming weeks to quickly add that in.

aaronleopold:
Also also, found this and love it so will adopt it (maybe) https://gitmoji.dev/

PR #155, opened 2023-09-08
:globe_with_meridians: New Crowdin translations

aaronleopold:


aaronleopold:
Adding all of the languages shown in Crowdin as selectable options in the UI and then merging these in. Thanks to all who completed some translations!

Issue #35, opened 2022-09-16
Basic PDF support


aaronleopold:
The server needs to be able to: process, properly handle and serve PDF files. This issue is for *basic* support, _page streaming_ support will either roll out separately or, depending on how the rest of the below goes, at the same time. 
 
Core Tasks:
- [x] Add logic to detect PDF mime
- [x] Add logic for PDF operations:
    - [x] Simple file serving
    - [x] Thumbnail extraction and/or generation
    - [x] Metadata extraction / interpretation
- [x] Start accepting PDF files during scan

Interface Tasks:
- [x] PDF Reader (via file serving)
- [x] Streamable PDF reader

berkingurcan:
I can take this.

aaronleopold:
Sure thing, just assigned you. Thanks!

aaronleopold:
I think this is promising: https://github.com/ajrcarey/pdfium-render

Depending on how fast of an operation it is to convert a PDF page to a bitmap, that could potentially work for streaming PDF pages. However, that library would easily allow for converting PDF to archive formats I think (also very great). Static compilation might be tricky, but there is an entire section for that in their README

KrautByte:
Agreed, I would like to create a general design spec for this first though, making sure every file format has the same methods so interfacing and documentation become easier if that makes sense.

aaronleopold:
Sounds good!

aaronleopold:
Mostly completed this today 🎊 per-page streaming as well as the option to use your browser's native PDF viewer are both available now.

The only thing left at this point is to update the Dockerfile to pull in the PDF dependency.

aaronleopold:
Resolved by https://github.com/stumpapp/stump/pull/158. I am deploying updated docs to reflect PDF support, but should eventually make a note about PDFium being used

Issue #156, opened 2023-09-09
PDF-to-ZIP conversion


aaronleopold:
I think it could be useful to support converting PDF files to archives of images. Fairly low effort to support this once #35 is officially merged.

DieselTech:
It certainly can be useful, but it's not as straight forward on paper. Depending on the library used and the complexity of the PDF, the conversion to images can miss layers that are more complex. PDF has a lot of nuances that can cause edges cases.

aaronleopold:
Good point(s). Honestly I’m not too familiar with how reliable PDF-page-to-image converting is using Pdfium, but also in general, since I generally avoid collecting PDFs myself for these very reason. That library is pretty much the only viable option for Rust I could find, though. Hopefully for scenarios I’d imagine this feature would be used (e.g. comic books, as opposed to more complicated PDF files with text or forms), it won’t present with too many issues.

hollisticated-horse:
Is this feature meant to enable users to effectively convert PDFs into CBR/CBZ files ?

PR #158, opened 2023-09-14
:sparkles: basic PDF support

aaronleopold:
Resolves https://github.com/stumpapp/stump/issues/35

With these changes, PDFs are mostly supported:
- PDFs can be scanned, and metadata (if any) can be extracted
- PDFs can have generated thumbnails
- PDF pages can be streamed
- PDF files can be sent to the client for download and/or viewed in a native PDF viewer

This pulls in two additional libraries in the core:

1. `pdf`
2. `pdfium-render`

The `pdfium-render` allows the core to properly process PDF pages as images. However it should be noted the concerns brought up from https://github.com/stumpapp/stump/issues/156#issuecomment-1712655256. I think, considering how few options there are, this will just have to be tried and tested over time.

There are some areas that need improvements, but I will follow up with them at another time:
- PDF page streaming will **only** output PNG images for the time being. This is hard coded, but will be replaced with some sort of option to configure it
- I added basic PDF-to-ZIP conversion support in the core, but am not hooking it up with UI elements yet. I'll leave that for the dedicated ticket for that feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -191,7 +191,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2c3d816ce6f0e2909a96830d6911c2aff044370b1ef92d7f267b43bae5addedd"
 dependencies = [
  "atk-sys",
- "bitflags",
+ "bitflags 1.3.2",
  "glib",
  "libc",
 ]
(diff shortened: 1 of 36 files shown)


aaronleopold:
A little update after some experimentation with the Docker builds locally. The added dependencies (the PDF crate and then the PDFium binary) and the additional PDFium stage to the docker build seem to have added about 40MB to the uncompressed size, totaling about 160MB now 😭 This isn't awful, but between the switch from Alpine to Debian I'd like to try and maintain and not add too much more. I imagine there are some fancier things to be done with the layers that can reduce this increase, but it's fine for now. Otherwise, everything functions great!

PR #167, opened 2023-09-29
✨ Select custom thumbnails

### LINKED ISSUE #166: [FEATURE] abilty to set thumnails by selecting from current page
so as the title says, a feature in the web reader to make the current page the thumnail of the series or the current issue/vol/chapter

aaronleopold:
Resolves #166

PR adds logic to pick a page from a book to use as the thumbnail for itself of its associated series/library. After the initial release, I'll be sure to refactor this implementation to be more resilient against regeneration (e.g. so you don't have to manually select your override whenever you force-regenerate thumbnails)

For media and series, they are getting their own new management pages. Rather than adding a bunch of modals for future features, e.g. metadata management, all similar items will live there. For library thumbnails, it will go in the existing library management page.

Note that EPUB files won't be supported for this feature at this time. In the future, it could be possible to extend this to query for all image resources inside the EPUB and select from that, but I don't want to prioritize figuring that out now.


aaronleopold:
I'm going to eagerly merge this after it passes CI again, there is a bit of duplication for some of the UI logic, and it doesn't look _perfect_ IMO, but I'll tackle that another day so I can focus on other non-UI things for now.

This PR was adjusted to also add a fix for an unsafe `unwrap`, and the `order_by` clause that was missing on thumbnail API calls. Regarding the first one, before a release I should do a search and replace for all of the old `unwrap`s that might be left from when I first started the project 😅

PR #175, opened 2023-10-11
🚧 Add integrations crate

LRotenberger:
Started the development work towards getting various notification integrations (e.g. Discord, Telegram, etc) possible throughout Stump. 

I’ve added a new integrations crate that houses all of the notifier logic. I nested all notification-related clients in that sub crate so the integrations crate can eventually be extended to support non-notification centered things (e.g Google books API, see the shell file I started just for experimenting).


aaronleopold (approved):
Thank you!

Issue #123, opened 2023-04-22
[FEATURE] Previous/Next page keyboard shortcuts


omnisci3nce:
Being able to hit the left/right arrow keys or pageup pagedown keys to change pages in the EPUB reader would be nice

_Bonus points would be being able to list keyboard shortcuts somewhere for when you forget them (press `?` to open a shortcuts reminder?)_

omnisci3nce:
Scope on this one seems like it's reasonably small (excluding the Bonus points section which is really its own Issue) so I can have a go at this if you want.

aaronleopold:
Yeah the scope is relatively small, you're more than welcome to take a stab at it if you’d like!

JMicheli:
It seems like there hasn't been much movement on this issue, so I took a cut at it as a way of familiarizing myself with the repo (I love a Rust project). I will submit a pull request shortly.

aaronleopold:
Awesome, thanks! Keep [this comment](https://github.com/stumpapp/stump/issues/179#issuecomment-1773264561) in mind when you dig into that part of the code

omnisci3nce:
Yeah sorry I got pretty busy after commenting. I forked and made something that sort of worked but ran into some edge cases working with epub.js (because it gets focused) if I recall correctly and then got sidetracked by other stuff. Glad that you're picking it up tho!

aaronleopold:
Added in https://github.com/stumpapp/stump/pull/182

PR #182, opened 2023-10-24
feat: add support for arrow key navigation

JMicheli:
This pull request is a response to issue #123, which requested the left/right arrow keys be usable for previous/next page navigation when reading books.

I implemented this with a new effect in BookReaderScene.tsx which adds a keydown listener and navigates on `ArrowLeft` and `ArrowLeft` if the current page isn't the first/last page.

Happy to make any changes needed, I'm less familiar with React so I may not have done this the ideal way.

### DIFF
--- a/packages/interface/src/components/readers/epub/EpubJsReader.tsx
+++ b/packages/interface/src/components/readers/epub/EpubJsReader.tsx
@@ -1,3 +1,5 @@
+import test from 'node:test'
+
 import { API, updateEpubProgress } from '@stump/api'
 import {
 	type EpubReaderPreferences,
@@ -168,6 +170,24 @@ export default function EpubJsReader({ id, initialCfi }: EpubJsReaderProps) {
 					//* Color manipulation reference: https://github.com/futurepress/epub.js/issues/1019
 					rendition_.themes.register('stump-dark', stumpDark)
 					rendition_.on('relocated', handleLocationChange)
+
+					// This callback is used to change the page when a keydown
+					// event is recieved.
+					const keydown_callback = (event: KeyboardEvent) => {
+						// Check arrow keys
+						if (event.key == 'ArrowLeft') {
+							rendition_.prev()
+						}
+						if (event.key == 'ArrowRight') {
+							rendition_.next()
+						}
+					}
+					// The rendition fires the event when the epub page is in focus
+					rendition_.on('keydown', keydown_callback)
+					// The window should fire the event when the epub page isn't in focus
+					// However turning the page doesn't work, possibly because
+					// rendition.start() hasn't been called yet?
+					window.addEventListener('keydown', keydown_callback)
 
 					applyEpubPreferences(rendition_, epubPreferences)
 					setRendition(rendition_)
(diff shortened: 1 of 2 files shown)


aaronleopold:
Hello 👋 Thanks for your contribution!

I think for #182 your changes will have to be moved to/adjusted for the [EpubJsReader](https://github.com/stumpapp/stump/blob/main/packages/interface/src/components/readers/epub/EpubJsReader.tsx). All image-based readers actually already have [keyboard interactions](https://github.com/stumpapp/stump/blob/main/packages/interface/src/components/readers/image-based/PagedReader.tsx#L93-L118) for navigation. Since the few different media types have to be consumed in very different ways (largely a different between EPUB vs image-based media), the `BookReaderScene` is more of a weigh station to route to the proper route and/or reader component. It probably doesn't help that there are [3 routes](https://github.com/stumpapp/stump/blob/main/packages/interface/src/scenes/book/BookRouter.tsx#L24-L26) for book readers in Stump as a result 😅 

 I sort of (strongly) dislike working with epub.js, so wanted to look into this a little to help. A quick glance at [some examples](https://github.com/futurepress/epubjs-reader/tree/master/reader/js#L3310) on their repository suggests you would need to use the `rendition` object for this, maybe somewhere [here](https://github.com/stumpapp/stump/blob/main/packages/interface/src/components/readers/epub/EpubJsReader.tsx#L171). Reach out if you need any help moving forward and I'm more than happy to help!
 
 (Also if you're wanting to participate in Hacktoberfest let me know and I'll apply the appropriate label to this PR)

JMicheli:
Okay this is what I've got so far. This allows arrow navigation of epubs but it has a problem I haven't yet been able to solve: when the epub is navigated to a new "location" (e.g. going from the main text to the cover), the epub renderer loses focus and the arrow navigation won't work until the user clicks on the screen to focus on it.

I tried to traverse this by adding an additional listener to the window, since that will fire on keydowns when the epub is _not_ in focus, but this doesn't work. Although the event does fire, the `rendition_.next()` call does nothing. Not totally sure how things are working under the hood yet, but I'll continue investigating. Any ideas in the meantime?

aaronleopold:
Honestly I am not entirely sure. I did try out your changes locally, though:

> I tried to traverse this by adding an additional listener to the window, since that will fire on keydowns when the epub is not in focus, but this doesn't work.

I've tried a few different browsers and this additional work around you added seems to resolve the issue for me in all of them, strangely. And then removing your line:

```js
window.addEventListener('keydown', keydown_callback)
```

presents the issue I believe you are describing, where it loses focus after a seemingly random number of page changes. You're welcome to continue debugging, but I'm satisfied with the functionality I am seeing 😄 In the future, once epub chapter streaming is properly implemented, I have a dream of removing epub.js entirely.

JMicheli:
Hmm maybe I just forgot to hit F5 when testing it 😅, because I just rebuilt and it works fine for me as well. I'm going to push a cleaned up commit without all the fixme comments.

aaronleopold (approved):
Thank you!

PR #193, opened 2023-11-18
♻️ Various UI refactors and fixes

aaronleopold:
Mostly some odds and ends and chores

- Redesign sidebar, component itself was a little messy
- Add and use aliases throughout the interface package
- Add emoji support for libraries and book clubs because it is cute and I wanted it
- Add a config to enable WAL mode on the SQLite DB
- Fix the ability to unset age restrictions


aaronleopold:
Didn't have as much time this weekend to wrap up these changes, but since opening it I have added:

- Refactor of the `EntityCard` component(s) to not be a hot mess but also support fallback image if the initial failed to load
- Refactor of the `PrismaSessionStore` to *safely error*, instead of panicking with a bunch of `expect`s

I think I'll try and add the ability to _remove_ emojis (so you aren't stuck with it) and then I'll merge this shortly after

PR #198, opened 2023-11-22
🌐 Update translations

aaronleopold:


aaronleopold:
Note to future self: some of these overwrite existing translations 😮‍💨 don't just merge this in without going through them by hand

Issue #206, opened 2023-12-02
[BUG] "Read Again" starts at end


jacobkossman:
**Describe the bug**
When you've finished reading, the "read again" button starts at the end of the book. This should be simple to just lop off the ?page= at the end of the URL.

**To Reproduce**

1. Read an entire book
2. Go to the main page for the book
3. Click "Read Again" button
4. Starts on last page

**Expected behavior**
Should start on the first page

aaronleopold:
Hey 👋 thanks for reporting this!

What is the ideal behavior in this situation, in your opinion?

1. Maintain completion status (`completed_at` and `is_completed`) but change `page` state for progress (i.e. reset to 1)
2. Remove completion status and change `page` state for progress (i.e. hard progress reset)
3. Some other third thing

I wanted to double check because, depending on which is the ideal set of actions, this could affect things like [getting the in progress media](https://github.com/stumpapp/stump/blob/main/apps/server/src/routers/api/v1/media.rs#L455-L458) (which currently filters for media with `is_completed` set to false)

jacobkossman:
I think `completed_at` should be a list of dates (with at least one) then `is_completed` and `page` is reset. Basically the way Plex works where you can watch things multiple times but reset the watch state of an episode / show / season etc.

aaronleopold:
> I think completed_at should be a list of dates (with at least one)... Mostly so that in the future you can look and be like "i read this on X dates" etc

I do like this functionality where you can see all the times you've completed a book like that, but it would require a surprising amount of changes to support. I'm going to branch that aspect off into a feature request, and lower the scope of this bug to just reset `is_completed` and `page` while leaving the `completed_at` date

aaronleopold:
> can you not make a new variable instead of changing the existing one? not sure if thatd be easier

Yeah, I outlined basically that in the linked ticket I created for this functionality:

```
Maybe rename read_progress(es) to current_read_session(s)
```

I maybe didn't explain it very well, but the idea would then be that completion status could be tracked by something like `completion_history` or honestly `read_history`, and `read_session` would track stuff like the page/epubcfi/etc. The issue is the work to do it either way is mostly the same, since the end goal would be to use the new column instead of the old one. So all the API and UI work would have to get done all the same

aaronleopold:
@jacobkossman This should be resolved once the next [nightly build completes](https://github.com/stumpapp/stump/actions/runs/7108960572). If you still encounter the issue let me know, otherwise I'll close this as completed

Issue #204, opened 2023-12-01
[FEATURE] Button to set as read


iwa:
**Is your feature request related to a problem? Please describe.**
You have to read the entire book to have it considered as read

**Describe the solution you'd like**
A simple button on the book page to set the book as read

jacobkossman:
I think a minor addition to this would be to mark as unread (so make this a toggle)

aaronleopold:
This was added as part of #208, and will be live once the next [nightly build completes](https://github.com/stumpapp/stump/actions/runs/7108960572). Thanks for the suggestion!

PR #208, opened 2023-12-03
✨ Book completed toggle button and fix read again button

aaronleopold:
Resolves https://github.com/stumpapp/stump/issues/206
Resolves https://github.com/stumpapp/stump/issues/204

### DIFF
--- a/apps/server/src/routers/api/mod.rs
+++ b/apps/server/src/routers/api/mod.rs
@@ -73,8 +73,13 @@ mod tests {
 		)?;
 
 		file.write_all(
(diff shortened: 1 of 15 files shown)


aaronleopold:
This should be good to merge once CI passes, a couple notes:

- I added endpoints to query for a book's completion status AND a series' completion status
- I added an endpoint for toggling a book's completion status
- I added an **unimplemented** endpoint for toggling a series' completion status
    - This will _eventually_ mark each book in the series as read. I'll try and add that in sometime soon
- Caught and fixed a couple of bugs related to prefetching, primarily the prefetch media and series queries. The TLDR; for those is the raw response, rather than just the data, was being cached, and the invalid type caused UI issues/errors

Issue #191, opened 2023-11-11
Restructure config


JMicheli:
As currently written, Stump's global state is stored in environment variables. For example, in [thumbnail.rs](../blob/master/core/src/filesystem/image/thumbnail.rs), `generate_thumbnail()` gets `thumbnail_path` from `config::get_thumbnails_dir()`, which is defined as follows:

```rust
/// Gets the Stump config directory. If the directory does not exist, it will be created. If
/// the path is not a directory (only possible if overridden using STUMP_CONFIG_DIR) it will
/// panic.
pub fn get_config_dir() -> PathBuf {
	let config_dir = std::env::var("STUMP_CONFIG_DIR")
		.map(|val| {
			if val.is_empty() {
				home().join(".stump")
			} else {
				PathBuf::from(val)
			}
		})
		.unwrap_or_else(|_| home().join(".stump"));

	check_configuration_dir(&config_dir);

	config_dir
}

// [omitting unrelated code]

pub fn get_thumbnails_dir() -> PathBuf {
	let thumbnails_dir = get_config_dir().join("thumbnails");

	check_configuration_dir(&thumbnails_dir);

	thumbnails_dir
}
```

Basically, any time a configuration variable is needed, an `std::env::var` call is used to extract the variable from the environment and then that value is used. Because of this, Stump must read configuration variables from the config TOML file at startup and then write them into the environment to ensure that they will be accessible when such calls are made.

This strikes me as suboptimal because it makes most functions nondeterministic based on their inputs. If the environment variables were to be changed (for whatever reason) while Stump is running then they would behave differently, even with the same input variables. I think that, ultimately, this is an antipattern, and it is one that will become increasingly difficult to undo as development continues.

My proposal is to rewrite the configuration of Stump so that it behaves the same way that it does now (i.e., loaded from the TOML file and then from the environment at startup) but instead stores the config in a struct which the program holds onto throughout its runtime and passes to the appropriate functions when they are called. Given that this is not a difficult task, just a laborious one, I'd of course be willing to work on it myself, but I want to open the floor to discussion before starting. Any and all thoughts (including reasons that this may not be as good an idea as I think) would be appreciated.

aaronleopold:
Hey 👋 thanks for writing this up! Sorry I'm only just now getting to it:

[...]

PR #214, opened 2023-12-09
:art: Refactor core config

JMicheli:
As detailed in Issue #191, Stump currently relies on storing configuration variables in the environment to maintain global state in the application. This is undesirable because 1) it means functions' inputs are not the only thing that determines their behavior, 2) configuration is scattered throughout the repo, and 3) configuration doesn't benefit from Rust's robust validation ecosystem.

To address this, this pull request applies fairly comprehensive changes to the stump_core crate as well as the server app (to maintain compatibility with the changes to the core). All values that were configured with environment variables are still configured with the same environment variables, however, those values are now stored in a `StumpConfig` struct at startup and this struct is maintained in an `Arc` throughout Stump's runtime. 

Other than this, the application should behave exactly the same.


core/src/filesystem/content_type.rs
aaronleopold, line 1: This isn't the only place you've done it, but thank you for all of the cleanup in the doc comments and throughout!

crates/cli/src/config.rs
```diff
@@ -11,3 +12,8 @@ pub struct CliConfig {
 	#[clap(long, env = "HASH_COST", default_value = "12")]
 	pub password_hash_cost: u32,
 }
+
+pub struct BundledConfigs {
```
aaronleopold, line 16: I'm not 100% sure we need this struct. The `StumpConfig` has both of the values present in `CliConfig`, so you could probably just get away with just using `StumpConfig`? Renaming `hash_cost` to `password_hash_cost`

core/Cargo.toml
```diff
@@ -50,3 +49,6 @@ toml = "0.8.2"
 tracing = { workspace = true }
 tracing-subscriber = { version = "0.3.15", features = ["env-filter"] }
 tracing-appender = "0.2.2"
+
+[dev-dependencies]
+tempfile = "3.8.1"
```
aaronleopold, line 54: Can we move this to the root `Cargo.toml` for the workspace? This way it is easier to maintain a pinned version for all crates who use it

aaronleopold (approved):
I had a few comments/nitpicks, but otherwise this looks great! Thank you so much for taking the time to work on this overhaul

core/src/config/mod.rs
```diff
@@ -1,72 +1,41 @@
...
-}
-
-pub fn get_thumbnails_dir() -> PathBuf {
-	let thumbnails_dir = get_config_dir().join("thumbnails");
-
-	check_configuration_dir(&thumbnails_dir);
-
-	thumbnails_dir
+		},
+		Err(_) => get_default_config_dir(),
```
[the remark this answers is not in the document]
JMicheli: I changed the bootstrap function to log an error for both non-set STUMP_CONFIG_DIR and in the case that it's set but empty. I think you're right that a user will want to know if the config directory is different than expected.

core/src/config/stump_config.rs
```diff
@@ -1,191 +1,290 @@
...
-				return Err(CoreError::ConfigDirDoesNotExist(config_dir));
-			}
+		if let Ok(client_dir) = env::var(CLIENT_KEY) {
+			env_configs.client_dir = Some(client_dir);
 		}
 
-		config.config_dir = get_config_dir().to_string_lossy().to_string();
-		config.write()?;
+		if let Ok(config_dir) = env::var(CONFIG_DIR_KEY) {
+			env_configs.config_dir = Some(config_dir);
```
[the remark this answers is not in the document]
JMicheli: I considered this but my thinking was that, because the `write_config_dir()` function will create a non-existing config directory, we might want to support that use case (i.e., the user sets STUMP_CONFIG_DIR to a directory that doesn't exist yet and lets Stump create it for them).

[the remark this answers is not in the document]
aaronleopold: Yeah you're totally right, that's a valid scenario

crates/cli/src/config.rs
JMicheli: Great point - I changed things up a bit to remove the surplus struct. Give it a glance and see what you think.

aaronleopold (approved):
Thanks for making those changes - I think this is good to go! Thanks again 👍

PR #264, opened 2024-02-09
✨API and DB definition for notifier feature

LRotenberger:
Implemented back-end for notification features + defined user permissions:

- Defined prisma model for Notifier and corresponding config
- Current supported platforms: Discord and Telegram
- Added swagger-ui types via utoipa
- Implemented associated permissions


core/src/db/entity/notifier.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+#[derive(Serialize, Deserialize, ToSchema, Type)]
+#[serde(untagged)]
+pub enum NotifierConfig {
+	Discord {
+		webhook_url: String,
+	},
+	Telegram {
+		encrypted_token: String,
```
aaronleopold, line 24: If we are storing an encrypted token for Telegram, we'll have to accepted an unencrypted token value during the creation flow. No UI exists yet, so this can be a follow up 🙂

core/src/db/entity/user.rs
```diff
@@ -162,14 +163,59 @@ pub enum UserPermission {
...
 	/// Grant access to manage the server. This is effectively a step below server owner
 	#[serde(rename = "server:manage")]
 	ManageServer,
 }
 
+impl UserPermission {
+	/// Return a list of permissions, if any, which are inherited by self
+	///
+	/// For example, UserPermission::CreateNotifier implies UserPermission::ReadNotifier
+	pub fn associated(&self) -> Vec<UserPermission> {
```
aaronleopold, line 190: ❤️

aaronleopold (approved):
LGTM! Thank you!

I had one comment but that can be tackled later once these features are closer to being exposed through the UI

PR #268, opened 2024-02-13
Simplify dockerfile

JMicheli:
This pull request removes some cruft from the Dockerfile, it does this by:

* Combining layers that need not be separate, such as ENV and RUN layers.
* Removing `WORKDIR /` lines - as the workdir is `/` by default.


aaronleopold (approved):
This looks great, thank you! I'll let CI run and will try to get it merged in after work

PR #282, opened 2024-02-23
🚧 Mobile App Foundations

LRotenberger:
- Overhauled previous mobile template with updated dependencies.
- Added basic tab navigation
    - Including home, explore, library, and settings screen
- Barebones comic reader integrated into app
    - Partial support for landscape mode
- Added first appearance settings
- Switched expo-router for explicit navigation

Note: Slight differences were noted in how iOS and Android load images, most importantly being that iOS includes the auth cookie while Android doesn't. Also a few other visual inconsistencies present in the theme, my main focus was not on visual design and since these issues don't affect usability, I didn't focus on fixing them right now.

Overall, mobile app is taking shape but still in very early stages


aaronleopold:
Super excited! Thank you ❤️

PR #289, opened 2024-03-09
:beers: POC EpubJS RN reader

aaronleopold:
Reorganized a bit of the reader screen to account for future readers a bit better, and added a really bare-bones EPUB reader. This frankly will not hold up in time, IMO, and I'll likely either have to move off of epubjs in favor of an alternative that works well with RN or just build something in house.


LRotenberger (approved):
LGTM

PR #331, opened 2024-04-29
:white_check_mark: Fix stump_config tests

JMicheli:
So I was investigating the failing tests as mentioned in the Discord earlier and I happened on a solution that seems sensible.

The crate [temp-env](https://crates.io/crates/temp-env) provides a way to set environment variables, run a test, then unset those variables. Embedding the tests involving environment variables in this crate's wrapper both makes it more ergonomic and also fixes the inconsistent passing of tests that I was experiencing (I interpret this as confirming that the previous tests were causing some sort of race condition where the environment variables for each were fighting each other).

Also, I set the telegram and discord tests to be ignored since they rely on local variables being set. It's probably a good idea to fix those in the future, but this will also open the way for automated benchmarking since that will need to rely on running tests and gauging their performance anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6423,6 +6423,7 @@ dependencies = [
  "serde-xml-rs",
  "serde_json",
  "specta",
+ "temp-env",
  "tempfile",
  "thiserror",
  "tokio",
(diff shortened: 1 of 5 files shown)


aaronleopold (approved):
LGTM, thank you!

PR #337, opened 2024-05-12
Fix some series/libraries-related dead links

PragmaTwice:
These links are dead (404), so I made a PR to fix them.


aaronleopold (approved):
Thank you!

PR #340, opened 2024-05-18
:sparkles: Historical read progress tracking

### LINKED ISSUE #207: [FEATURE] Historical record for completion dates of books

This feature request is a subset of some of the items discussed in https://github.com/stumpapp/stump/issues/206. Namely:

```
completed_at should be a list of dates (with at least one)... Mostly so that in the future you can look and be like "i read this on X dates" etc
```

I don't love the idea of storing a list of dates in SQLite, since I either have to stringify it or store as bytes, so perhaps just make a small lookup table for this. Maybe rename `read_progress(es)` to `read_session(s)`, I know I've been _dying_ to rename it. The idea would then be that completion status could be tracked by something like `completion_history` or `read_history`, and `read_session` would track stuff like the page/epubcfi/etc.

**Additional context**

The changes to the schema would be somewhat small, but there would be a decent amount of changes in the API and UI required to accommodate this.

aaronleopold:
Resolves #207

This PR splits the `read_progresses` table into two separate tables:

- `active_read_sessions` to track the **current** reading session for a user+book pair
- `finished_read_sessions` to track the **past and completed** reading sessions for a user+book pair

There were a lot of fundamental changes to filtering on read progress, as well as displaying progress throughout the UI, so I'll be testing these changes before being comfortable merging. A few behavioral notes:

- A user may have 1 `active_read_session` per book
- A user may have any number of `finished_read_sessions` per book
- Marking a book as complete sets the `started_at` and `completed_at` timestamps to:
    - The same value (current timestamp) if no `active_read_session` exists
    - The `started_at` value from `active_read_session` and the current timestamp, respectively
- Marking a book as unread will not clear historical record, it will clear the `active_read_session` if present

I think these changes also open the doors a bit to more time-based statistics features


aaronleopold:
I've done a bit more testing and fixed a few small issues. I'll try to simulate an upgrade from the current experimental to test the migration, since I wrote bits of it by hand I'd like to be extra cautious to not mess up peoples' read progress

PR #349, opened 2024-06-03
:sparkles: Page dimensions analysis task

### LINKED ISSUE #180: [FEATURE] API for getting dimensions of each page in a book
No

For all **image based media**, add an API endpoint that returns a list of dimensions for each page in the media file. If this information is not available via metadata, either:

1. generate it on the fly, update the internal metadata (this kinda goes again REST patterns, unless I do separate `POST` and `GET` options for fetching and potentially updating vs just fetching)
2. generate it on the fly, return results
3. return an empty list

Not sure what the ideal route is at this point

**Additional context**

https://discord.com/channels/972593831172272148/972595078554079232/1166115940539322438

JMicheli:
This pull request aims to close #180.

I've added a new type of task to the media analysis job created in #307, `AnalyzePageDimensions`. This task attempts to determine the dimensions of each page in a media item and then stores its findings to the database so that they can be accessed through the server's API.

The database structure for storing the dimensions data is shown below:
```prisma
model PageDimensions {
  id String @id @default(cuid())
  dimensions String
  metadata MediaMetadata @relation(fields: [metadata_id], references: [id], onDelete: Cascade)
  metadata_id String @unique

  @@map("page_dimensions")
}
```

An individual height, width pair is represented in Rust by `PageDimension`.

One of the concerns discussed when planning this addition was the possibility that a very long book would have a very long list of dimensions, thus necessitating the separate table to store dimensions, making them optional when loading metadata. But for this same reason, the string encoding a vector of `PageDimensions` could potentially be very large.

To avoid database bloat for users with large libraries/long books, I've implemented a simple compression scheme that serializes `Vec<PageDimension>` as follows:

```rust
// Comma and semicolon-separated serialization of height/width pairs.
let list = vec![
  PageDimension::new(800, 600),
  PageDimension::new(1920, 1080),
  PageDimension::new(800, 600),
];
let serialized_list = dimension_vec_to_string(list);
assert_eq!(serialized_list, "800,600;1920,1080;800,600");

// Repeated values are compressed with shorthand.
let list = vec![
  PageDimension::new(800, 600),
  PageDimension::new(800, 600),
  PageDimension::new(800, 600),
];
let serialized_list = dimension_vec_to_string(list);
assert_eq!(serialized_list, "3>800,600");
```

I think that this should be effective since many books should have uniform page sizes (excepting the cover and the occasional two-page spread). Thus, many books won't actually need to store that much data in the database. The deserialization function I wrote is copy-free and fast, I think it should not introduce any unwanted overhead.

Finally, the API matches the first option discussed on the issue:

`curl -X GET http://stump.cloud/api/v1/media/:id/page/:page/dimensions` gives a single dimension for a single page.
`curl -X GET http://stump.cloud/api/v1/media/:id/dimensions` gives a list of dimensions for each page (0-indexed).


aaronleopold:
@JMicheli The compression scheme you outlined is simple and smart! I'm excited to give this a review. I'll try to make time after work sometime this week, but it might land on the weekend. As always, thanks for your time and contributions - I really appreciate it!

core/src/filesystem/media/analyze_media_job/utils.rs
```diff
@@ -0,0 +1,65 @@
...
+		ContentType::PNG => Ok(image::ImageFormat::Png),
+		ContentType::JPEG => Ok(image::ImageFormat::Jpeg),
+		ContentType::WEBP => Ok(image::ImageFormat::WebP),
+		ContentType::GIF => Ok(image::ImageFormat::Gif),
+		_ => Err(JobError::TaskFailed(format!(
+			"Unsupported image format: {}",
+			content_type
+		))),
+	}
+}
```
JMicheli, line 65: Commenting on this to float a modification I may make. I think it makes sense to have a function like this (assuming there isn't one already but I overlooked it), but I think it would be better to do two things:

1. Include it in the module defining `ContentType`, maybe as a `From<image::ImageFormat` impl instead.
2. Remove the default handler from the match statement. We want this to fail to compile if `ContentType` gets a new variant in the future and prompt the person adding that variant to update this function.

core/src/filesystem/media/analyze_media_job/mod.rs
```diff
@@ -152,93 +163,26 @@ impl JobExt for AnalyzeMediaJob {
...
+			},
+			AnalyzeMediaTask::AnalyzePageDimensions(id) => {
+				task_analyze_dimensions::execute(id, ctx, &mut output).await?
+			},
+			AnalyzeMediaTask::FullAnalysis(id) => {
+				// First page count needs to be updated
+				task_page_count::execute(id.clone(), ctx, &mut output).await?;
+				// Then we can queue a dimensions analysis job
+				subtasks.push(AnalyzeMediaTask::AnalyzePageDimensions(id));
 			},
```
JMicheli, line 180: This is my solution for the need to run the page count analysis before the page dimensions analysis (because page dimensions relies on having page count).

I think a superior solution that is outside the scope of this pull request would be proving some sort of dependency graph for job tasks that enforces scheduling dependent tasks to occur first.

JMicheli (commented):
Adding some comments re: concerns I had while writing this code.

core/src/filesystem/media/analyze_media_job/utils.rs
aaronleopold: > Include it in the module defining ContentType, maybe as a From<image::ImageFormat impl instead.

I don't see one for `image::ImageFormat`, but I do see one for `image::ImageOutputFormat` in the `image/process.rs` file. I think either location would work (adjacent to `ContentType` or in the `image` crate), but I am definitely in favor of utilizing explicit `From` traits when possible.

> Remove the default handler from the match statement. We want this to fail to compile if ContentType gets a new variant in the future and prompt the person adding that variant to update this function

Yep I totally agree 👍

core/src/filesystem/media/analyze_media_job/mod.rs
JMicheli: Elaborating on this, this has one undesirable effect on user experience: when the task starts. As an illustrative example, imagine a user has a library with 100 items in it, and they select "Analyze Media." This fires the AnalyzeMedia job whose `init` queues up one `AnalyzeMediaTask::FullAnalysis` for each of the 100 items in the library.

This causes the indicator on the web UI to display `0/100`. However, when the first task runs (executing the page count logic) and queues up `AnalyzeMediaTask::AnalyzePageDimensions`, it's going to update the UI to read `1/101` and so on until it slowly ticks up to `200/200`. This is very annoying, and also confusing ("why does my 100 item library count to 200 when analyzing?" "why did an update 2 months later make it go to 300?").

I'm going to refactor to make `AnalyzeMediaTask::FullAnalysis` do both tasks' logic in series, which is fine because it looks like tasks run sequentially in the current job system anyway. This will make the UI count up to the number of media items being analyzed. It might be worth exploring task parallelism in the future.

aaronleopold: I know I was overall iffy about the UX when I realized that how I structured the job/scanner rewrite makes it easy for a growing queue, and I called it out in the docs so people aren't too confused: https://www.stumpapp.dev/guides/basics/scanner#walk-series

I think the performance gain during a scan is worth the cut to UX slightly, since calculating _all_ tasks at the beginning of a scan would be potentially very expensive, but I agree that ideally we should aim to reduce that sort of behavior where we can

> I'm going to refactor to make AnalyzeMediaTask::FullAnalysis do both tasks' logic in series, which is fine because it looks like tasks run sequentially in the current job system anyway

I think this works 👍 down the road, it might be good to explore an opt-in matrix sort of approach where I check various analysis tasks which the job understands at the start.

They do run sequentially because of my below point:

> It might be worth exploring task parallelism in the future

If they can be structured in a way to not issue DB writes, tasks can be parallelized. Throughout Stump's lifetime, I have observed and tested that Prisma+SQLite does *not* play well with concurrent writes (even with WAL enabled). Sometimes I wish I chose something like Postgres out the gate so folks can run 2+ scans concurrently without worry, or even just proper parallelism within a scan, but that can be a consideration down the road 😅

core/src/db/entity/metadata/media_metadata.rs
```diff
@@ -151,6 +156,12 @@ pub struct MediaMetadata {
 	/// the number of pages detected by the file processor.
 	#[serde(alias = "PageCount")]
 	pub page_count: Option<i32>,
+	/// The (height, width) dimensions of each page in the associated media. This is
+	/// generated by the image analysis job and is [None] before being generated.
+	#[serde(alias = "PageDimensions", default = "Option::default")]
```
aaronleopold, line 161: `PageDimensions` isn't actually part of the [ComicInfo.xml schema](https://github.com/anansi-project/comicinfo/blob/main/schema/v2.0/ComicInfo.xsd), so this alias won't be used for the time being. I think it Stump ever supports writing metadata, it could be useful though! It would _technically_ deviate from the schema, but I think this would be acceptable.

Also, there _is_ a [`Pages` field](https://github.com/anansi-project/comicinfo/blob/main/schema/v2.0/ComicInfo.xsd#L98C41-L98C50) in that schema. It has an attribute called `ImageSize`, which I think is less useful than dimensions.

None of this is to say I'm asking for changes, to be clear, just vomiting some context 😅

apps/server/src/routers/api/v1/media.rs
```diff
@@ -1577,14 +1583,123 @@ async fn start_media_analysis(
...
+
+	// Then pull off the page dimensions if available
+	let dimensions_entity = media
+		.metadata
+		.ok_or(APIError::NotFound(
+			"Media did not have metadata".to_string(),
+		))?
+		.page_dimensions
+		.ok_or(APIError::NotFound(
+			"Media metadata did not page resolutions".to_string(),
```
aaronleopold, line 1701:
```suggestion
			"Media metadata does not have generated page dimensions. Run analysis to generate them.".to_string(),
```

apps/server/src/routers/api/v1/media.rs
```diff
@@ -1577,14 +1583,123 @@ async fn start_media_analysis(
...
+	Ok(Json(page_dimension.to_owned()))
+}
+
+async fn fetch_media_page_dimensions_with_permissions(
+	ctx: &Arc<Ctx>,
+	session: &Session,
+	id: String,
+) -> APIResult<PageDimensionsEntity> {
+	// First get user permissions/age restrictions
+	let user = enforce_session_permissions(session, &[UserPermission::ManageLibrary])?;
```
aaronleopold, line 1667: I don't think we need any elevated permissions to access page dimensions, this should be public for any authenticated user with access to the book IMO. Did you have differing thoughts?

aaronleopold (commented):
I had a few really low priority comments/questions but this looks great!

I'll give it another review for my approval stamp once the conflicts are resolved ✅ 

Edit to add that I meant to call out all of your comments throughout the code. They are very helpful during review and will be beneficial in the long run. I appreciate it!

core/src/db/entity/metadata/media_metadata.rs
JMicheli: That's a good note, I realized halfway through this process that `MediaMetatdata` was doing double duty as a database object representation and also a deserialization type for xml metadata. I didn't appreciate the reasoning for "alias" so I just mirrored it.

I think it might be a good idea to make two types in the future - one that represents the DB object and another that explicitly represents ComicInfo.xml.

In the meantime, I'll remove the unnecessary alias.

JMicheli: I also added a `skip_deserializing` since it shouldn't ever be contained in the xml.

core/src/filesystem/media/analyze_media_job/mod.rs
```diff
@@ -154,85 +190,17 @@ impl JobExt for AnalyzeMediaJob {
...
+				task_page_count::execute(id, ctx, &mut output).await?
+			},
+			AnalyzeMediaTask::AnalyzePageDimensions(id) => {
+				task_analyze_dimensions::execute(id, ctx, &mut output).await?
+			},
+			AnalyzeMediaTask::FullAnalysis(id) => {
+				// First page count needs to be updated
+				task_page_count::execute(id.clone(), ctx, &mut output).await?;
+				// Then we can do the dimensions analysis
+				task_analyze_dimensions::execute(id, ctx, &mut output).await?
```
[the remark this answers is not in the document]
JMicheli: Yeah that's a good point - I should fix these tasks to use the same buffer instead of doing it twice.

I think it seems likely that this will be fixed in a future change that establishes a more pipeline-based method for media analysis. I'll add a todo for now flagging this inefficiency.

core/src/db/entity/metadata/page_dimension.rs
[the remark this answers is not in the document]
JMicheli: I'd be really into applying some compression to other database stuff. Let's discuss on Discord.

apps/server/src/routers/api/v1/media.rs
JMicheli: No, I think I just don't completely understand the permissions/authorization scheme being used and so I erred on the side of caution. I see that there's no need for library management permissions here so I changed this line to `let user = get_session_user(&session)?;` does that do what we want?

aaronleopold: I gotcha, yes I think your changes work perfectly. If you ever want an explainer on the permissions/authorization I'm happy to give a TLDR; on Discord

JMicheli:
I'm working on getting things ready to merge - as is tradition, this part is often the most difficult for me.

aaronleopold:
No rush on my part! Reach out if you need anything, but I would guess all you need to do is update your feature branch with this repo's experimental branch, either directly if you have a remote set up or just update your fork's experimental and then merge that in

JMicheli:
Alright I think that sound do it - you were right, just needed a little `cargo clean` to get things playing nice again.

aaronleopold:
I fixed a couple of clippy lints and a missing `Err`, but this should be good to go once the experimental finishes building. I'll merge it shortly, thanks again!

Issue #327, opened 2024-04-28
[FEATURE] Minimal file reads during initial scan


Strontium:
**Is your feature request related to a problem? Please describe.**
When working with a remote file system to the stump host, like keeping files on a cloud storage provider, there may be both bandwidth and egress limits when accessing data. 
Currently when doing an initial scan for a book, stump appears to request to read the entire file. When using cloud storage via an rclone mount, this scan causes the file to be downloaded in its entirety, which is inefficient and likely slow for large files. 
Opening the book for reading, stump appears to only request the pages it requires, hence rclone will only download those parts of the file, reducing traffic and loading times.

**Describe the solution you'd like**
Allow an option for reducing file access during the initial scan. 
Scan should be limited to:
- reading the first page for thumbnail creation
- reading metadata
Could be an option for 'minimal scan' if there are other benefits to reading the entire file.
Noting this may only be supported on file types that allow reading data without the entire file (ie ZIP with no compression).

**Describe alternatives you've considered**
Stump being remote storage aware and will limit itself appropriately automatically.

**Additional context**
In comparison to other applications in my testing with an rclone mount:
- Komga: Reads full file on SCAN and full file on OPEN
- Kavita: Reads partial file on SCAN and full file on OPEN
- Stump: Reads full file on SCAN and partial file on OPEN

aaronleopold:
> Stump: Reads full file on SCAN and partial file on OPEN

The main reason Stump reads the full file on scan is to determine the _actual_ page count. That operation involves iterating through each file e.g. in an archive to determine whether it is a valid page (re: image file). The validity check uses actual byte content, and falls back onto the extension, as a way of attempting to be more accurate in knowing what is truly a valid page. The only feasible way to allow for partial reads would be:

- Drop the accuracy and operate only on file extensions, trusting the extension is always correct
- Don't count the pages and rely on metadata. This has problems in that missing or inaccurate metadata would cause problems.

Strontium:
Is it often you would have invalid file with an incorrect extension? 
What is the downside to having a *potentially* inaccurate page count? 
I think that dropping that level of accuracy is a reasonable trade-off to improve both the initial scan performance and in my case, reducing remote traffic.
If it is easier to implement, i'd still be happy even if it was an optional feature and/or there were some pre-requisites to making it work (ie having the required metadata in the file).
Thanks for considering it.

aaronleopold:
[...]

PR #353, opened 2024-06-19
:bug: Fix nested file structure support for ZIP/RAR format

aaronleopold:
This PR corrects an issue where Stump incorrectly reads byte content (without unpacking) from archive files in the scenario where the underlying structure is nested, for example:

```bash
Comic book.cbz
  |_ ComicInfo.xml
  |_ Comic book folder
       |_ 000.jpg
       |   ...
       |_ 120.jpg
```

I've also adjusted some of the logic to cut the file reads for #327. Once I'm confident no other changes are necessary, this PR should be ready


aaronleopold:
This is likely good to go, I'll give it another quick look over after work and then aim to merge it. I'll likely put out a 0.0.4 sometime in the next 1-2 weeks, too.

PR #364, opened 2024-07-08
:construction_worker: Add code coverage tracking

JMicheli:
This is a very simple pull request that adds CodeCov-based automated code coverage tracking. It will require some repository owner intervention to get fully functional.

I've added an action that runs [cargo-llvm-cov](https://github.com/taiki-e/cargo-llvm-cov). This runs all tests in release mode and generates lcov.info, which contains coverage data (NB: if you're using a coverage tracking extension then you can use visualize this file in your IDE, very helpful for designing tests).

The next action uploads this data to [codecov.io](https://about.codecov.io/), which offers free code coverage hosting for open source projects. To get this working, you'll need to create an account on that website and write the API token to the `secrets.CODECOV_TOKEN` variable. You'll also want to set up their GitHub app and give it permission to access this repository. Once that's done, there should be a handy little bot on each pull request that tracks coverage.

On that front, I've added a bot config file at `.github/codecov.yml` which currently sets no expectations (`target: 0%`). You can increase this percentage to set minimum coverage goals for the project, and I imagine that someday you will want to begin ratcheting it up to prevent regressions in coverage. A bot-initiated task tests coverage against this target on each pull request and can block merge if it's below target.

If you want, you can also copy the badge codecov gives you markdown code for into the readme for a live readout on the repo front page (or maybe wait to do that until the coverage is higher).

Finally, I've re-enabled running tests in the ci (though the integration tests do still need fixing). I think this is a good idea - I've noticed that two of the rar tests currently fail, which they did not when I wrote them. This is a regression that running tests before merge would have caught. I'm going to try and figure out why it's happening and fix it before this merges, because right now the ci should be blocking merge on account of this failing test.


JMicheli:
I've also just added the following line to the vscode `settings.json`:

`"rust-analyzer.check.command": "clippy",`

This easy little step will ensure that users (me) are running clippy instead of `cargo check` and catching those little errors that occasionally prevent (my) merges from passing ci the first time.

aaronleopold:
The looks great - thanks for putting this together! 

> To get this working, you'll need to create an account on that website and write the API token to the secrets.CODECOV_TOKEN variable. You'll also want to set up their GitHub app and give it permission to access this repository

Done 👍

aaronleopold (approved):

PR #367, opened 2024-07-12
:sparkles: OPDS V2 Support

### LINKED ISSUE #311: [FEATURE] OPDS v2.0 support

Stump should expose a v2.0-compatible OPDS API

**Additional context**

https://drafts.opds.io/opds-2.0.html

aaronleopold:
Resolves #311

This PR brings in *experimental* OPDS V2 support 🎊 The TLDR; for those who are not aware of the difference: it is the latest (technically still in draft) version of the spec which outputs JSON instead of XML. You can read the specifics at https://drafts.opds.io/opds-2.0

There are still a few open questions, and a few pieces which need to be implemented. So, I have it annotated as a draft while I wrap up the remaining chunk(s) of work so CI can at least start testing the build for me. Expect this to be fully ready in a few weeks, most likely.

The main highlights are as follows:

- Add all relevant OPDS-related structs/enums/types in the core
    - Introduce the `derive_builder` crate for a builder instantiation of these
    - Add tests which assert builder and serialization functionalities
- Add `OPDSV2PrismaExt` for OPDSV2-specific queries/operations
- Add `HostExtractor` extractor to get host details
- Fix missing sessions relation loading during basic auth
- Fix favicon not loading correctly in server routes
- Fix nullish values being serialized in media metadata
- Add miscellaneous notes about Prisma 5.16.0 on the horizon, which will bring a good amount of SQLite improvements once PCR is updated

I am likely forgetting a few things, but I'll update the list as needed

## Notes

- The spec doesn't mention whether links have to be absolute or if relative is sufficient. ~I assume the latter, which might require some work if that is the incorrect route.~ it was not the correct route 😄 
- I haven't implemented facets yet, frankly I'm not 100% sure what their purpose is
- I'm using my best judgement when creating the feeds. E.g. what are the groups, when to load publications, etc. I would really appreciate user feedback on this once it lands, if possible (I know there aren't many clients which implement the V2 spec).
- There seem to be a few issues with the only OPDS V2 client I know of (Cantook). Some searching around shows certain things like missing covers and labels are an issue on Cantook's side ~, however download is not working yet~
- [This query](https://github.com/stumpapp/stump/pull/367/files#diff-4e142f23004ab85a892e5d7926ecbbb3835b4f31dda95107b641c2dd4db54c72R56-R85) isn't fully correct and needs to be sorted out


aaronleopold:
I worked on this feature a bit while traveling last week. I think this is mostly ready to go, at least as an experimental implementation (read: it isn't perfect but is more than functional enough for user testing and iteration).

I've called out a few things in the notes section of the PR description, but the two I want to call out again are:

1. Cantook seems to have various issues which affect Stump and other OPDS-V2-compatible servers (i.e. Komga). Namely, covers and labels for OPDS publications don't seem to be working
2. Facets and Readium Progression are not implemented. I'm not sure how important the former is, but will work on the latter sometime after this initial implementation is merged.

core/src/opds/v2_0/utils.rs
```diff
@@ -0,0 +1,85 @@
...
+		&format!(
+			r#"
+			WITH ranked AS (
+				SELECT id, RANK() OVER (ORDER BY name ASC) AS position
+				FROM media
+				WHERE series_id = {{}}
+			)
+			SELECT id, position
+			FROM ranked
+			WHERE id IN ({})
```
aaronleopold, line 74: If any SQL wizards are watching, this needs to be refactored at some point

PR #369, opened 2024-07-17
👷 Release server binary

### LINKED ISSUE #310: [TRACKING] Build native binaries during release
An issue to capture the ask for building native binaries during release. The following platforms are planned to be supported:

- [x] Linux (x86_64)
- [x] macOS (x86_64, arm64)
- [x] Windows (x86_64)

## Additional context

There are a few actions which have already been defined that just need testing and refining. The following are relevant:

- [`build-server`](https://github.com/stumpapp/stump/blob/main/.github/actions/build-server/action.yml)
- [`build-web`](https://github.com/stumpapp/stump/blob/main/.github/actions/build-web/action.yml)

Realistically, caching needs to be implemented to speed up the build process before I would want to include this in the release process.

JMicheli:
This pull request adds a new workflow to build and release a binary package of the Stump server, so that users can directly run the server rather than using a Docker container. It aims to close #310 once complete.

The workflow is intended to be manually triggered, with the desired version number/tag for the release being input at that time. For example, the user could trigger the workflow with `version = 1.0.0` and the result would be a draft release called "Release 1.0.0" tagged "v1.0.0."

### DIFF
--- a/.github/workflows/release_binary.yml
+++ b/.github/workflows/release_binary.yml
@@ -1,0 +1,94 @@
+name: 'Binary release'
+
+# Manual release with input version number
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Enter the release version (e.g., 0.1.0)'
+        required: true
+
+jobs:
+  # This job builds the binaries for each release
+  build-binary:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+        include:
+          - os: ubuntu-latest
+            artifact_name: 'linux-artifact'
+            artifact_filename: 'stump_server.o'
+          - os: windows-latest
+            artifact_name: 'windows-artifact'
+            artifact_filename: 'stump_server.exe'
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Build web application
+        uses: ./.github/actions/build-web
+
+      - name: Build stump server
+        uses: ./.github/actions/build-server
+
+      - name: Upload server artifact
+        uses: ./.github/actions/upload-artifact
+        with:
+          upload-name: ${{ matrix.artifact_name }}
+          upload-path: target/release/${{ matrix.artifact_filename }}
+  # This job creates the release and organizes associated files
+  create-release:
+    needs: build-binary
+    runs-on: ubuntu-latest
+    permissions: write-all
+    steps:
+      - name: Download artifact (Linux)
+        uses: actions/download-artifact@v4
+        with:
+          name: linux-artifact
+          path: StumpServer-linux
+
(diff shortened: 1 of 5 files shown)


.github/workflows/release_binary.yml
```diff
@@ -0,0 +1,94 @@
...
+        uses: actions/create-release@v1
+        id: create_release
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        with:
+          draft: true
+          prerelease: false
+          release_name: Release ${{ github.event.inputs.version }}
+          tag_name: v${{ github.event.inputs.version }}
+          body_path: ./.github/CHANGELOG.md
```
aaronleopold, line 74: Do you know if this just dumps the entire changelog? Or will it somehow isolate changes relative to the diff? I assume the former, but if not possible perhaps we could use the `gh` CLI tool instead since it has a `--generate-notes` argument

JMicheli: Probably just dumps the entire changelog. I didn't realize there was a way to excerpt only the current part, I'll look into updating this later.

.github/workflows/release_binary.yml
```diff
@@ -0,0 +1,84 @@
...
+          - os: windows-latest
+            artifact_name: windows-artifact
+            artifact_filename: stump_server.exe
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Build web application
+        uses: ./.github/actions/build-web
```
aaronleopold, line 33: Based on the latest CI run, uploading with the same key within the context of the same workflow seems to be problematic:

```bash
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

Since it is just a bundle of CSS/JS/etc, it shouldn't _really_ matter if the build happens within the matrix here as it should be the same across platforms. It might be acceptable to move this step to a separate job in this workflow just for the webapp (e.g. `build-webapp`)

JMicheli:
I just pushed another commit that cleans up the testing stuff since this appears to be working. Won't be shocked if it requires further tuning, but I think it's likely ready to merge in.

One thought: should it be merged with main instead since it probably needs to be for the action to be manually triggered, and that's where releases come from anyhow?

aaronleopold:
> One thought: should it be merged with main instead since it probably needs to be for the action to be manually triggered, and that's where releases come from anyhow?

It's okay to merge into `develop`, it'll be in main before the next release (I create the tag after merge to main, which triggers the docker build)

aaronleopold (approved):
LGTM! Thanks for knocking this out!

PR #373, opened 2024-07-27
:hammer: Update docker-compose file example in docs to remove version

fehwalker:
Current `docker compose` file standard no longer uses the `version` element and having it throws a warning.

```
WARN[0000] ./stump/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
```

https://docs.docker.com/compose/compose-file/04-version-and-name/

Updated the example in the docs to remove `version`.

\* let me know if you want an issue opened as well, but this seemed tiny.


aaronleopold (approved):
Thank you!

PR #374, opened 2024-07-29
:construction_worker: Add `macOS` platform to binary release workflow

### LINKED ISSUE #310: [TRACKING] Build native binaries during release

aaronleopold:
This builds off of the work from #369 to close #310

I'll try to test it during the week with some debug commits to trigger the workflow before merging this in

### DIFF
--- a/.github/actions/build-server/action.yml
+++ b/.github/actions/build-server/action.yml
@@ -10,10 +10,15 @@ runs:
     - name: Setup rust
       uses: ./.github/actions/setup-rust
 
-    # This action can be called to build in a Linux or Windows runner
+    # This action can be called to build in a Linux, macOS, or Windows runner
     # This step only runs when using Linux
     - name: Compile server (Linux)
       if: runner.os == 'Linux'
+      shell: bash
+      run: cargo build --package stump_server --release
+    - name: Compile server (macOS)
+      # Alternatively, when running macOS
+      if: runner.os == 'macOS'
       shell: bash
       run: cargo build --package stump_server --release
     - name: Compile server (Windows)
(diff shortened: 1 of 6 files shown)


JMicheli (commented):
Looks good to me!

One thing I'm not familiar with in the MacOS context - do you need two binaries, one for Apple Silicon and one for non-Apple hardware?

aaronleopold:
> One thing I'm not familiar with in the MacOS context - do you need two binaries, one for Apple Silicon and one for non-Apple hardware?

Ah yeah, I think you're correct. I'll look more into it this week to see how much would change. It might just be a matter of adding an additional matrix for arm macOS vs intel

aaronleopold:
Just because I find it interesting, the builds completed in the following order:

- macOS (ARM) 11m 32s
- Ubuntu 13m 20s
- macOS (intel) 15m 17s
- Windows 24m 1s

DieselTech:
Interesting but weird at the same time. Then again `rustup` on windows is a bit jank.

aaronleopold:
Just verified the arm binary worked as expected, so I reverted the debug commit to trigger the workflow and I'll merge this momentarily. Thanks for the catch @JMicheli!

PR #379, opened 2024-08-01
:pencil2: Fix infered typo

fehwalker:
Quick fix of a typo I noticed while looking at logs (`infered` -> `inferred`). Doesn't appear anywhere else, just this one place.


aaronleopold (approved):
Ha, good catch. Thank you!!

PR #380, opened 2024-08-01
🐛 Fix bug in `WebpProcessor::generate`

JMicheli:
This pull request reverts some changes made in #351 which inadvertently caused generating Webp thumbnails to fail.

The problem, as identified in the Discord by @fehwalker, was caused by changes to `WebpProcessor::generate` which assumed that the input data was formatted as webp, meaning that attempts to convert non-webp images into webp would fail.

I've retained my changes to `WebpProcessor::generate_from_path` because they allow tests to pass that, in my view, should pass. In any event, this function is not called within Stump at any point, so it is likely irrelevant.

### DIFF
--- a/core/src/filesystem/image/webp.rs
+++ b/core/src/filesystem/image/webp.rs
@@ -1,6 +1,6 @@
 use std::fs;
 
-use image::{imageops, io::Reader, DynamicImage, EncodableLayout, GenericImageView};
+use image::{imageops, DynamicImage, EncodableLayout, GenericImageView};
 use webp::Encoder;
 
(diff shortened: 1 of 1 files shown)


aaronleopold (approved):
LGTM. Thanks for the quick fix!!

> I've retained my changes to `WebpProcessor::generate_from_path` because they allow tests to pass that, in my view, should pass. In any event, this function is not called within Stump at any point, so it is likely irrelevant.

I think this is the right call, at least for a small PR aimed at fixing a bug. We can reevaluate the trait and this particular struct's impl of it as needed 🙂

PR #387, opened 2024-08-10
🔨 Fix cargo watch tooling

JMicheli:
This pull request fixes an issue I was experiencing on my Windows development environment (though I do not believe it occurs on other environments).

The problem I had was that when running a dev server with `yarn dev:web`, `cargo watch` would constantly refresh the app. Running `cargo watch --debug` revealed that it was being caused by changes detected in the `core/prisma` directory, which seemed to coincide with database writes. 

The rust-lang GitHub page has an [issue discussing this problem](https://github.com/rust-lang/www.rust-lang.org/issues/827) and notes that `cargo watch` triggers on any changes to the directories it is watching (not just source files) unless that file is covered by the .gitignore. Although the .gitignore does cover `*.db*`, it seems that writes to the `core/prisma` directory generally were triggering recompiles.

What I ended up doing was putting the `dev.db` etc. files in `core/prisma/dev` instead, then added that to the .gitignore directly. It worked - no more random recompiles on Windows. Admittedly, this is a bit more cluttered than before, but I think it's worthwhile for me at least, and might help other devs in the future.

I also added indentation to a few doc comments because it made clippy mad.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,7 +44,6 @@ apps/server/client/*
 *.db*
 *.sqlite*
 *prisma.rs*
-core/prisma/dev
 
 # nix
 .envrc
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ apps/server/client/*
 *.db*
 *.sqlite*
 *prisma.rs*
+core/prisma/dev
 
 # nix
 .envrc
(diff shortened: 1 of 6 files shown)


aaronleopold:
Have you tried passing an additional --ignore flag to cargo watch?

JMicheli:
> Have you tried passing an additional --ignore flag to cargo watch?

Oh. yeah you're right, that's an easier fix. I removed those other changes and fixed the problem by adding `--ignore core/prisma`.

aaronleopold:
Nice! Happy it worked

aaronleopold (approved):

PR #389, opened 2024-08-10
:bug: Fix 422 error relating to invalid user permission

aaronleopold:
Fixes an issue discussed on Discord: https://discord.com/channels/972593831172272148/972595078554079232/1271953758326227065

### DIFF
--- a/apps/server/src/routers/api/v1/user.rs
+++ b/apps/server/src/routers/api/v1/user.rs
@@ -34,7 +34,7 @@ use utoipa::ToSchema;
 use crate::{
 	config::{session::SESSION_USER_KEY, state::AppState},
 	errors::{APIError, APIResult},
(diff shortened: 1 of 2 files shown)


apps/server/src/routers/api/v1/user.rs
```diff
@@ -454,26 +454,24 @@ async fn create_user(
...
 				.collect::<Vec<String>>();
 
 			let created_user = client
 				.user()
 				.create(
 					input.username.to_owned(),
 					hashed_password,
-					vec![
-						user::is_server_owner::set(false),
-						user::permissions::set(Some(permissions.join(","))),
```
aaronleopold, line 474: This line was responsible for adding an empty string when no permissions were provided during user creation

PR #392, opened 2024-08-11
👷 Run ci on pushes to develop

JMicheli:
This small change will run the CI when a change is made to `develop` (in addition to `main`) and should make it easier to track improvements in coverage over time.


aaronleopold (approved):
Thank you!

Issue #372, opened 2024-07-26
[BUG] Cannot delete a user from the list


gpatruno:
**Describe the bug**
I'm trying to delete a user I've created. When I look at the documentation, it says : 
To delete a user, navigate to /settings/users in your browser. Locate the Users table and click the action menu button (three dots) for the user you wish to delete.
However, I can't find the menu button for deleting a user. 
I used the Docker image aaronleopold/stump:nightly, aaronleopold/stump:0.0.4 and aaronleopold/stump:0.0.3 and no one of these versions have a menu button in the users table.

**To Reproduce**
As administrator, create a user with rights. Still as admin, go to /settings/users to display the list of users. 

As Server Owner:
1. Go to '/settings/users' 
2. Click on 'Create User' 
3. reload the page '/settings/users'
4. the new user appears in the list but without a menu

**Expected behavior**
I don't know if this feature is currently under development. I assume it should be present as it is documented. 
Maybe I'm using the wrong Docker image or a wrong config... . 

**Screenshots**
![tableusers](https://github.com/user-attachments/assets/2bd8af2b-7b37-4b3f-a62e-a4c94531791d)

Docker config:
```
services:

  stump:
    image: aaronleopold/stump:nightly
    container_name: stump
    volumes:
      - ./config:/config
      - ./media/books:/data
    ports:
      - 10801:10801
    environment:
      - PUID=1000
      - GUID=1000
      - STUMP_CONFIG_DIR=/config
    restart: unless-stopped

volumes:
  stump-volume:
```

**Additional context**
I've tried to use Postman to delete the user with his ID but I can't use the route. This may be a misuse, as I'm not familiar with the language used on the server side. However, I've retrieved the list of users by API with Postman.

aaronleopold:
Hey 👋 thanks for reporting this!

I think this is just an instance where the UI needs to be tweaked a bit. The menu should be present but might be overflowing and require a horizontal scroll to access it. Can you confirm that the menu exists after attempting to scroll horizontally within the table?

gpatruno:
Hi, Thank you for your quick feedback. 

Indeed, the button does exist in the table! 
After adding the following css classes to the table tag, the menu appears. 

`
display: block;
overflow: auto;
`
Interface with this CSS : 
![image](https://github.com/user-attachments/assets/cedf8c9c-ff1d-470b-80bc-ec6cc75c0517)

do you think it would be possible to add a little styling to the table to make the scrollbar appear if not all the elements are displayed? 

[...]

PR #393, opened 2024-08-12
:lipstick: Pin actions column on user table

JMicheli:
This pull request addresses an issue reported in #372. Although that issue has been closed with the addition of a scroll bar to the table, this pull request additionally pins the `actions` column on the user table to the right so that it will always be visible regardless of where the table has been scrolled to.

### DIFF
--- a/packages/browser/src/scenes/settings/server/users/user-table/UserTable.tsx
+++ b/packages/browser/src/scenes/settings/server/users/user-table/UserTable.tsx
@@ -120,6 +120,9 @@ export default function UserTable() {
 					onPaginationChange: setPagination,
 					pageCount,
 					state: {
+						columnPinning: {
+							right: ['actions'],
+						},
 						pagination,
 					},
 				}}


aaronleopold (approved):
I added basic support for pinned columns using the `Table` component, but it wasn't a finished implementation (see this [TODO comment](https://github.com/stumpapp/stump/blob/main/packages/browser/src/components/table/Table.tsx#L35)). The TLDR; is that there is no background on the action cell, which doesn't provide enough contrast if we were at a viewport which places the action above another cell.

I think this can be merged as-is, but wanted to put that context up there in case you were interested in working on the `Table` component further

JMicheli:
I'm interested in making sure this is done right - I saw the TODO but I didn't really understand the steps it was suggesting needed to be taken.

Could you describe, at a high level, what you're thinking needs to be done to Table here?

aaronleopold:
> Could you describe, at a high level, what you're thinking needs to be done to Table here?

I think for the short term, a few smaller changes in a few places of the table component could work without rewriting the whole thing to be modular (which that comment alludes to):

- I think there is currently padding on both the table _and_ the cell:
    - Remove padding on table, this will cause issues where elements are visible behind the cell
    - Place padding on cells (including header cells) instead: `first:pl-3` if not pinned left and `last:pr-3` if not pinned right
- The background needs to be set for pinned columns, but the table is currently unaware of what that should be. For example, the background would be `bg-background` but in the cards it is `bg-background-200`. So it needs a way to pass the class. I think the short term could just be adding a `tableCellClassName` prop and manually passing `last:bg-background-200` in this instance

Offhand that's all I can think of

aaronleopold:
I added changes relating to this here: https://github.com/stumpapp/stump/pull/402/commits/cc2b67eeef19befe161448345987624d1fd2c374. I'm going to close this out in favor of the linked PR

PR #397, opened 2024-08-15
:recycle: Implement a config macro to simplify `stump_config.rs`

JMicheli:
Although I like the way that `StumpConfig` works (taking a file, layering the environment on that, etc.), I think it was pretty cumbersome to update, requiring a lot of little edits throughout the file to add a single variable. There's no reason for that to be the case given Rust's macro system.

I wrote a procedural macro that takes an annotated struct as input and generates something functionally identical to the current contents of stump_config.rs, but many lines shorter. Better still, adding new variables should only take a few lines added to the struct definition. I'm pretty pleased with this one.

It passes all of the tests, which I suspect could be simplified as well.


aaronleopold (approved):
To the best of my ability, I think this looks good! I am _very_ unfamiliar with macros. Thanks for adding this!!

I'm going ahead and approving this, but I likely won't include it as part of `0.0.5`. I've already cut that release branch and plan to push it out today (was going to do it yesterday but then that GitHub outage happened). If I have to delay the release until the weekend, I'll include it though 🙂

JMicheli:
Don't delay the release, I'd like to do a bit more work on this and mainly put the pull request up for review purposes.

I'd like to add better error handling to the macro so that it properly highlights the cause of errors instead of panicking and highlighting the entire input.

I also noticed that it causes something to happen with the verbosity setting that differs from before this change - it turns off all logging (which results from verbosity = 0, the default) but prior to this change verbosity on dev instances of the server would be set to 1. Not sure what the interaction is there and I need to investigate a bit before we merge to avoid nightly users having empty logs.

aaronleopold:
> Don't delay the release, I'd like to do a bit more work on this and mainly put the pull request up for review purposes.

I mostly meant if something came up that made me have to delay 😅 but no problem! I'll re-request myself for review and mark it as draft until you're ready for me to take another look. Hopefully by then I'll have had the time to get a little more acquainted with macros. 

If you need anything during your investigation give me a shout!

crates/stump-config-gen/src/config_vars.rs
```diff
@@ -0,0 +1,124 @@
...
+	pub is_optional: bool,
+	pub is_vec: bool,
+	pub attributes: StumpConfigVariableAttributes,
+}
+
+#[derive(Debug)]
+pub struct StumpConfigVariableAttributes {
+	pub default_value: Option<Expr>,
+	pub debug_value: Option<Expr>,
+	pub required_by_new: bool,
```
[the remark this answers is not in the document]
JMicheli: I added a big doc comment to the macro (doesn't look like individual attributes can be document commented) that explains how to use it.

Take a look at that and see if it communicates the usage well enough, I can edit it if it's unclear still.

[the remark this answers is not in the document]
aaronleopold: That's perfect, thank you!

JMicheli:
I've made a few more additions, namely adding some documentation to the macro and adding a set of integration tests that fill the role of several tests on `stump_config.rs`, allowing them to be removed. That file has been significantly reduced in complexity now.

Assuming everything looks good, this should be ready to merge.

core/src/config/stump_config.rs
```diff
@@ -65,219 +67,91 @@ use defaults::*;
 ///   let core = StumpCore::new(config).await;
 /// }
 /// ```
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[derive(StumpConfigGenerator, Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[config_file_location(self.get_config_dir().join("Stump.toml"))]
 pub struct StumpConfig {
 	/// The "release" | "debug" profile with which the application is running.
+	#[default_value("debug".to_string())]
```
aaronleopold, line 74: I think part of the fix for the bug I reference in the 0.0.5 release notes would be to set this to `release` and then the debug default value to `debug`

core/src/config/stump_config.rs
```diff
@@ -65,219 +67,91 @@ use defaults::*;
...
+	#[env_key(PRETTY_LOGS_KEY)]
 	pub pretty_logs: bool,
+
 	/// An optional custom path for the database.
+	#[default_value(None)]
+	#[env_key(DB_PATH_KEY)]
 	pub db_path: Option<String>,
+
 	/// The client directory.
+	#[default_value("./dist".to_string())]
```
aaronleopold, line 100: Similarly as above,this default value should be `./client` as the second part of fixing the bug

core/src/config/stump_config.rs
JMicheli: Done, but do note that the `debug()` function is only used in tests - the actual runtime values in debug are set by the `debug_setup()` function in the server's `main.rs` by overriding environment variables.

We may be able to simplify this in the future.

Issue #219, opened 2023-12-11
[FEATURE] Provision session tokens on successful auth


aaronleopold:
**Is your feature request related to a problem? Please describe.**

Future development towards Windows desktop builds will be blocked by a Windows issue where webview2 blocks the cookies the server sends. macOS and Linux seem otherwise unaffected.

**Describe the solution you'd like**

Support sending some sort of token that can be used in the interface, this way the issue of the blocked cookie does not matter. This should ideally work alongside the server-side auth, as I don't necessarily want to remove that system (useful in existing features like log out on all devices, etc).

PR #400, opened 2024-08-16
:recycle: Update `axum` and refactor middlewares

aaronleopold:
PR primarily upgrades the `axum` dependency from `0.6.x` to `0.7.0`, but there is a LOT of refactoring as a result of the upgrade:

- Update `axum` and related dependencies
- Add `once_cell` to support JWT encoding/decoding secret
- Refactor the custom prisma session store accordingly after breaking changes
    - Update `expires_at` field in session model to `expiry_time` for easier (de)serialization into session record after upgrade
- Refactor all existing struct-based extractors into middleware functions
    - The big and important change relating to this would be the `RequestContext` extension effectively replacing the use of a `Session` extractor
    - Removed permission-specific extractors and created a new `permission_middleware` as a replacement 
- Add `api_error_message` to start standardizing error messages
- Removed unused logging middleware
- Add `access_token_ttl` to stump config
- Add `axum-test` as dev dependency to start introducing and think through server testing

I also want to specifically call out the one bit of *new* functionality in this change-set: access token support. This relates to #219, and will be bare-bones until that issue can be properly prioritized. I will also considered permission-gating it as a security measure. Noteworthy bits relating to JWT changes:

- Add `AuthenticationOptions` to login which affects the response, which is no longer just a `User` but a `LoginResponse` enum
    - Note I can see the `LoginResponse` changing once an in-house client uses JWTs, as I don't necessarily want to juggle an intersection type


aaronleopold:
Small note for anyone running `experimental`, this will delete all active auth sessions as part of the migration. Just an FYI

PR #405, opened 2024-08-19
:construction_worker: Restore self-hosted runner for certain CI tasks

aaronleopold:
This PR reintroduces my self-hosted actions runner for the majority of rust CI tasks, excluding the binary release workflow. The highlights include:

- Refactor the coverage tasks into a separate composite action
- Adjust the path filters for the CI workflows to be more specific to the job's domain (e.g. rust changes for rust tasks, etc)
- Fix issue on my runner relating to docs tests where hard crashes happened when attempting to run them
- Add step to enforce up-to-date bindings have been generated


aaronleopold:
I'm going to merge this in to unblock the nightly CI, but I'll want to revisit for additional testing before the next release

Issue #406, opened 2024-08-20
[BUG] Unable to run executables due to missing DLL


JMicheli:
This is likely a windows-specific problem (possibly related to my build environment).

After pulling down the latest develop, I am unable to run the executables generated by the project, including codegen and the stump_server executables. For example:

```bash
B:\Code\External\stump> cargo run --package codegen      
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.44s
     Running `target\debug\codegen.exe`
   Compiling prisma-cli v0.0.1 (B:\Code\External\stump\crates\prisma-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.90s
     Running `B:\Code\External\stump\target\debug\prisma-cli.exe generate`
Prisma schema loaded from prisma\schema.prisma

✔ Generated Prisma Client Rust to .\src\prisma.rs in 4.97s

Prisma client has been generated successfully!
   Compiling stump_core v0.0.5 (B:\Code\External\stump\core)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 11.69s
     Running unittests src\lib.rs (target\debug\deps\stump_core-6553912f58d5025a.exe)
error: test failed, to rerun pass `-p stump_core --lib`

Caused by:
  process didn't exit successfully: `B:\Code\External\stump\target\debug\deps\stump_core-6553912f58d5025a.exe tests::codegen --ignored` (exit code: 0xc0000135, STATUS_DLL_NOT_FOUND)
```

The same error occurs when starting the dev server using `yarn dev:web`.

Prior to this error, I had set up my environment to include the new dav1d dependency from #385 using vcpkg, and `dav1d-sys` was able to build correctly. I am wondering if the expected DLL might be related to this addition.

aaronleopold:
I’ll try to get it running on my desktop this week to see what I can help to figure out. Offhand I don’t what could have gone amiss with your environment, but almost certainly related to the dav1d requirements. 

I'd say double check that any path variables are still set and accessible in your shell and try a good ol' cargo clean as surface level  tests to start debugging

wolffshots:
I think this could be related to the issue that I'm seeing with the Docker container:
```
stump  | /app/stump: error while loading shared libraries: libdav1d.so.7: cannot open shared object file: No such file or directory
stump exited with code 0
```
Looks like just a lil oversight in the Dockerfile in #385?
[...]

PR #408, opened 2024-08-20
🔨 Fix docker `dav1d` shared library linking 

wolffshots:
Adding explicit linking to the process appears to fix the error that I was experiencing in Docker builds that I mentioned in [#406](https://github.com/stumpapp/stump/issues/406#issuecomment-2299452296):

```
stump  | /app/stump: error while loading shared libraries: libdav1d.so.7: cannot open shared object file: No such file or directory
stump exited with code 0
```

The Dockerfile still does not build locally despite adding a step to generate `prisma.rs` because of this error but that seems to be a separate issue:
```
                                                      ^^^^^^^^^^^^ help: a static with a similar name exists: `DATAMODEL_STR`
175.6
175.6 error[E0425]: cannot find value `MIGRATIONS_DIR` in module `super`
175.6  --> core/src/prisma.rs:3:3446816
175.6   |
175.6 3 | ...igrate_deploy (super :: DATAMODEL_STR , super :: MIGRATIONS_DIR , & self . 0 . url ()) . await ; :: prisma_client_rust :: tokio :: tim...
175.6   |                                                     ^^^^^^^^^^^^^^ not found in `super`
175.6
175.6 error[E0425]: cannot find value `MIGRATIONS_DIR` in module `super`
175.6  --> core/src/prisma.rs:3:3447215
175.6   |
175.6 3 | ...e (migration , super :: DATAMODEL_STR , super :: MIGRATIONS_DIR , & self . 0 . url () ,) . await } pub fn _db_push (& self) -> :: pris...
175.6   |                                                     ^^^^^^^^^^^^^^ not found in `super`
175.6
175.6 For more information about this error, try `rustc --explain E0425`.
175.6 error: could not compile `stump_core` (lib) due to 5 previous errors
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -63,15 +63,7 @@ WORKDIR /app
 
 COPY . .
 
-#Prisma generate makes running the Docker build locally/directly possible without extra config and shouldn't cause extra build time in CI because of caching
-#RUN cargo prisma generate --schema=./core/prisma/schema.prisma; \
-#    ./scripts/release/utils.sh -w; \
-#    set -ex; \
-#    ./scripts/release/utils.sh -p; \
-#    cargo build --package stump_server --bin stump_server --release; \
-#    cp ./target/release/stump_server ./stump_server
-
-RUN cargo codegen #cargo prisma generate --schema=./core/prisma/schema.prisma
+RUN cargo codegen
 RUN ./scripts/release/utils.sh -w 
 RUN set -ex
 RUN ./scripts/release/utils.sh -p
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,7 +2,7 @@
 # Frontend Build Stage
 # ------------------------------------------------------------------------------
 
-FROM node:20.0.0-alpine3.16 as frontend
+FROM node:20.0.0-alpine3.16 AS frontend
 ARG TARGETARCH
 
 WORKDIR /app
@@ -51,18 +51,32 @@ WORKDIR /dav1d/build
 
 RUN meson setup ../; \ 
     ninja; \
-    ninja install 
+    ninja install; \
+    ln -s /usr/local/lib/x86_64-linux-gnu/libdav1d.so.7.0.0 /usr/local/lib/libdav1d.so.7; \
+    ln -s /usr/local/lib/x86_64-linux-gnu/libdav1d.so.7.0.0 /usr/local/lib/libdav1d.so.7.0; \
+    ln -s /usr/local/lib/x86_64-linux-gnu/libdav1d.so.7.0.0 /usr/local/lib/libdav1d.so.7.0.0; \
+    echo "/usr/local/lib" >> /etc/ld.so.conf.d/mylibs.conf \
+    && ldconfig
 
 #Cargo build for stump
 WORKDIR /app
 
 COPY . .
 
-RUN ./scripts/release/utils.sh -w; \
-    set -ex; \
-    ./scripts/release/utils.sh -p; \
-    cargo build --package stump_server --bin stump_server --release; \
-    cp ./target/release/stump_server ./stump_server
+#Prisma generate makes running the Docker build locally/directly possible without extra config and shouldn't cause extra build time in CI because of caching
+#RUN cargo prisma generate --schema=./core/prisma/schema.prisma; \
+#    ./scripts/release/utils.sh -w; \
+#    set -ex; \
+#    ./scripts/release/utils.sh -p; \
+#    cargo build --package stump_server --bin stump_server --release; \
+#    cp ./target/release/stump_server ./stump_server
+
+RUN cargo codegen #cargo prisma generate --schema=./core/prisma/schema.prisma
+RUN ./scripts/release/utils.sh -w 
+RUN set -ex
+RUN ./scripts/release/utils.sh -p
+RUN cargo build --package stump_server --bin stump_server --release
+RUN cp ./target/release/stump_server ./stump_server
 
 # ------------------------------------------------------------------------------
 # PDFium Stage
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -2,7 +2,7 @@
 # Frontend Build Stage
 # ------------------------------------------------------------------------------
 
-FROM node:20.0.0-alpine3.16 as frontend
+FROM node:20.0.0-alpine3.16 AS frontend
```
wolffshots, line 5: This is just to appease the linting

docker/Dockerfile
```diff
@@ -51,7 +51,12 @@ WORKDIR /dav1d/build
 
 RUN meson setup ../; \ 
     ninja; \
-    ninja install 
+    ninja install; \
+    ln -s /usr/local/lib/x86_64-linux-gnu/libdav1d.so.7.0.0 /usr/local/lib/libdav1d.so.7; \
+    ln -s /usr/local/lib/x86_64-linux-gnu/libdav1d.so.7.0.0 /usr/local/lib/libdav1d.so.7.0; \
+    ln -s /usr/local/lib/x86_64-linux-gnu/libdav1d.so.7.0.0 /usr/local/lib/libdav1d.so.7.0.0; \
+    echo "/usr/local/lib" >> /etc/ld.so.conf.d/mylibs.conf \
+    && ldconfig
```
wolffshots, line 59: This could probably be done a lot more elegantly and generically but this seems to work so I'd be interested to see how I could do it better

docker/Dockerfile
```diff
@@ -51,16 +51,21 @@ WORKDIR /dav1d/build
...
 
 #Cargo build for stump
 WORKDIR /app
 
 COPY . .
 
-RUN ./scripts/release/utils.sh -w; \
+RUN cargo prisma generate --schema ./core/prisma/schema.prisma; \
+    ./scripts/release/utils.sh -w; \
```
wolffshots, line 67: This change allows me to build locally for testing but I'm happy to revert if needing

docker/Dockerfile
aaronleopold: I won't have time to test this until after work, but this won't account for arm64 builds

wolffshots:
I had to overwrite the entrypoint on my docker compose to get this to successfully run: `entrypoint: ["/bin/sh", "-c", "echo '/usr/local/lib' >> /etc/ld.so.conf.d/mylibs.conf && ldconfig && /entrypoint.sh"]` but I did get it to work with this branch!

wolffshots:
The change that I just pushed allowed me to just run the image without changing the `entrypoint` in my `compose.yml` and it appears to be up and running a-okay now

docker/Dockerfile
wolffshots: Yeah I'm happy to look at it again tomorrow to work out how to make it more generic (like using `TARGETARCH` or something). Defs just a hack for now to get it to work for me
If you have an idea of how to improve I'll get to it in the morning but it's late over here so gonna call it for this attempt

docker/Dockerfile
```diff
@@ -51,16 +51,21 @@ WORKDIR /dav1d/build
...
     set -ex; \
-    ./scripts/release/utils.sh -p; \
```
wolffshots, line 63: Note that I actually had to remove the `prisma.rs` text replacements because it was actually generated in the `Dockerfile` above so was already correct

docker/Dockerfile
```diff
@@ -93,6 +98,8 @@ FROM debian:buster-slim
 RUN apt-get update && apt-get install -y locales-all && rm -rf /var/lib/apt/lists/*; \
     mkdir -p config && mkdir -p data && mkdir -p app
 
+COPY --from=builder /usr/local/lib /usr/local/lib
+COPY --from=builder /etc/ld.so.conf.d/mylibs.conf  /etc/ld.so.conf.d/mylibs.conf
```
wolffshots, line 102:
```
COPY --from=builder /usr/local/lib /usr/local/lib <- this is necessary
COPY --from=builder /etc/ld.so.conf.d/mylibs.conf  /etc/ld.so.conf.d/mylibs.conf <- this can probs be removed because I do it in `entrypoint.sh` now
```

aaronleopold:
@wolffshots I made a few changes, since nightly is broken I wanted to make sure at the very least a working image goes out sooner rather than later. I appreciate your fixes! I translated them slightly to work for the arm64 build and then tweaked the `ldconfig` to the final layer in order to omit having to add it in the entrypoint exec. I left your change to generate the prisma client in the docker file, mostly because I am curious how it performs. If it winds up having to download the engines each time, I'll likely revert it. Best case scenario, though, it makes it easier for folks like you to build the images locally without needing a rust environment setup 🙂 

After some discussions on Discord, once this fix is merged I will likely revert the AVIF-specific additions until another time when a more batteries-included solution can be put together.

wolffshots:
Great stuff! Glad to have been able to help and thanks for all the hard work

aaronleopold:
> I left your change to generate the prisma client in the docker file, mostly because I am curious how it performs. If it winds up having to download the engines each time, I'll likely revert it. Best case scenario, though, it makes it easier for folks like you to build the images locally without needing a rust environment setup 🙂

I had to remove it, with prisma generation it stalled after 4+ hours on the arm64 build. The build without it completed both amd64 and arm64 in just over an hour 😅 

**But** I added some parameters you can use to generate it in the docker build. You can do something like:

```bash
PLATFORMS='linux/amd64' TAG='local' RUN_PRISMA_GENERATE=true ./docker/build.sh
```

wolffshots:
> I had to remove it, with prisma generation it stalled after 4+ hours on the arm64 build. The build without it completed both amd64 and arm64 in just under 2 😅

Fair that makes sense

> But I added some parameters you can use to generate it in the docker build. You can do something like:
>```
> PLATFORMS='linux/amd64' TAG='local' RUN_PRISMA_GENERATE=true ./docker/build.sh
> ```

Great stuff, that's all good by me

PR #415, opened 2024-08-23
ImageSize Crate for dimension information

smldub:


aaronleopold:
@smldub Did you mean to close this? I just took a look and I think it was looking good. Happy to help get it finished if you want, or I can rebase your efforts here into another image-related feature branch I have

smldub:
Unfortunately, I was running into an issue where the scan wouldn't work (it would say complete, but finish super fast and not update any dimension values) unless I loaded the entire file (which defeats the purpose a little bit). Even if I loaded the first 10000 bytes it would have an issue, which I was unable to recreate with just a simple test. I don't know really anything about async rust, or how to test/debug it :(. But feel free to use any of the code!

aaronleopold:
> Even if I loaded the first 10000 bytes it would have an issue

Interesting 🤔 Well, if I have time this weekend I'll pull it down and see if I can debug it further. I'll definitely keep you posted if I figure anything out!

PR #416, opened 2024-08-23
:construction_worker: Switch to `podman` and introduce better caching

aaronleopold:
This PR swaps docker CI builds to use `podman` and adds cache mounts to improve overall build times

### DIFF
--- a/.github/actions/build-docker/action.yml
+++ b/.github/actions/build-docker/action.yml
@@ -71,19 +71,23 @@ runs:
         username: ${{ inputs.username }}
         password: ${{ inputs.password }}
 
-    - name: Run buildx build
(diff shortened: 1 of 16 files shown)


aaronleopold:
I've refactored the docker workflow to use the podman engine instead and added some cache mounts. Note that before swapping to podman, the cache mounts caused a hang in arm64 for whatever reason. 

I've gotten a 44s total build time for a consecutive build, each layer hit the cache, but that is very likely an overly eager and unrealistic scenario. The next test will be ensuring the compiled deps stay cached when actual code changes

aaronleopold:
I've gotten successful builds multiple runs now, so I think this PR should be good to merge

PR #417, opened 2024-08-24
:sparkles: Avif file serving support

smldub:
Minimal changes to allow for AVIF files to be served to browser (I believe almost all browsers support [AVIF](https://caniuse.com/avif) at this point).


aaronleopold (approved):
Looks great, thank you!

I'll merge this momentarily. Note that I'm currently messing with the runners so CI might not behave

PR #424, opened 2024-08-30
:green_heart: Fix docker build push

aaronleopold:
lol so I forgot to add `--push` when publishing the builds 🙃

### DIFF
--- a/.github/actions/build-docker/action.yml
+++ b/.github/actions/build-docker/action.yml
@@ -80,7 +80,7 @@ runs:
 
     - name: Run podman buildx
       run: |
-        ENGINE=podman PLATFORMS="${{ inputs.platforms }}" TAGS="${{ env.TAGS }}" GIT_REV="${{ env.GIT_REV }}" ./docker/build.sh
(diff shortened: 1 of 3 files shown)


aaronleopold:
So I have a suspicion that `podman buildx`, when used in this way for building multi-arch images, will overwrite the tag for whichever build happens last (since it is not parallel builds). If this is true, then the next nightly push will only have an `arm64` build on dockerhub. This is obviously a problem. I'll divert my attention to the build system again this weekend 😭

aaronleopold:
https://github.com/redhat-actions/buildah-build?tab=readme-ov-file <-- viable

aaronleopold:
> So I have a suspicion that `podman buildx`, when used in this way for building multi-arch images, will overwrite the tag for whichever build happens last (since it is not parallel builds). If this is true, then the next nightly push will only have an `arm64` build on dockerhub

Yup, it did just that: https://hub.docker.com/layers/aaronleopold/stump/nightly/images/sha256-2633f62ee8a4ed8fa3016cbfd441229e7eddc0fc2a1667181fd2b67297c1537e?context=explore

PR #425, opened 2024-08-30
:construction_worker: Use `buildah` in docker CI workflows

aaronleopold:
See https://github.com/stumpapp/stump/pull/424


aaronleopold:
I've ran these changes a few times on each runner, pushing arm/amd platforms successfully. So I think these changes are good 👍

PR #431, opened 2024-09-01
✨ Support full windows path navigation

JMicheli:
This pull request adds support for navigating to Windows drives other than the `C:\` drive by making use of several (unsafe - but I followed the documentation and linked it in the comments) syscalls to collect the available drives and their names,

Here is what would be seen previously on a Windows machine (note the disabled back button):

image

This is what you see now:

image

The only lingering issue is that when going back to the "root" view from a drive, the path shown in the `Input` component doesn't seem to revert to a blank. If I can work this out, I'll fix it (though I welcome suggestions - my frontend is weak).


aaronleopold (commented):
I think this looks great! I'll try to find time to test it this week on my laptop to ensure it doesn't regress on unix systems.

> The only lingering issue is that when going back to the "root" view from a drive, the path shown in the Input component doesn't seem to revert to a blank. If I can work this out, I'll fix it (though I welcome suggestions - my frontend is weak).

So this is one of the older bits of mostly unchanged code in the repo, so offhand I can't say I'm familiar with how I originally put it together. I'll take a look when I load these changes up locally as I mentioned above. A guess could be the input needs to have an empty string value instead of `undefined` as I [currently have it](https://github.com/stumpapp/stump/blob/main/packages/browser/src/components/DirectoryPickerModal.tsx#L85)

aaronleopold:
Hey, wanted to check in and say I haven't forgotten about this. I've been caught up in the performance bug/issues and haven't had a chance to take a look yet. I'll aim to give it a proper review over the weekend

aaronleopold:
> A guess could be the input needs to have an empty string value instead of undefined as I [currently have it](https://github.com/stumpapp/stump/blob/main/packages/browser/src/components/DirectoryPickerModal.tsx#L85)

This was it 🙂

aaronleopold (approved):
LGTM, thank you!

PR #433, opened 2024-09-02
⬆️ Update dependencies

JMicheli:
This pull request updates a few of the dependencies to their latest version, just a little housekeeping.

The only substantive change is that the `pdfium_render` crate's latest version changed its interface so that you now have to import everything from the prelude module, e.g.,: `use pdfium_render::prelude::{PdfRenderConfig, Pdfium}`. Weird choice imo.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,9 +176,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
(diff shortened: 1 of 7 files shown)


aaronleopold (approved):
Thanks for doing this!

PR #441, opened 2024-09-10
🖥️ Desktop app server management

aaronleopold:
This PR gives the desktop app some love by implementing the following:

- Update `tauri` et al to latest v1 (I am not going to do a v2 migration yet)
- Implement a persisted key/value store on the tauri side, primarily for saved server functionality
- Refactor a lot of the existing rust code for the desktop app for safety (remove `unwraps`, add errors, etc)
- Add type generation for desktop-specific types
- Implement management suite for saved servers on the UI
    - Interact with the `useTauriStore` hook
    - Create / edit / delete / delete all
    - Switch to server (logout + change base URL)

Opening the PR early mostly to start testing the CI to ensure everything compiles okay. Before merge, I will add:

- [x] Select server screen when no server is set but the store has servers configured
- [x] Add tests to new components
    - I didn't add for each, I have limited time unfortunately. But I tried to add tests for the important ones
- [x] Ensure I didn't break anything on the webapp side of things

I want to note that these changes have implications, primarily that I am leaning towards always bundling the server with the desktop app but having the enablement a user setting. This is mostly because I don't necessarily want to have to build two versions of the app: one with the server and another without. I am leaving it feature flagged in the code in case I change my mind, I'd like to see how much bloat it realistically adds.


aaronleopold:
I did some additional testing, I think this is ready enough considering the desktop app is not part of the releases yet. There are definitely many things to work through before starting to include it in the releases, and I'll make a separate issue to record them

PR #442, opened 2024-09-13
⚡ Refactor thumbnail generation and scanner IO operations (#426)

aaronleopold:
This PR updates develop with the latest in experimental. There is only one important commit, so I plan on rebasing experimental afterwards. This will bring in the thumbnail generation and scanner refactors.

I want to reiterate the warning from the original PR:

> ⚠️ There are potentially breaking changes in this PR from the restructuring of library options. The migration was hand written, and while I tested it on my personal experimental instance without issue I am still weary enough to put out this disclaimer.

I am aiming to merge this in over the weekend, and I'll try to test the update path (nightly -> this) beforehand


aaronleopold:
Okay so I've tested the migration, it's definitely "breaking" changes. Running the migration in isolation doesn't really break anything, but when Stump orchestrates the migration for you for whatever reason it removes all of the media from the DB. I assume there is a foreign key constraint when shuffling things around which causes the library series to cascade, and thereby the media cascading. Since scanning is a fairly quick operation, I think this is acceptable TBH, especially with all of the WIP warnings contained in this repo. However, I will not make a habit of allowing this kind of change.

The only feature which I expect there might be some series DB breaks like this would be audiobook support at whatever point that is even feasible to start

PR #449, opened 2024-09-18
♻️ Refactor upload image and add max_size

JMicheli:
As I was reading through the code while preparing to eventually handle the file upload/deletion/bulk management features, I happened on some code for uploading images with a todo marked asking to eventually add configurable max size.

That was easy enough, so I added it as a configuration variable with a default value equal to the prior 20 MB hard-coded limit. I also took the opportunity to refactor the code in the `upload` module so that it can be more easily reused when other upload types are added.

### DIFF
--- a/apps/server/src/utils/upload.rs
+++ b/apps/server/src/utils/upload.rs
@@ -119,19 +119,11 @@ async fn load_field_up_to_size(
 /// Formats the validation errors used elsewhere in this module when a type doesn't
 /// match the expected type (optionally specified as `expected_type_name`).
 fn validation_err(expected_type_name: Option<&str>) -> APIError {
-	if let Some(name) = expected_type_name {
-		if !name.is_empty() {
-			// This is a bit extra, should be guaranteed to succeed if name.len >= 1.
-			let first_char = name.chars().next().expect("name should not be empty");
-			let article = if ['a', 'e', 'i', 'o', 'u'].contains(&first_char) {
-				"an"
-			} else {
-				"a"
-			};
-
+	if let Some(type_name) = expected_type_name {
+		if !type_name.is_empty() {
 			return APIError::BadRequest(format!(
-				"Uploaded file is not {} {}",
-				article, name
+				"Uploaded file was expected to be {}",
+				type_name
 			));
 		}
 	}
--- a/apps/server/src/utils/upload.rs
+++ b/apps/server/src/utils/upload.rs
@@ -149,3 +149,21 @@ fn max_size_err(max_size: usize, name: &str, actual_size: usize) -> APIError {
 }
 
 // TODO: validate_media_upload (books)
+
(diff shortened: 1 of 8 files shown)


apps/server/src/utils/upload.rs
```diff
@@ -1,59 +1,169 @@
...
 
-	tracing::trace!(?content_type, file_size, "Validated image upload");
-
+	tracing::trace!(?content_type, file_size, "Validated upload");
 	Ok((content_type, bytes.to_vec()))
 }
 
+/// Load up to `max_size` bytes of a field (erroring if `max_size` is exceeded).
+/// Returns the loaded bytes as [Vec<u8>] and the total bytes as [usize].
+async fn load_field_up_to_size(
```
aaronleopold, line 84: I'll have to double check the docs tomorrow, but I believe the `DefaultBodyLimit` middleware will enforce the limit for us. I'm not opposed to having extra safeguards to enforce it, but wanted to at least note it

JMicheli: My thinking here was:

1) It seems better for the max_size to be enforced when this function is called, in case it is ever reused for something else in the future. I don't think the overhead is too notable, and theoretically many of these lines shouldn't run at all since the `DefaultBodyLimit` should immediately catch the request and return an error without running the body.
2) Looking at this documentation for stream size info, it looks like it's possible to send requests that are structured to deliver a large payload without signaling the size accurately in the headers. I don't know this for sure, but it might also prevent `DefaultBodyLimit` from catching the request. In this case, these lines should prevent users from downloading a massive image despite their set limits.

aaronleopold: Yeah like I said I'm not opposed to the extra safety, I agree that they shouldn't even hit for files which are too large anyways if the middleware does its job

> it looks like it's possible to send requests that are structured to deliver a large payload without signaling the size accurately in the headers

👀 I'd be curious to see if we can breal/trick the middleware at some point

aaronleopold (approved):
This looks great! There were a couple of my old `TODO`s which could be cleaned up I think, but otherwise this looks good to me

apps/server/src/utils/upload.rs
```diff
@@ -1,59 +1,169 @@
...
+fn validation_err(expected_type_name: Option<&str>) -> APIError {
+	if let Some(name) = expected_type_name {
+		if !name.is_empty() {
+			// This is a bit extra, should be guaranteed to succeed if name.len >= 1.
+			let first_char = name.chars().next().expect("name should not be empty");
+			let article = if ['a', 'e', 'i', 'o', 'u'].contains(&first_char) {
+				"an"
+			} else {
+				"a"
+			};
```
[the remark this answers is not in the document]
JMicheli: Agreed, it's a little much. I'm simplifying the logic.

PR #453, opened 2024-09-19
:sparkles: Add image reader experience options

### LINKED ISSUE #384: [FEATURE] Distinguish reading preferences between mangas and comic books (left vs right).

Being able to distinguish defaults for mangas and comic books. Both are in CBZ/CBR format, but they are not read in the same direction. 

So I'd like to be able to set those parameters for a specific library (manga vs comic).

Modifies: 

- reading direction 
- if double spread, make sure that the correct image is displayed left and right
- maybe add some ui design nudge to help along a user to remember to go in the right direction if unfamiliar
    - progress bar going left or right ? 

**Describe alternatives you've considered**
Kavita tries to do this, and I'll go into details later why it doesn't work imo. 

**Additional context**
Probably either a user preference, with a default that the admin can set server wide or
admin preference only.

### LINKED ISSUE #391: [FEATURE] PDF/CBZ/CBR custom reader display double spread pages
The pdf/cbr/cbz reader doesn't have an option to display in double spread, or two pages side by side. Mentioned in #384 

- Add an option somewhere to have double spreads
- also define the read direction, left ← right for manga or right → left for western comics. 
     - the pages view at the bottom might need to mirror this, by scrolling in the right direction.
- add subtil hint of the read direction for people not used to reading certain mediums. Could be a dark to light gradient in the background, or a border color ? Needs designing.  
- add an option to set the read direction per library, if you organize per medium. 

**Describe alternatives you've considered**
n.a.
maybe just having the admin options and not let the users choose ? 

**Additional context**
kavita sort of has this, but it never worked well...

### LINKED ISSUE #437: [BUG] Image scaling makes long strip comics unreadable
**Describe the bug**
The bug varies by device type / reader mode:
- on desktop in both continuous and paged mode each image ("page") is scaled to the visual height of the browser, which makes each image smaller in width the longer it is
- on mobile in continuous mode the bug looks like on the desktop version
- on mobile in paged mode each page image is stretched to the view, so the longer the image, the more squashed it is

**To Reproduce**

1. Add a long strip comic to the application
2. Open the comic on an affected device
3. Go to the long strip page and witness the beforementioned behavior

**Expected behavior**
Readable long strips

**Screenshots**
example
(I pixelized the images a bit but it's still visible I hope; the left shows desktop browser with raw images of same width [you can see the dimensions of the pictures on two opened tabs] and the right is the squashed mobile paged mode view)

**OS**
Windows, Linux, iOS, Android

**Additional context**
To specify I encounter this while reading manhwa/manhua which are almost exclusively long stripped. I'm not sure if there are other popular media types which use such form (and if continuous mode was even made with this purpose in mind).

Not sure what the best approach would be, but since long strip comics are propably only read with continuous mode, making it so images in that view are scaled to the device width with height on auto on mobile and some sensible value for max width (there is a grayed out viewport option in appearance tab for this i guess?) on desktop I think would be the best solution.

aaronleopold:
Resolves #384
Resolves #391
Resolves #437

This PR implements a few image-based reader features:

1. View two images at once (double spread)
2. Specify the scaling mode for the images (height, width, none)
3. Support right-to-left reading
4. Horizontal scroll reading mode (not an ideal UX yet, IMO)

You will be able to configure defaults for the reading experience on a per-library basis, and override any individual setting per-book.

> ⚠️ I've only implemented the first and third features for the paged reader. This change-set is rather large, so I'd rather tackle adding them to the horizontal scroll reader at another time

I've sneaked in a few other changes, as well: 

- Tweaked the contrast for some of the new themes
- Add optional gradient support for themes
- Refactored all of the tailwind configuration to use TypeScript
- A few others I can't remember

There are some cleanup tasks remaining, and I'd like to add the corresponding library configuration I mentioned above in the settings pages, and then this should be ready to go. I'm aiming for sometime next week


aaronleopold:
Alright, I think this is almost ready to go. There are a few things I want to highlight as things I'll have to either improve or fix as follow-ups:

1. The reader isn't aware when the viewport is too small for double spread. I definitely will add this kind of detection logic, but the work around for the time being is... don't do that lol
2. When double spread is active, highlight both pages in the image previews (bottom bar)

There is one other issue that I'd like to tackle before merge where the bottom bar preview is too "strict" and will scroll you back as you try to navigate ahead. Otherwise, I'm anxious to get this out for further testing. My tentative plan moving forward is:

- Wrap up the sdk rewrite
- Probably release after this change-set sits for a week in nightly
- Either flesh out smart list UI (e.g. actually support creating a list outside the API) OR start bulk management features
- Go back to reader enhancement follow-ups
- I'm itching to take a "break" and work on the mobile app, but we'l see. There is a lot of other important bits to tackle

aaronleopold:
> There is one other issue that I'd like to tackle before merge where the bottom bar preview is too "strict" and will scroll you back as you try to navigate ahead

I've fixed this for the most part, so I am going to merge this in shortly. I don't think the UX is perfect TBH, but am eager to get it out there and tested

Issue #246, opened 2024-01-20
[FEATURE] Authenticate via LDAP, OAuth2, SAML etc.


aaronleopold:
### Discussed in https://github.com/stumpapp/stump/discussions/242

<div type='discussions-op-text'>

<sup>Originally posted by **nk44-official** January 17, 2024</sup>
Hey,

Are there plans to implement LDAP or other SSO Solutions in the future? :)</div>

paulcsiki:
OIDC would fit very nice in my home lab setup.

aaronleopold:
Found this recently, will be a likely candidate for implementation: https://github.com/pfzetto/axum-oidc

nk44-official:
> Found this recently, will be a likely candidate for implementation: https://github.com/pfzetto/axum-oidc

that would be wonderful!

aaronleopold:
For those following this issue, we are getting closer to a point where I can work on this now that #219 is complete. My current **rough** idea is that there would be some new configuration added to define your oidc / sso provider which "unlocks" different authentication endpoints on the API and login methods on the UI. If you have any thoughts on this, please let me know.

Otherwise, stay tuned 🙂

nk44-official:
that sounds awesome!

wreid23:
Please use PocketID as a good easy test subject to connect with once you reach the testing phase not sure where we are at this moment but exciting news this is a very cool project and getting OIDC working is a must nowadays especially with so many good "easy" middlewares today. The code from https://github.com/9p4/jellyfin-plugin-sso might also help give you some good ideas on how to keep it simple as well.

b3nd1k:
Any updates on this?

aaronleopold:
> Any updates on this?

No updates. My time and focus has been absorbed by the massive rewrite effort. Just compare the `experimental` branch with `main` 😅 

I would love contributions if folks want it sooner, otherwise it will come sometime after the rewrite stabilizes a bit

aaronleopold:
> > Any updates on this?
> 
> No updates.

I decided to [work on this a bit](https://github.com/stumpapp/stump/tree/oidc) to better gauge the effort. I've used PocketID for testing (thank you, @wreid23) and so far session-based auth seems functional. There are lots of kinks I would need to work through, and getting this functional using JWTs for the mobile app will be a bit involved, but I'll try and get something out to an `unstable` image by January

aaronleopold:
[...]

PR #457, opened 2024-09-28
:recycle: Rewrite `api` package as standalone `sdk`

aaronleopold:
This may seem insignificant or unnecessary on the surface, but this PR is really important for things down the road. The two big things this paves the way for are:

- Optional JWT/token-based auth (e.g. https://github.com/stumpapp/stump/issues/246)
- A standalone TypeScript SDK which allows much easier 3rd-party integrations to eventually be created

The built-in web app and desktop app (WIP) will continue to use sessions, the mobile app however will now be able to swap to tokens. It's as simple as `new Api(baseURL, 'tokens')`. I will eventually add OPDS V2 to this new SDK, as well. 

> ℹ️ I want to note that while I did structure this package to be standalone, I don't have plans on publishing it to something like `npm`. If there winds up being an actual demand, sure I'll add some workflow to publish it. Otherwise, the effort isn't worth it if the built-in apps are the only consumers 😅 

Super excited to see what this refactor allows down the road


aaronleopold:
I'm going to merge this to `experimental` knowing that there is almost certainly some things which I've likely missed. I'm working on improved filtering and sorting to include relations and aggregates, and want to minimize conflicts

PR #462, opened 2024-10-05
🌐 Update translations

aaronleopold:


aaronleopold:
It looks like some translations are still trickling in, so I'll give it until tomorrow before merge. This will tentatively be the last change to land in dev before I cut the next release this week

PR #464, opened 2024-10-06
:sparkles: Smart list filter builder

### LINKED ISSUE #174: [FEATURE] Smart Lists (IE; ComicRack)
Nope.

The one thing every current reading server is missing is ability to create smart lists like ComicRack can. This is a feature that I know is desperately needed by a lot of users before being able to move away from the program. 

The ability to sort and filter based on ANY metadata at all leads to being able to create very unique lists and makes researching a topic so much easier. Being able to make up lists like this leads to infinite possibilities:

I know this would be a lot of work setting up all the relationships, but it's not impossible as this 14 year old program has shown.

A good way to think about it overall is: Assuming the user has a perfectly curated library and every file has metadata, _you should be able to have all the files in a single folder and it wouldn't matter_. You could create rich views and have limitless possibilities just off the metadata regardless of the location of the files.

https://discord.com/channels/972593831172272148/1161355142877163641

aaronleopold:
Resolves #174 

> ℹ️ There are a few things in the linked issue which this PR does not cover, e.g. virtualization or sorting options baked into the config, but to trim down the project board I am marking this PR as resolving the issue. If people want this feature extended for x, y, or z, we can make a smaller-scoped feature request

I'm super excited about this one. This PR adds a proper filter builder to the UI so you can construct smart lists without having to interact with the API directly. A few other highlights besides the builder itself:

- Stepped form pattern for creating smart lists
- Rewrite schema for the create/update form(s)
- Rewrite the settings to align with other settings pages (see image below)
- Almost full translation coverage for smart list create/update pages/sections 

There are a lot of changes here, though, so rather than enumerating them all I think a picture is worth a thousand words:

image

There are a few remaining tasks before this can go to experimental, the big one being that numeric fields are 50/50 for functionality. E.g. filtering a number field by applying a range. I'm going to prioritize getting `0.0.7` out this week, so this might not land in `experimental` until the following week.


aaronleopold:
I believe I've sorted all of the numeric issues. I've been thinking about it and I think I might just merge this in tonight, I'm excited to get it out and tested. It will also be a good buffer to have it in experimental for a few days, release `0.0.7`, then promote it to nightly shortly after. So, once CI is ✅ I'll merge this

I am **certain** there are multiple bugs here and there, it is still an experimental feature with rough edges and that should be expected. FYI @JMicheli, if you were planning on updating the macros and pushing to this feature branch it can just be a follow-up instead

PR #466, opened 2024-10-07
📝 Add OPDS testing reviews for Foliate, Thorium, Cantook and Librera

n1ghthouse:
Adds more supported clients and their issues to the OPDS tested clients section.


aaronleopold (approved):
Thanks for taking the time to add your testing and experiences!

PR #469, opened 2024-10-09
🚨 Run clippy pedantic

JMicheli:
This is a housekeeping PR. I went into the roots of the core and server crates, enabled clippy::pedantic, and fixed anything that seemed sensible to fix. I think the result is nice, lots of little readability and consistency fixes. Plus it was relaxing.

There were even a few instances where parameters in functions could become references, which could theoretically improve performance but won't because it was like 4 string manipulation functions.

Passes tests and runs fine for me after as well.


aaronleopold (approved):
This is awesome, thank you! I glossed through the change-set, but I'll trust clippy on this one

> Plus it was relaxing

I love this, I also enjoy clippy doing its thing

PR #472, opened 2024-10-11
✅ Resurrect core integration tests

JMicheli:
I'm starting this as a draft because I think there's work to do, but I want to see codecov's results for it.

This moves the contents of the old, unused core/integration-tests to the normal path for integration tests `core/tests` and adds a couple of actual integration tests.

I think the tests could be expanded and better-written. I tried to translate some other old tests but it seems the core API has changed enough that it wasn't obvious how they should be re-written.


core/package.json
```diff
@@ -10,7 +10,7 @@
 		"studio": "yarn dlx prisma studio",
 		"build": "cargo build --release",
 		"format": "cargo fmt --package stump_core",
-		"integration-tests": "cargo integration-tests",
+		"run-tests": "cargo test",
```
aaronleopold, line 13: I think we can just remove this tbh, thanks for updating it though!

core/tests/scanner_tests.rs
```diff
@@ -0,0 +1,104 @@
...
+			ignore_rules: IgnoreRules::new(vec![]).unwrap(),
+			library_id: None,
+		},
+	)
+	.await;
+
+	ctx.enqueue_job(LibraryScanJob::new(lib_data.id, lib_data.path))
+		.unwrap();
+	/// TODO - something less heinous, there must be a way to wait on a job
+	tokio::time::sleep(Duration::from_secs(1)).await;
```
aaronleopold, line 44: Haha I know we chatted on Discord about this, but to reiterate here I think `tokio::select!`ing on the event bus and a reasonable timeout is the way to go here

core/tests/scanner_tests.rs
```diff
@@ -0,0 +1,104 @@
...
+		.media()
+		.find_many(vec![media::extension::in_vec(vec![
+			"zip".to_string(),
+			"cbz".to_string(),
+		])])
+		.exec()
+		.await
+		.unwrap();
+
+	assert_ne!(items.len(), 0);
```
aaronleopold, line 57: We can probably assert the actual expected number of items, since we know(?) how many should be scanned when we create the test library

core/tests/utils/temp_core.rs
```diff
@@ -0,0 +1,58 @@
...
+	let config_dir = temp_dir.path().to_string_lossy().to_string();
+	let config = StumpCore::init_config(config_dir).unwrap();
+	let core = StumpCore::new(config).await;
+
+	let migration_res = core.run_migrations().await;
+	assert!(
+		migration_res.is_ok(),
+		"Failed to run migrations: {:?}",
+		migration_res
+	);
```
aaronleopold, line 17: Just a note that IIRC this won't actually run migrations in debug/test. I _think_ it will just push onto the DB. I'd have to double check to be sure. Either way, it is needed

aaronleopold (commented):
I gave this a read through, thanks for getting them more up-to-speed with the current API and long backlog of changes which happened since I removed them from the workspace!

> I think the tests could be expanded and better-written. I tried to translate some other old tests but it seems the core API has changed enough that it wasn't obvious how they should be re-written.

I totally agree, IIRC the scanner + job system itself has had 1-2 rewrites since these tests were written. Happy to collaborate as needed to think through what the best system moving forward might be 🙂

PR #481, opened 2024-10-16
🐛 Fix bug in `FileExplorer` that broke clicking on books.

JMicheli:
I noticed an issue in the `FileExplorer` component that seems to have followed on the sdk overhaul. The function `getBook` defined here requires an `sdk: Api` parameter:

https://github.com/stumpapp/stump/blob/2409a6e04371af8ba2421154da96a0588ff1811a/packages/browser/src/components/explorer/FileThumbnail.tsx#L112-L134

However, a few calls to `getBook` didn't include the parameter, which caused them to fail. As a result, when double clicking a file in the explorer, you'd get this response:

image

The changes I made seem to fix it, but it would be good to have review by someone more familiar with React to make sure that any other changes that I may have missed are made.

### DIFF
--- a/packages/browser/src/components/explorer/FileExplorerProvider.tsx
+++ b/packages/browser/src/components/explorer/FileExplorerProvider.tsx
@@ -1,4 +1,4 @@
-import { useDirectoryListing } from '@stump/client'
+import { useDirectoryListing, useSDK } from '@stump/client'
 import { DirectoryListingFile } from '@stump/sdk'
 import React, { useState } from 'react'
 import toast from 'react-hot-toast'
@@ -22,6 +22,7 @@ type Props = {
 export default function FileExplorerProvider({ rootPath }: Props) {
 	const navigate = useNavigate()
 	const isMobile = useMediaMatch('(max-width: 768px)')
+	const { sdk } = useSDK()
 
 	const [layout, setLayout] = useState<ExplorerLayout>(() => getDefaultLayout())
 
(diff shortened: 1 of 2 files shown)


packages/browser/src/components/explorer/FileExplorerProvider.tsx
```diff
@@ -39,7 +40,7 @@ export default function FileExplorerProvider({ rootPath }: Props) {
 			setPath(entry.path)
 		} else {
 			try {
-				const entity = await getBook(entry.path)
+				const entity = await getBook(entry.path, sdk)
```
aaronleopold, line 43: I wonder why the lints didn't catch this. I'll have to revisit the frontend lint scripts

aaronleopold (approved):
This looks perfect to me. Thanks for finding and fixing this issue!!

packages/browser/src/components/explorer/FileExplorerProvider.tsx
JMicheli: That's a good point, I found the problem easily because ESLint threw an error in VS Code and underlined it in red for me.

JMicheli: Well actually here's your answer: https://github.com/typescript-eslint/typescript-eslint/issues/352#issuecomment-472597115

The salient part:
> we purposely do not care about errors that should arise as a result of type checking errors.
We only care that the code is syntactically valid (i.e. can be parsed without errors), not about whether it's semantically valid (i.e. don't care if it is type sound).

We need to use `tsc` to catch this sort of thing.

aaronleopold: Ahh yeah, okay. That makes sense. I could have sworn I defined a tsc script somewhere to check types but I'll take a look tomorrow

PR #484, opened 2024-10-18
:rotating_light: Wrap large enum variants in `Box`

JMicheli:
The new version of rust comes with a new clippy lint that triggers on enums with large sizes. Enums have a size equal to their largest contained variant.

The `CoreError` enum came in at 144 bytes as the result of containing `prisma_client_rust::QueryError`, which is huge. This is easily fixed by boxing the type, which shut clippy up.

There are some other big variants still, but nothing over 100 bytes.


aaronleopold (approved):
Thank you!

PR #491, opened 2024-10-25
:bug: Exclude cover image from double spread and fix rtl

smldub:
- The first page is the cover, so it should be shown by itself instead of in double spread (maybe this could be optional)
   - requires skip variable to make it possible to go back to page 1 (else it will try to subtract two pages from page 2)
 - Reading right to left vs left to right requires flipping the ordering of the pages
 

(never coded a line of javascript before tonight, so sorry if it the code is awful...)


aaronleopold:
Thanks so much for fixing this! I'm traveling for work but will give it a proper look once I return

aaronleopold (approved):

PR #497, opened 2024-11-03
🐛 Fix unresponsive app after closing modal within dropdown

xendke:
I noticed there is an issue where if you open a dropdown menu and one of the items opens a modal, once the modal is closed the app becomes unresponsive to clicks until you refresh the page.

https://github.com/user-attachments/assets/658aeaa0-72ee-4c54-ae06-b4445691c4ca

Here is the [radix-ui issue](https://github.com/radix-ui/primitives/issues/837) that talks about it. Although it is marked as fixed, some folks are still running into this issue.

This PR fixes the issue but with the drawback that you can interact with the outside of the dropdown menu while it's open.
I can keep looking into it if we think that's worth it but regardless this is probably a better UX than what is happening right now.


aaronleopold (approved):
> This PR fixes the issue but with the drawback that you can interact with the outside of the dropdown menu while it's open

I assume that means you can't click out of it, either? Regardless I agree, though,a freeze is definitely not the more ideal UX. This should be more than acceptable for now.

Thanks for finding, outlining, and fixing exactly what was wrong - I really appreciate it!

xendke:
when you click outside of the dropdown it does close but that same click that closes the menu will interact with whatever you clicked on, here's a before and after:

https://github.com/user-attachments/assets/defa15cd-7f1e-4b0e-b995-042e07cf2980

with fix:

https://github.com/user-attachments/assets/7dbc5c2a-d021-40c8-99ee-11cca367e573


thanks your attention on this tiny fix! feel free to merge 🙇

PR #499, opened 2024-11-04
🎨 Server housekeeping

JMicheli:
This pull request just does a couple of things:

1. It boxes the `QueryError` in `APIError`, which silences a clippy warning.
2. It walks back the silly each-item-individually change I made in routers::api::v1 for exported values. That was messy.


aaronleopold (approved):
LGTM. Thank you!!

PR #501, opened 2024-11-07
🐛 Fix popover and dropdown on mobile sidebar

xendke:
I noticed my [previous PR](https://github.com/stumpapp/stump/pull/497) introduced an issue on mobile viewports 😱  :

https://github.com/user-attachments/assets/f9b7beac-308b-4267-a198-adf98f70b65b

This PR fixes the issue! I took some time to look at the other uses of `<DropdownMenu/>` to make sure they work as expected in mobile and desktop viewports. Also fixed UserMenu's Popover which was having the same issue as shown above.

https://github.com/user-attachments/assets/763dbf6c-a72d-442c-9a50-e23a759d77fa

------

Separately, I noticed some small UI issues that only happened if the viewport was exactly `768px` so I took care of those as well:
- missing header and footer in sidebar
- library settings header padding offset 
<details><summary>Screenshots</summary>
<p>

#### After

</p>
</details>

### DIFF
--- a/packages/browser/src/components/UserMenu.tsx
+++ b/packages/browser/src/components/UserMenu.tsx
@@ -24,7 +24,7 @@ export default function UserMenu({ variant = 'sidebar' }: Props) {
 	return (
 		<AutoSizer style={{ height: '2.35rem', width: isSidebar ? '100%' : '2.35rem' }}>
 			{({ width }) => (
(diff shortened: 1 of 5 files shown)


aaronleopold (approved):
I pulled these changes down and didn't see anything out of place, so looks good to me!

Thanks again for your contribution and for the detailed description! 🙂

Issue #110, opened 2023-03-27
[FEATURE] Delete book + Optional delete file 


dancamdev:
I've seen most of the ebook servers out there do not support this feature, not sure why, but I feel it makes a lot of sense to be able to manage the library nicely.

**Describe the solution you'd like**
We should allow for deleting books from the library and have a dialog asking wether we want to delete the actual file on the file system as well.

aaronleopold:
There were similar discussions a while back about whether or not the server should have that kind of power, and I'm having trouble remembering the resolution. 

That being said, there is currently logic that will attempt to 'trash' RAR files that have been successfully converted. This will obviously fail in docker, as the crate I use for that is not 'docker-aware,' so to speak. I'm personally not against adding some utilities in the stump_core package to generalize trashing files per user request, so long as it's trashing them and not completely, hard deleting them.

Issue #336, opened 2024-05-08
[BUG] Missing a delete button ?


Leiasticot:
**Describe the bug**
I didn't have any button to delete a file, and when I delete it from then files still be inside my libraries. Even inside Settings.

**To Reproduce**

1. .cbz from a webtoon
2. delete files from the server with rm -r
3. Inside the library

**Expected behavior**
When I delete a file inside my server then it is deleted from the library (or at least use a button to not see it always).

**Screenshots**
![image](https://github.com/stumpapp/stump/assets/50156473/4748a00e-3ac3-4f97-bc3c-8c4df5022903)
![image](https://github.com/stumpapp/stump/assets/50156473/f02b6f44-b4f0-4c2f-a95e-21aae87717d4)
![image](https://github.com/stumpapp/stump/assets/50156473/9893760a-18f5-4bd9-a9a8-4116a355fd8d)
![image](https://github.com/stumpapp/stump/assets/50156473/43193426-0a0e-4ee3-9a28-3a7f753fc519)

**Desktop (please complete the following information):**

- OS: Docker on debian 12 for server, and client is Windows 11 Chrome

Cantello:
One related question here: When I rename a book/series after it has been scanned, I have duplicates in my library. Is there any way to remove the duplicates or the `missing` entities you mentioned above?

Cantello:
Thanks, I don't know how I missed this when I was looking...

aaronleopold:
> Thanks, I don't know how I missed this when I was looking...

Depending on what version you are trying out, it can be easy to miss. The latest builds are just a long scrollable page, but I've been experimenting with more segmentation in nightly/experimental images and have this housed in a `Deletion` section in a secondary navigation menu in library settings

PR #502, opened 2024-11-07
✨ Add delete media functionality

JMicheli:
This pull request addresses #110 by adding the requested feature.

I added a new endpoint to the api::v1::media::individual module that allows deletion of files by users with the `ManageLibrary` permission. The endpoint takes an optional query parameter to allow deletion of the underlying file.

For the UI, I added a button to the management screen for deleting books. Clicking it opens a confirmation modal that allows a user to check a box to delete the underlying file.


aaronleopold:
There are two big things I'd like to touch on before digging into this feature: Deletion from the perspective of the DB vs deletion of files directly

I'll create explicit sections to better organize my thoughts

## Deletion (DB)

It isn't actually used today, but media records do have a [`deleted_at` field](https://github.com/stumpapp/stump/blob/main/core/prisma/schema.prisma#L196). There is an issue that, after rereading I realize is a duplicate of the file deletion lol, references soft vs hard deletion: https://github.com/stumpapp/stump/issues/336.

This does have some implications though, primarily that most queries would need to be refactored as to exclude any deleted books by default.

I thought I had more to discuss here, but I'm coming up blank. I think the section below is more relevant to the conversation at hand, and it touches on the tentative interactions between the DB and file deletion. I'll leave this here for the useful links and move on

## Deletion (files)

Originally, in the referenced issue, I expressed the preference for some kind of "trash" functionality:

> I'm personally not against adding some utilities in the stump_core package to generalize trashing files per user request, so long as it's trashing them and not completely, hard deleting them

I think this might be a more palatable first step into the deletion realm for users. Programmatic deletion is generally non-recoverable unless someone has their own backup system, which is somewhat scary. The elephant in the room is that "trashing" a file in Docker doesn't necessarily make sense. I think a viable route would be to implement some kind of custom trash system where:

- When in docker, move files to a mounted `/trash` path. Develop an API for emptying the trash, which does the full file removal
  - This extra barrier to deletion IMO is a good, inherent safety net
- When outside docker, use the existing `trash` crate to push things into the native trash for the operating system

I also think we need to add something like `file:delete` to the user permission enum. This functionality extends beyond just managing a library IMO.

---

OK. Rant ended. All of this to say, I'm not entirely sure what the best combination of these deletion strategies is. A potentially helpful use-case to consider could be:

- A user reads a book, decides they don't like it, soft-deletes it from the library which moves it to the trash and sets the `deleted_at` field
- A few weeks later, a different user wants the book. The original user can restore the book from the trash, which moves it back to the library and unsets the `deleted_at` field. Since the file was never actually deleted, the book is still available for the second user. Additionally, since the record was never deleted, the reading history is still intact
- If the second point never happened, the trash could be emptied instead. The file would be deleted, the record dropped, and any relations (e.g. reading progress) should automatically cascade and get deleted.

I can't enumerate all the potential use-cases, and this is just the first I could think of, but I think this is a good starting point for how I see it in my head at least. I'm okay making broad strokes / decisions here since, generally speaking, if people _are_ actually using Stump they aren't very vocal about what they want vs definitely don't want.

Once we land on a concrete path forward, I think we can move the more granular and implementation-specific conversation to Discord for convenience. But please let me know if you have any thoughts or differing opinions! I'm happy to discuss further

JMicheli:
Okay so the `deleted_at` part makes sense to me, and adding a trash space should be easy using that. I do think there are two use cases not discussed above:

1. The user made a mistake when structuring their library and wants to delete a book from the database, then immediately rescan the library to get things right. This would be a situation where the trash option probably makes sense, since they could forget about it and simply rescan
2. The user made a mistake when uploading a book and wants to remove the file immediately and re-upload. I think this is probably something that should be accommodated, since people will make mistakes with the new upload functionality. In this instance, having it go to the trash and only be soft deleted might not make sense.

aaronleopold:
> The user made a mistake when structuring their library and wants to delete a book from the database, then immediately rescan the library to get things right.

They wouldn't have to rescan at all, assuming it is implemented as I described above: 

> This does have some implications though, primarily that most queries would need to be refactored as to exclude any deleted books by default.

So if I were to delete a book this way, the UI should refetch the current query which doesn't include the book anymore. Let me know if I'm not explaining that bit well

> The user made a mistake when uploading a book and wants to remove the file immediately and re-upload

I don't see much difference in this scenario, but I do understand your point. The problem I see is that to accomodate this would mean two separate-ish code paths:

- Delete when the record is "fresh" or "new"
- Delete when the record is "stale" or "old enough"

What would those relative time descriptors be? Should it be X minutes? Would Y minutes be too much? Etc. The UI would also need to present what would happen in the same manner, since there would be an inconsistency in what an action does between these two scenarios.

If we can come up with acceptable definitions for these kinds of things and document them accordingly, I'd be okay with it though 🙂

aaronleopold:
This PR is a bit stale at this point, so I'm going to close it out for now

PR #503, opened 2024-11-08
💄 Add splash animation within web index file

xendke:
Is this something we're interested in? Any concerns or improvements? we could support more themes if needed!

https://github.com/user-attachments/assets/d90c7fa3-8044-444b-8386-7e67652e314a


JMicheli:
I actually think that rocks!

aaronleopold (commented):
I think I can dig it! I left a few comments I'd like to resolve before giving my ✅, but otherwise I think we can give this UI a try

aaronleopold (approved):

PR #505, opened 2024-11-12
♻️ Create Preformatted component to render JSON

xendke:
Noticed `<pre>` being used in a couple of places. This PR introduces a reusable component for those use cases!

Also found that a random modal was not closing in certain cases so I fixed that as well.

Some uses of the new component:

Screenshot 2024-11-11 210033
Screenshot 2024-11-11 210247
Screenshot 2024-11-11 215432


aaronleopold (approved):
Thanks!

Issue #334, opened 2024-05-03
[FEATURE] Subheadings not present on (some?) epubs


LRotenberger:
On the desktop reader in Stump I am unable to see subheadings for certain e-books as seen in the screenshots below.

![image](https://github.com/stumpapp/stump/assets/62265343/a52d8c6c-3cf6-4d21-8218-c8305f2d1e98)

The image above is the table of contents on stump vs. mobile reader Lithium below.

![image](https://github.com/stumpapp/stump/assets/62265343/9f2e3220-7763-4758-ba18-de7f568f5aa5)

PR #521, opened 2024-12-10
✨ render children of an epub content item

xendke:
Some epub content items have more content items inside! This PR exposes them to the reader.

First time working with Rust 😅 - criticism welcome!

#### Before

Screenshot 2024-12-09 203725

#### After

Screenshot 2024-12-09 204133

### DIFF
--- a/core/src/db/entity/epub.rs
+++ b/core/src/db/entity/epub.rs
@@ -14,6 +14,7 @@ use super::{media::Media, MediaAnnotation};
 pub struct EpubContent {
 	label: String,
 	content: PathBuf,
+	children: Vec<EpubContent>,
 	play_order: u32,
 }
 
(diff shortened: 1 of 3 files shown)


aaronleopold:
I'll be able to take a look tomorrow after work to review, but based on your screenshot I actually think this would resolve https://github.com/stumpapp/stump/issues/334! Excited to give it a proper look! Thank you!

core/src/db/entity/epub.rs
```diff
@@ -14,6 +14,7 @@ use super::{media::Media, MediaAnnotation};
 pub struct EpubContent {
 	label: String,
 	content: PathBuf,
+	children: Vec<EpubContent>,
```
aaronleopold, line 17: I'm surprised that clippy didn't complain about this recursive type, but I won't complain about it not complaining about it 😛

aaronleopold (approved):
This looks great! Thanks again!

PR #525, opened 2024-12-15
🐛 Fix bookmark click navigation

xendke:
Bookmarks were actually redirecting to the relevant page and instead navigated to the closest chapter. 

#### Before
https://github.com/user-attachments/assets/65a3719a-5d7d-4a81-9033-30ba8fb070f1

#### After
https://github.com/user-attachments/assets/d921a4ad-b4ba-4c6a-a679-801e1c8f538f


aaronleopold (approved):
Thank you!

PR #528, opened 2024-12-18
✨ Add GET and PUT endpoints for media metadata

aaronleopold:
This was asked on Discord to allow for better scripting capabilities when using the SDK

Note that PUT'ing the metadata requires `library:manage` permission

### DIFF
--- a/apps/server/src/routers/api/v1/media/mod.rs
+++ b/apps/server/src/routers/api/v1/media/mod.rs
@@ -50,6 +50,11 @@ pub(crate) fn mount(app_state: AppState) -> Router<AppState> {
 				.route(
 					"/page/:page/dimensions",
 					get(individual::get_media_page_dimensions),
+				)
+				.route(
+					"/metadata",
+					get(individual::get_media_metadata)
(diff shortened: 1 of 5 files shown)


apps/server/src/routers/api/v1/media/mod.rs
```diff
@@ -50,6 +50,11 @@ pub(crate) fn mount(app_state: AppState) -> Router<AppState> {
 				.route(
 					"/page/:page/dimensions",
 					get(individual::get_media_page_dimensions),
+				)
```
JLesDev, line 53: Consider spacing of lines here

aaronleopold: CI enforces a standard formatting and this PR passed that check. Is there something in particular about the spacing you found issue with? Not sure what you mean.

JLesDev: Ah, I must confess you are right here and I was indeed incorrect. We can mark this as resolved and I will further learn the standard formatting guidelines.

Issue #537, opened 2024-12-26
[BUG] "STUMP_ALLOWED_ORIGINS" wildcard causes infinite restarts


vorpalhex:
```
thread 'main' panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/tower-http-0.5.2/src/cors/allow_origin.rs:55:13:
Wildcard origin (`*`) cannot be passed to `AllowOrigin::list`. Use `AllowOrigin::any()` instead
note
```

This happened because I added `*` as an entry to `STUMP_ALLOWED_ORIGINS` on the docker image. This caused that origin to be written to the Stump.toml, which then resulted in Stump not booting even after it was removed from the env (because it was still in the toml due to how origins are added)

Possible fixes:
1. Just reject/ignore `*` in the STUMP_ALLOWED_ORIGINS. Quick fix that does actually resolve the problem. 
2. Detect `*` in STUMP_ALLOWED_ORIGINS, update logic to be `AllowOrigin::any()`.

JMicheli:
Good catch, I'll get a PR up for this one shortly. I favor solution 2 above at first blush.

aaronleopold:
@JMicheli's fix has made its way into the `nightly` image, so I'll close this issue. If using a wildcard still presents some kind of issue, we can re-open this and figure it out

PR #538, opened 2024-12-26
🐛 Change CORS layer construction to allow wildcard

JMicheli:
This pull request addresses the problem raised in #537, which causes Stump to fail to start when a wildcard ("*") is used in the allowed origins configuration variable.

The cause is that Axum doesn't like wildcards being added via the `AllowOrigin::list()` function and wants you to use `AllowOrigin::any()` instead. I slightly reorganized the server's `get_cors_layer()` function to detect when a "*" is present in the list of allowed origins and exit early with an `AllowOrigin::any()` configuration. 

I'm making the assumption is that there is no case where the allowed origin list would include a wildcard and then other origins but the user would expect behavior other than any origin being allowed. I think that's right.

### DIFF
--- a/apps/server/src/config/cors.rs
+++ b/apps/server/src/config/cors.rs
@@ -32,14 +32,46 @@ fn merge_origins(origins: &[&str], local_origins: Vec<String>) -> Vec<HeaderValu
 pub fn get_cors_layer(config: StumpConfig) -> CorsLayer {
 	let is_debug = config.is_debug();
 
-	let mut allowed_origins = Vec::new();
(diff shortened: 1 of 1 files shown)


aaronleopold (approved):
This looks great, thank you!!

> I'm making the assumption is that there is no case where the allowed origin list would include a wildcard and then other origins but the user would expect behavior other than any origin being allowed. I think that's right.

Yeah I think this is the correct assumption, it doesn't make sense to say "I want any origin, but also this specific one" since it would be covered by the wildcard

PR #550, opened 2025-01-06
:bug: Fix `TemplateRegistrationFailed` error when sending email

### LINKED ISSUE #549: [BUG] Emailer errors out when sending with "TemplateRegistrationFailed"
## Description of the Bug

Registering an smtp server and then attempting to use seems to not work. I don't believe this is a configuration error on my side, but it's possible I'm missing something.

## Expected Behavior

Email/book to be sent

## To Reproduce

1. Register new emailer through UI (I'm using remail)
2. Navigate to any book and attempt to email, either to a device or a manually entered email address, or both
3. Get error in logs instead

## Debugging Information

```
2025-01-06T02:21:38.543847Z ERROR stump_server::routers::api::v1::emailer: Failed to send email, e: TemplateRegistrationFailed(TemplateError { reason: IoError(Os { code: 2, kind: NotFound, message: "No such file or directory" }, "base"), template_name: None, line_no: None, column_no: None, segment: None })
    at apps/server/src/routers/api/v1/emailer.rs:589
```

This is on 0.0.9 on Docker on an Ubuntu host.

aaronleopold:
Resolves #549

I've included the default templates in the binary, something I previously forgot to do. This *should* fix the issue, but I have to do some actual verification before I am comfortable merging this in.

I also took the time over lunch to actually write some documentation for the email feature. It isn't the best, but better than it was (i.e., nothing lol)

### DIFF
--- a/apps/server/src/main.rs
+++ b/apps/server/src/main.rs
@@ -18,10 +18,6 @@ fn debug_setup() {
 		"STUMP_CLIENT_DIR",
 		env!("CARGO_MANIFEST_DIR").to_string() + "/../web/dist",
 	);
-	std::env::set_var(
-		"EMAIL_TEMPLATES_DIR",
-		env!("CARGO_MANIFEST_DIR").to_string() + "/../../crates/email/templates",
-	);
 	std::env::set_var("STUMP_PROFILE", "debug");
 }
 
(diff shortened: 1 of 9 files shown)


aaronleopold:
This will likely be the last experimental commit (after merge) before I promote the change-set to nightly

PR #569, opened 2025-01-24
💄 improve experience with IOS 'add to home screen' PWA

RileyMathews:
Adds a manifest file for progressive web app specification. This should allow users to 'save to home screen' on mobile devices (I was only able to test on iOS) and get a better 'native' feel without browser controls.


RileyMathews:
Closing in favor of branch https://github.com/stumpapp/stump/tree/al/vite-pwa which has a much better implementation of the manifest file being managed by vite itself for better tooling integration.

PR #571, opened 2025-01-25
:construction: Rewrite RN app to use expo router and build out PoC

aaronleopold:
I've rewritten a large chunk of the RN app in order to upgrade some dependencies (e.g., NativeWind) while modernizing a bit (e.g., swap back to expo router)

It isn't really an _app_ yet, but I think it is in a much better spot and looks moderately decent as a proof of concept. I'll aim to clean this up before merge, but it shouldn't have any impacts elsewhere.


aaronleopold:
This is more than good enough to merge to experimental, so I'll do so once CI is ✅. I originally wrote that this would have minimal impact on other areas of the code, but there are some changes besides the massive mobile app rewrite that are worth mentioning:

- Add support for Readium/Cantook progression for OPDS v2
- Fix bug which flipped the dimensions for book page analysis
- Add support for providing a custom `X-Stump-Save-Session` header when using basic auth that will skip creating a session
- Miscellaneous OPDS v2 changes in the `core`

This has been super fun, hopefully the app will be ready for test flight before next year

PR #576, opened 2025-01-30
:bug: Fix path error in koreader router

aaronleopold:
This PR fixes the following error:

```
Wrong number of path arguments for `Path`. Expected 1 but got 2. Note that multiple parameters must be extracted with a tuple `Path<(_, _)>` or a struct `Path<YourParams>`
```

I just did the same fix I did when this happened for OPDS v1.2 after adding API key support to that router. The issue is that I can't force Axum to "ignore" the parent Path params for handlers of nested routers (e.g., `:document`). So when I construct a handler that only has one Path (e.g., `Path(document): Path<String>`) it rejects the request outright with that error.

Ideally, I don't need this work around but I haven't figured out how to achieve that wiht Axum yet


aaronleopold:
I was able to both push and pull progress from a book which has the `koreader_hash` set in the database. So I'm comfortable merging this 👍

PR #582, opened 2025-02-07
:bug: (api) Fix path normalization for epub resource lookup

darb5:
When doing resource lookups via the api, the path normalization for relative links (e.g. "A/../B/0.css") is broken. This fix pops up a directory when ".." is used, and removes a path component if "." is used.

The previous behavior mapped "OEBPS/../Styles/style.css" -> "OEBPS/Styles/style.css", when I think it should be "Styles/style.css".

I did try to use `std::path::absolute` since the mentioned ticket is now marked closed. However it looks like they choose to leave ".." in the path on unix systems. I did consider using it anyways, then fixing just the ".." occurrences. While feasible, I avoided it since it does resolve the whole path (e.g. "OEBPS/../Styles/style.css" -> "/home/user/stump/OEBPS/Styles/style.css". In order to fix that, the current working direct would have to be removed. Again, feasible, but it feels like that is getting messy.

### DIFF
--- a/core/src/filesystem/media/format/epub.rs
+++ b/core/src/filesystem/media/format/epub.rs
@@ -387,6 +387,10 @@ mod tests {
 		let result = normalize_resource_path(path, "OEBPS");
 		assert_eq!(result, PathBuf::from("OEBPS/Styles/style.css"));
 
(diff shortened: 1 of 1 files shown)


aaronleopold (approved):
Thank you!

PR #585, opened 2025-02-10
 🐛 (api) Fix metadata api lookup

darb5:
There is an issue with serde_qs (https://docs.rs/serde_qs/latest/serde_qs/#square-brackets) with non-flattened substructs. Specifically, the way to query them is to use square brackets `metadata[writer]=Sanderson`, but this needs to be url encoded so it turns into `metadata%5Bwriter%5D=Sanderson`. In strict mode, the key gets parsed as the string `"metadata[writer]"` instead of an object and key. In non-strict mode, it will correctly parse it as nested keys.

An example would be `http://localhost:10801/api/v1/media?metadata%5Bwriter%5D=Sanderson`

Another option would be to flatten all the sub-structs. While this is possible, it seems like it might cause naming conflict issues in the future. I do not really see a valid case where the key name would have square brackets.

### DIFF
--- a/apps/server/src/routers/api/v1/media/mod.rs
+++ b/apps/server/src/routers/api/v1/media/mod.rs
@@ -3,11 +3,13 @@ pub(crate) mod individual;
 pub(crate) mod thumbnails;
 
 use axum::{
-	extract::DefaultBodyLimit,
+	extract::{DefaultBodyLimit, Extension},
 	middleware,
 	routing::{get, post, put},
 	Router,
 };
+
+use serde_qs::axum::QsQueryConfig;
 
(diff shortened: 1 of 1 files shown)


darb5:
The first param in the config is `max_depth` and the default value is 5 (https://github.com/samscott89/serde_qs/blob/main/src/de/mod.rs#L93), so I kept it the same as the default.

aaronleopold (approved):
Thank you! Also, I really appreciate the explanations and doc links

> Another option would be to flatten all the sub-structs. While this is possible, it seems like it might cause naming conflict issues in the future. I do not really see a valid case where the key name would have square brackets.

Agreed

PR #588, opened 2025-02-10
♻️ Rework the BookOverviewScene to include metadata information

darb5:
Adds metadata into the book scene header. Each element links to a search page. I moved genres into this section but left the external links where they were.

The inspiration here came from [audibookshelf](https://audiobooks.dev/audiobookshelf/item/1551eb8a-7b16-48a7-9357-16af75d825d8) and audible. Both have a similar header section right below the titles with links to search results. The user/pass for the demo audiobookshelf is demo/demo. 

I also moved out the render header logic into a separate file. I generally try to keep files small, and this seemed link enough logic to move it out.

### DIFF
--- a/packages/browser/src/scenes/book/BookOverviewSceneHeader.tsx
+++ b/packages/browser/src/scenes/book/BookOverviewSceneHeader.tsx
@@ -7,104 +7,104 @@ import { Heading, Text } from '@stump/components'
 import paths from '../../paths'
 
 interface MetadataTableItem {
-	keyname: string
+	keynameBase: string
 	prefix: string
 	values: string[]
 	searchKey: string
 }
 
-export default class BookOverviewSceneHeader {
-	media: Media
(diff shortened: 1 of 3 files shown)


packages/browser/src/scenes/book/BookOverviewSceneHeader.tsx
aaronleopold, line 1: Could you refactor this as a functional component? I feel like we might be approaching an anti-pattern here by making this a kinda-almost class component, which aren't used anywhere else in Stump and generally considered legacy patterns

darb5: Done. I also fixed an issue I had where the key prop was not being set correctly.

darb5:
This was one of the major navigation features I wanted before switching over. I appreciate the fast turn around here, thank you!

PR #597, opened 2025-02-15
:sparkles: Add clickable progress bar for epubs

darby42:
This adds support for a yomu style progress bar for epubs. The progress bar is really a list of progress bars where each bar corresponds to a section in the book.spine. Clicking on the progress bar will jump to that section of the book.

I did consider using toc instead, but things got a little messy since you have to combine sections that are not in the toc. At least a few of my books had no toc. Checking yomu, it seems to also use sections and not the toc.

I tried making it so that it jumps into the section by the same percentage of the bar the user clicked on but could not get that working reliably with epubjs. This would be useful for books that do not have chapters (e.g. convert txt files).

One significant issue here is that it has to load each section to get the text length. A better solution here would be to cache the results or have some metadata per section server side. This might be easier to implement after streaming is supported.

### DIFF
--- a/packages/browser/src/components/readers/epub/EpubJsReader.tsx
+++ b/packages/browser/src/components/readers/epub/EpubJsReader.tsx
@@ -207,9 +207,13 @@ export default function EpubJsReader({ id, initialCfi }: EpubJsReaderProps) {
 			if (book.spine) {
 				const defaultLoc = book.rendition?.location?.start?.cfi
 
+				const boundingClient = ref.current?.getBoundingClientRect()
+				const height = boundingClient?.height ? boundingClient.height - 2 : '100%'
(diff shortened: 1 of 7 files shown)


aaronleopold (changes requested):
This is so awesome, thank you for this! I pulled it down locally and have a few UI-specific adjustments I'd like, mostly to move away from the `ProgressBar` and make it look a bit more like Yomu. I didn't want to push directly to your work here so just made a diff. If preferred, I can just push though.

aaronleopold:
[epub-progress-adjustements.diff.zip](https://github.com/user-attachments/files/18812831/epub-progress-adjustements.diff.zip)

darby42:
applied

aaronleopold (approved):
Thank you!

PR #601, opened 2025-02-16
:bug: Fix fallback checks when looking for cover image

darby42:
The previous fallback had an issue with PathBuf usage. `starts_with` and `ends_with` only consider whole path components [link](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.starts_with), meaning `path.starts_with("cover")` would match `cover/image.png` but not `cover.png`. Similarly `path.ends_with("png")` would match `path/png` but not `path/image.png`. The second issue was that `Cover.png` would not work because the checks were case sensitive.

This change checks the lowercased version of the [file stem](https://doc.rust-lang.org/std/path/struct.Path.html#method.file_stem) for `cover` both as a prefix (cover.png) and suffix (123_cover.png).

I added a couple of other fallbacks
* Check any file for "cover" prefix or suffix (e.g. path/images/cover.png, path/images/cover123.png, path/images/123cover.png)
* A last resort fallback, sort all images, return the first one

A more reliable check in the future would be to get the first image tag in the first chapter.

### DIFF
--- a/core/src/filesystem/media/format/epub.rs
+++ b/core/src/filesystem/media/format/epub.rs
@@ -166,6 +166,71 @@ impl EpubProcessor {
 		EpubDoc::new(path).map_err(|e| FileError::EpubOpenError(e.to_string()))
 	}
 
+	fn get_cover_path(resources: &HashMap<String, (PathBuf, String)>) -> Option<String> {
+		let search_result = resources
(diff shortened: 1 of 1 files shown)


aaronleopold (approved):
LGTM! Thank you!!

PR #605, opened 2025-02-17
:bug: Fix scanning epub without cover

darby42:
Fix an error when an epub is completely missing the cover.


aaronleopold (approved):
Thank you!

Issue #593, opened 2025-02-13
[FEATURE] Use metadata from matching opf file in directory for ePUB


Gibson431:
**Is your feature request related to a problem?**

Systems like LazyLibrarian store any downloaded metadata in a separate opf file in the same directory as the epub file.

The directory usually looks something like this (though this can be customized):

library/
├─ series 1/
│  ├─ Book 1/
│  │  ├─ Book 1.epub
│  │  ├─ Book 1.opf
│  ├─ Book 2/
│  │  ├─ Book 2.epub
│  │  ├─ metadata.opf **# Alternative**

**Describe the solution you'd like**

Ideally I would like the solution to parse the opf file and override any matching metadata that exists in the epub. Prioritize the opf in the database.

**Describe alternatives you've considered**

Adding the ability to fetch metadata and update the embedded metadata from Stump might mitigate this issue.

aaronleopold:
> Adding the ability to fetch metadata and update the embedded metadata from Stump might mitigate this issue.

Someone has actually started this effort 🙂

Gibson431:
Would it make sense (or be possible) to create an implementation of the MetadataSource trait that attempts to extract the metadata from the file instead of pulling from an online source?

aaronleopold:
I'm sure it's possible, I'm not sure it would be the best place for it though. That trait, correct me if I'm wrong @JMicheli I haven't looked closely at your work yet, is geared towards explicit API interactions between a set of supported **remote** providers.

You could argue the filesystem is a "provider" but it feels more like the scanner's domain. Something like "when I see book, is there adjacent file" or something. Happy to discuss on here, though!

JMicheli:
I agree that, the way the the trait is being written, it really makes sense for it to be focused on remote providers in the "online API that tells you about books" sense.

The scanner feels like the right place for reading an adjacent metadata file. It would be along the same lines as how metadata is populated from an internal comicinfo.xml right now.

aaronleopold:
Resolved by https://github.com/stumpapp/stump/pull/607

PR #607, opened 2025-02-18
:sparkles: scan matching opf file when scanning metadata for epubs

Gibson431:
This is a potential solution to #593 

I am not proud of the MediaMetadata merge but it works.

This only solves the situation where the opf file has the same name as the epub.

Open to suggestions for better solutions.

### DIFF
--- a/core/src/filesystem/media/format/epub.rs
+++ b/core/src/filesystem/media/format/epub.rs
@@ -127,11 +127,12 @@ impl FileProcessor for EpubProcessor {
 						}
 					},
 					Event::Text(e) => {
-						let text = e.unescape().unwrap().to_string();
-						opf_metadata
-							.entry(current_tag.clone())
-							.or_default()
-							.push(text);
+						if let Ok(text) = e.unescape() {
+							opf_metadata
+								.entry(current_tag.clone())
+								.or_default()
+								.push(text.to_string());
+						}
 					},
 					Event::Eof => {
 						break;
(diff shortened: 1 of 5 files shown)


core/src/filesystem/media/format/epub.rs
```diff
@@ -89,7 +90,104 @@ impl FileProcessor for EpubProcessor {
...
+							);
+						}
+					},
+					Event::Text(e) => {
+						let text = e.unescape().unwrap().to_string();
+						opf_metadata
+							.entry(current_tag.clone())
+							.or_default()
+							.push(text);
+					},
```
aaronleopold, line 135: Let's not unwrap, instead maybe something like:

```suggestion
					Event::Text(e) if let Ok(text) = e.unescape() => {
						opf_metadata
							.entry(current_tag.clone())
							.or_default()
							.push(text);
					},
```

Or `let Some(..)` if the result of the function is an option

aaronleopold (changes requested):
Thanks for taking a stab at this! I think it is good, but ask we remove the `unwrap()` before giving my full approval.

> I am not proud of the MediaMetadata merge but it works

It is a bit verbose to go field-by-field like this, but IMO more than acceptable for the time being since the merge story isn't fully thought through. On Discord we discussed potentially using something like https://docs.rs/merge/0.1.0/merge down the road. You're welcome to give it a try if you'd like

> This only solves the situation where the opf file has the same name as the epub

I also think this is the correct route for now. I don't necessarily want to overly complicate it more than this

core/src/filesystem/media/format/epub.rs
Gibson431: The method you suggested seems to be an experimental rust feature. I have implemented a similar way without the experimental feature.

Gibson431:
I have added the merge crate and updated the merge to suit

aaronleopold (approved):

PR #609, opened 2025-02-19
✨ (expo): Add basic browse stack

aaronleopold:
- Add links for browsing:
   - Libraries
   - Series
   - Books
   - Smart Lists (if allowed)
- Add user state for stump server context
- Add (ugly) Android icon
- Fix Android safe area views
- Fix stale progress state when exiting reading and re-entering
- Fix 401 response when loading ebooks
- Add better OPDS error-states for invalid feeds (testing other servers)
- Build out mobile docs


aaronleopold:
There are some spacing issues to work through, and the smart list browsing is not implemented, but this is more than a sizable chunk so should be good for now

PR #610, opened 2025-02-20
🩹 small improvements to web manifest PWA file

RileyMathews:
The initial web manifest file wasn't quite working for me after it was released so I took a stab at fixing things up. The two changes here are...

* make the 'base' parameter the root app page. This parameter gets set as the 'scope' of the manifest file which determines which routes are a part of your web app. With this being set to 'assets' the browser thought none of the actual app pages were part of the PWA and rendering browser controls.
* fix the asset file name. The '.' in the previous file name was creating a file path of '/assets/./manifest.webmanifest'. this path seemed to not resolve to anything. Setting a more explicit filename here leads to the manifest link the html resolving to '/assets/manifest.webmanifest' and does actually pull up the file when that path is requested.

### DIFF
--- a/apps/web/vite.config.ts
+++ b/apps/web/vite.config.ts
@@ -29,7 +29,7 @@ export default defineConfig({
 				maximumFileSizeToCacheInBytes: 5 * 1024 * 1024, // 5MB
 			},
 			outDir: '../dist/assets/',
-			base: '/assets/',
+			base: '/',
 			// TODO(pwa): Add more manifest definitions for better overall experience
 			manifest: {
 				id: 'stump',
@@ -55,7 +55,7 @@ export default defineConfig({
 					},
 				],
 			},
-			manifestFilename: './manifest.webmanifest',
+			manifestFilename: 'assets/manifest.webmanifest',
 		}),
 	],
 	publicDir: '../../../packages/browser/public',


aaronleopold (commented):
Thank you for fixing this! I think we need one additional change to correct the `registerSW.js` file not being fetched by the frontend. I believe we just need to add `buildBase: '/assets/',` to the `manifest` config object. I'm happy to do so if preferred

RileyMathews:
@aaronleopold so I tried to take a quick look at that but I am not able to reproduce it locally. I do see in the version running on my home server that the `registerSW.js` file gets a 404. But then locally even without the changes in this PR the file is loaded correctly when the page loads. Is this some difference between the dev and prod environments? I can definitely just throw the line in and push it up but I would also like to understand and reproduce the problem locally for my own edification.

aaronleopold:
Interesting. I guess my first question would be how you're running it? Are you generating a new web build and then reloading your server? Thats how I've been doing it, rather than using the vite dev server as an example. 

I don't _think_ it would be an issue introduced by your changes either, I should have been more clear but was typing on my phone while making lunch. I mostly meant while we are fiddling with this bit at all we should add it because another on discord reported the error.

RileyMathews:
Ahh yes that appears to have been the issue. I was previously using `yarn dev:web` to run locally. After just doing a `yarn web build` and just running the rust server I was able to see that without these changes the file did return a 404. However it appears that the changes in my PR also fix that file. It loads correctly with only the two changes in this PR but then if I try to add the `buildBase` change you suggests then the `registerSW.js` file 404s again.

RileyMathews:
> I don't think it would be an issue introduced by your changes either, I should have been more clear but was typing on my phone while making lunch. I mostly meant while we are fiddling with this bit at all we should add it because another on discord reported the error.

No worries here I understood it was an unrelated suggestion. I only added that it was breaking with and without the changes for me to emphasize why I thought it was an environment related issue.

aaronleopold:
> It loads correctly with only the two changes in this PR but then if I try to add the buildBase change you suggests then the registerSW.js file 404s again

Let's just omit it for now, then. Sorry for the runaround on this one! If it still is an issue after merge I'll take a look at that point

aaronleopold (approved):
Thanks again!!

PR #613, opened 2025-02-21
🩹 Switch to ProcessorError and asyncify thumbnail removal

JMicheli:
This pull request addresses several todo items in core::filesystem::image.

First, it adjusts the `ImageProcessor` trait to use `ProcessorError` as requested in a todo. In the course of doing this, I added a new `FileError::IncorrectProcessorError` variant to avoid a few `FileError::UnknownError`s. This error type could also reasonably be made a `ProcessorError` if desired, but doing it this way matched the current implementation more closely.

Second, it makes the `remove_thumbnails` function async as requested in another todo. This was done pretty easily by switching to tokio::fs. One major change I made to this function when async-ifying it was eliminating chunking and the use of the rayon crate to attempt parallel deletion across each chunk. Admittedly, I do not have a large enough test library to get a perceptible speed difference between the two, so testing with a larger library would be good. 

I believe that this is reasonable because rayon is used for CPU-bound work, but this is going to be I/O bound. I also don't think the chunking was doing much - I am pretty sure each rayon thread would be spending most of its time waiting on disk I/O from other threads. [This stack overflow post](https://unix.stackexchange.com/questions/349535/parallelize-recursive-deletion-with-find) supports my intuition, but I'm open to hearing other thoughts here.

### DIFF
--- a/core/src/filesystem/image/thumbnail/utils.rs
+++ b/core/src/filesystem/image/thumbnail/utils.rs
@@ -27,7 +27,7 @@ pub async fn remove_thumbnails(
 	let mut read_dir = tokio::fs::read_dir(thumbnails_dir).await?;
 
 	// Asynchronously collect thumbnails
-	let mut found_thumbnails = Vec::new();
+	let mut found_thumbnails = Vec::with_capacity(id_list.len());
 	while let Some(entry) = read_dir.next_entry().await? {
 		let path = entry.path();
 		if let Some(filename) = path.file_name().and_then(|f| f.to_str()) {
--- a/core/src/filesystem/image/thumbnail/utils.rs
+++ b/core/src/filesystem/image/thumbnail/utils.rs
@@ -1,9 +1,9 @@
 use std::path::{Path, PathBuf};
 
-use crate::{config::StumpConfig, filesystem::FileError};
-use rayon::prelude::{IntoParallelIterator, ParallelIterator};
 use tokio::fs;
 use tracing::{error, trace};
+
+use crate::{config::StumpConfig, filesystem::FileError};
 
 pub async fn place_thumbnail(
 	id: &str,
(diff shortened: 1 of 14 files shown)


aaronleopold:
> One major change I made to this function when async-ifying it was eliminating chunking and the use of the rayon crate... I believe that this is reasonable because rayon is used for CPU-bound work, but this is going to be I/O bound

I think this was the correct move! I also think, in general, there are quirks when mixing rayon and async ops. An author of tokio had written a good writeup for just this, I'll try to find it.

Otherwise, thank you for this!! I'll aim to take a look later this afternoon

core/src/filesystem/image/thumbnail/utils.rs
```diff
@@ -16,61 +16,44 @@ pub async fn place_thumbnail(
...
-					}
-				})
-			})
-		})
-		.map(|iter| iter.collect::<Vec<PathBuf>>())
-		.unwrap_or_default();
+	let mut read_dir = tokio::fs::read_dir(thumbnails_dir).await?;
+
+	// Asynchronously collect thumbnails
+	let mut found_thumbnails = Vec::new();
```
aaronleopold, line 30: We could probably preallocate this vec to the id_list size by default

core/src/filesystem/image/thumbnail/utils.rs
```diff
@@ -16,61 +16,44 @@ pub async fn place_thumbnail(
...
-				} else {
-					res.ok()
-				}
-			})
-			.collect::<Vec<()>>();
-
-		trace!(deleted_count = results.len(), "Deleted thumbnail batch");
-		deleted_thumbnails_count += results.len() as u64;
+	for (idx, path) in found_thumbnails.iter().enumerate() {
+		if idx % THUMBNAIL_LOG_FREQUENCY == 0 {
```
aaronleopold, line 46: Nice, I like this

aaronleopold (approved):
This looks good to me! I had a minor suggestion but it is largely unimportant, so approving right away - feel free to merge whenever. Thank you!

core/src/filesystem/image/thumbnail/utils.rs
JMicheli: Good idea, I'll make the change later and merge.

PR #615, opened 2025-02-22
🐛 Fix reading progress resetting due to cached data

xendke:
when useQuery has a cache for a book entity, the first progress api request is triggered with outdated progress data.

Steps to Reproduce:
1.  Go to a book and click Continue Reading
2. Chapter one loads lets say
3. Open chapters modal and click on Chapter 2
4. Click the back arrow on the top left
5. Click Continue Reading again
6. repeat steps 2-5 and notice how the book loads alternating between chapter 1 and 2
-------

this PR removes cache from this query which is not ideal. open to suggestions! I tried using some properties that come back from useEpubLazy() but it was getting complicated same for invalidating the query cache.

Alternatively, we could prevent the very first progress api request after loading the book/entity since there should be no need for that first one. I havent thought that completely through.

### DIFF
--- a/packages/browser/src/components/readers/epub/EpubJsReader.tsx
+++ b/packages/browser/src/components/readers/epub/EpubJsReader.tsx
@@ -302,15 +302,19 @@ export default function EpubJsReader({ id, initialCfi }: EpubJsReaderProps) {
 	}, [rendition, epubPreferences, theme])
 
 	/**
-	 * This effect is responsible for invalidating the in progress media query whenever
-	 * the epub reader is unmounted. This is so that the in progress media query will
-	 * refetch the updated progress information.
+	 * Invalidate the book query when a reader is unmounted so that the book overview
+	 * is updated with the latest read progress
 	 */
-	useEffect(() => {
-		return () => {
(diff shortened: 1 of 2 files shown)


aaronleopold (commented):
> This PR removes cache from this query which is not ideal. open to suggestions!

I actually fixed essentially the same bug for the mobile app, if you want to try and give porting it here a shot! Basically, I refetch whenever the reader unmounts. That way, upon reentry, the progress should be correct. See [here](https://github.com/stumpapp/stump/blob/experimental/apps/expo/app/server/%5Bid%5D/books/%5Bid%5D/read.tsx#L75-L88)

aaronleopold (approved):
LGTM! Thank you!

xendke:
@aaronleopold oops it doesn't look like this PR solved the issue 😞 react-query still wants to return cached data on mount 🤷 

seems like `queryClient.removeQueries` does the trick by just trashing the cache. I can make a small PR for this- thoughts?

aaronleopold:
@xendke No worries! I'd say let's try and find the root issue for why it might be messing up. If disabling cache works, but refetching doesn't, the refetch logic might have needed some adjustments.

I'm also okay with a more short-term fix of just disabling cache, too 🙂

PR #625, opened 2025-03-05
:bug: Actually scan opf when getting metadata

Gibson431:
#607 Did not actually implement the feature as `process_metadata` was never called by `process`

### DIFF
--- a/core/src/filesystem/media/format/epub.rs
+++ b/core/src/filesystem/media/format/epub.rs
@@ -166,8 +166,11 @@ impl FileProcessor for EpubProcessor {
 		let epub_file = Self::open(path)?;
 
 		let pages = epub_file.get_num_pages() as i32;
-		// Note: The metadata is already parsed by the EPUB library, so might as well use it
-		let metadata = MediaMetadata::from(epub_file.metadata);
+		// Get metadata from epub file or opf if it exists
+		let metadata = match Self::process_metadata(path) {
+			Ok(Some(m)) => m,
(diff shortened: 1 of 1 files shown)


core/src/filesystem/media/format/epub.rs
```diff
@@ -166,8 +166,11 @@ impl FileProcessor for EpubProcessor {
 		let epub_file = Self::open(path)?;
 
 		let pages = epub_file.get_num_pages() as i32;
-		// Note: The metadata is already parsed by the EPUB library, so might as well use it
-		let metadata = MediaMetadata::from(epub_file.metadata);
+		// Get metadata from epub file or opf if it exists
+		let metadata = match Self::process_metadata(path) {
+			Ok(Some(m)) => m,
+			_ => MediaMetadata::from(epub_file.metadata),
+		};
```
aaronleopold, line 173: So this will actually buffer the file in memory twice:

1. Line 166, `let epub_file = Self::open(path)?;`
2. Again in `Self::process_metadata(path)`

I don't think this is _awful_, most ebooks are somewhat small relative to other digital content. I'm not 100% sure it is worth the effort to refactor, but just to spitball some thoughts:

Maybe we could either move this up to _before_ we load the file the first time, that way it gets dropped and re-loaded instead of holding the file twice at once?

Gibson431: Make sense, I'll fix it up

aaronleopold (approved):
Thank you!

PR #626, opened 2025-03-05
:arrow_up: Upgrade `axum` and related dependencies

JMicheli:
This pull request upgrades Axum to the latest version, 0.8.1.

This was a fairly extensive set of changes. Axum route handlers now follow a different syntax (`/:id` => `/{id}`) and a few other APIs changed. We also had to update a lot of downstream crates that rely on Axum themselves. The most significant change on that front was the necessary upgrading of utoipa to 5.3.1 and utoipa-swagger-ui to 9.0.0.

The utoipa changes were pretty far-reaching as well. I had to make a minor modification to the smart filter generation macro and implement `ToSchema` for a good number of structures because utoipa now requires all nested structs to also implement `ToSchema`. The `aliases` attribute was also removed, so those types are now represented by `pub type`s. Finally, because of changes to utoipa, it is now possible to cause stack overflows by having loops in the dependency graph for a struct/enum. To solve this, I added `#[schema(no_recursion)]` annotations to several fields. It would be a good idea for someone to review the API documentation and make sure it still feels sensible in light of these changes.


JMicheli:
Didn't realize that `is_none_or` was experimental in Stump's MSRV. I'll revert those changes later today.

aaronleopold (approved):
I gave this a good look over lunch and don't see any issues. I was curious what changed that made it so we couldn't use extractors in middleware, like `HostExtractor`, but didn't have time to look in the changelog for hints. Not a biggie either way.

Thanks for knocking this out! Feel free to merge whenever

aaronleopold:
Also:

> Didn't realize that is_none_or was experimental in Stump's MSRV. I'll revert those changes later today

We can probably bump that after this whenever

JMicheli:
Well, there isn't really a good explanation for why I changed it to not use `HostExtractor` in the function signature. I reverted my changes and redid auth.rs and came to this much simpler diff. I'm glad you pointed that out.

PR #628, opened 2025-03-06
:hammer: add android dev shell to flake

Gibson431:
This adds android development to the nix flake.

It works for hardware-in-loop testing. I have not tried with an emulator.


aaronleopold (approved):
Thank you!

PR #631, opened 2025-03-08
⬆️ Bump MSRV

JMicheli:
This PR bumps the minimum supported rust version (MSRV) to 1.85.0 and also fixes a couple of clippy lints that were added for that version.

Additionally, I went to fix a lint error by changing `UserProfile` in `core::db::entity::user` to use `Full(Box<User>)` instead and avoid bloating the size of the enum. However, I realized that the `UserProfile` isn't actually used anywhere, so I just removed it instead. Let me know if it has a future use planned and should be kept in.

### DIFF
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -18,7 +18,7 @@ runs:
     - name: Install Rust
       uses: actions-rs/toolchain@v1
       with:
-        toolchain: 1.81.0
+        toolchain: 1.85.0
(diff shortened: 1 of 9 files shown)


aaronleopold (approved):
lgtm, thank you!!

> I realized that the `UserProfile` isn't actually used anywhere, so I just removed it instead. Let me know if it has a future use planned and should be kept in

I probably forgot to remove it myself when a thought never panned out. It's good staying removed

PR #647, opened 2025-04-20
🐛 Fix reading progress resetting due to cached data again

xendke:
Finally had some time to look at this again! 

Following up on: https://github.com/stumpapp/stump/pull/615

Having both `cancelQueries` and `refetchQueries` ensures we invalidate the cache and refetch fresh data.


aaronleopold (approved):
Sorry I've really let this one sit! This looks good to me, so I'm going to merge it in now. Thank you!

Issue #646, opened 2025-04-19
[BUG] Docker compose - addgroup: number 65536 is not in 0..60000 range


sggt:
## Description of the Bug

I written a docker compose file, with my standard docker user on that system:
```yml
services:
  stump:
    image: aaronleopold/stump:latest
    container_name: stump
    volumes:
      - /config:/config
      - /data:/data
    ports:
      - 10801:10801
    environment:
      - PUID=1029
      - PGID=65536
      - STUMP_CONFIG_DIR=/config
    security_opt:
      - no-new-privileges:true
    restart: unless-stopped
```
On startup I get following error message in the log:
`addgroup: number 65536 is not in 0..60000 range `

## Expected Behavior

That should be working, I never hat problems with that GID. Why this low limit?

**Build Details:**

- Version: [0.0.10]
- Docker: [yes]

aaronleopold:
I won't be by a computer or a while today, but if I had to guess it's probably a constraint with busybox

aaronleopold:
The final layer in Stump's image is a distroless base, it's really stripped down and secure but lacks basic unix features, e.g., not even a shell. Busybox fills the gaps by providing implementations for some (e.g., `addgroup`).

Unfortunately it looks like it is [a baked in constraint](https://github.com/brgl/busybox/blob/abbf17abccbf832365d9acf1c280369ba7d5f8b2/loginutils/adduser.c#L33-L38), assuming I understand what I am looking at. One thing we could potentially try is adding some level of detection and manual intervention. Something like the following in the `entrypoint.sh` file:

```bash
if [ "$PGID" -gt 60000 ]; then
    echo "Warning: PGID $PGID is too large for BusyBox's addgroup (limit is 60000)"
    echo "Using fallback method to add group..."
    
    echo "$GROUP:x:$PGID:" >> /etc/group
else
    # ... Exiting logic ... #
fi
```

If you're able to give that a try, please let me know how it goes. Otherwise, I think the only alternative would be to accept the somewhat larger vulnerability surface area and image size and move off of busybox + distroless images

sggt:
I have tested the fallback method and it worked! Thank you!

aaronleopold:
@sggt if you're able to, please make a PR with the fix and we can get it in for others who might face the same

PR #652, opened 2025-04-25
:bug: Fix `addgroup` error for numbers beyond busybox limit

sggt:
Credit for that workaround to bug #646 goes to @aaronleopold, I have only successfully tested it and created this PR.


aaronleopold (approved):
Thank you!

Issue #533, opened 2024-12-21
[FEATURE] Logic for thumbnails generation size


hucurve:
The cause was seeing thumbnails of images that varied greatly in clarity
![image](https://github.com/user-attachments/assets/5fbd63ab-ced7-4e5c-9bbb-09f5e81b658e)

I'm thinking that since the height of the picture is the height of the frame in the UI,
would it be better to limit the  height in a value and scale the image weight equally?
There is a very ugly diagram I do not know if I understand it right.
![image](https://github.com/user-attachments/assets/c008e5bd-4188-4b8f-b409-008295893c83)

aaronleopold:
Just to make sure I'm understanding properly, is the ask to make a third option for thumbnail generation which allows you to set an explicit size for one dimension and just scale the other one?

hucurve:
Yes, of course you can also have new designs for your own design interactions

aaronleopold:
Okay, yeah I think that option makes sense. Thanks for the suggestion! I'll be sure to update this issue once I can get to it

aaronleopold:
@hucurve I have completed this in my branch with all the big migration changes. Here is a snippet of what it looks like:

<img width="1803" height="1278" alt="Image" src="https://github.com/user-attachments/assets/28ffc228-2425-4916-8976-50460297088d" />

Not 100% happy with the UI, but the form elements are what really matters here. I'll keep this open until the migration work is complete 👍

aaronleopold:
I'm closing this issue to clear the backlog a bit, but please note it won't be released until the SeaORM migration is stable

PR #665, opened 2025-06-28
:construction: Migrate library creation and management

aaronleopold:
I figured I should start creating PRs again for the migration work, instead of just tackling a bunch of work directly in the feature branch, so that it doesn't look like the project is dead

This PR, and most of the migration-related ones in the future, will almost certain fail CI. That will be the very last thing I fix. Anyways, there is a lot in this PR but it is primarily focused around migrating all of the library management and creation UIs to be functional again with the new backend/core. I'll leave it as draft until I do some more testing, mostly around the thumbnail configuration changes (the migration will ship with support for https://github.com/stumpapp/stump/issues/533)


aaronleopold:
It's far from perfect, but I think this is good for now. I've marked some remaining areas with `TODO`s that ideally will be revisited once I start pushing out some sort of `unstable` tag for folks to test the migration

PR #666, opened 2025-07-04
🚧 Migrate settings pages

aaronleopold:
This PR is primarily focused around migrating all of the settings pages to be functional again with the new backend/core. This includes:

- User preferences
- API key management
- Server management (logs, jobs, etc)
- User management

I'm hoping to wrap it up over the weekend


aaronleopold:
Yeah so this definitely wasn't completed over the weekend like I thought, but I think it is now at a good enough spot to merge

Issue #671, opened 2025-07-10
[BUG] Steady memory increase during thumbnail generation


teodorkostov:
## Description of the Bug

The more thumbnails are generated the more memory the application needs.

## Expected Behavior

The application memory should not depend on the amount of thumbnails that have to be generated.

## To Reproduce

1. Run the application in a container with a hard memory limit.
2. Add about a 100 books - this amount seems to require around 1GB of RAM.
3. Regenerate the thumbnails in the library with different memory limits.

## Debugging Information

`journalctl -r`

**Environment:**

- OS: Arch Linux

**Build Details:**

- Version: 0.0.10
- Docker: yes

**Additional context**

Run as Nomad job. Data (books) are mounted as read only.

aaronleopold:
> The more thumbnails are generated the more memory the application needs

Are you actually observing a memory leak, e.g., the memory continues to steadily increase after the job completes, or are you just observing high resource utilization during the lifecycle of the job?
 
Stump will, by default, generate 50 thumbnails at a time. This means it will load 50 full-res images into memory, convert and resize them in memory, then dump to the disk. You can configure this value to be less, so it doesn't process more than a preferred amount at once: https://www.stumpapp.dev/guides/configuration/server-options#stump_max_thumbnail_concurrency

teodorkostov:
Hey @aaronleopold, thank you for the fast reaction. Yes, it is not a leak but a steady increase in memory.

I just set `STUMP_MAX_THUMBNAIL_CONCURRENCY=10`. Deleted all thumbnails and started regenerating. Around thumbnail 144 OOM Killed. Memory limit is 512MB of RAM.

With `STUMP_MAX_THUMBNAIL_CONCURRENCY=3` the 512MB were enough. Maybe this default setting of `50` is a bit too high as a default?

aaronleopold:
Yeah of course! No problem

> I just set `STUMP_MAX_THUMBNAIL_CONCURRENCY=10`. Deleted all thumbnails and started regenerating. Around thumbnail 144 OOM Killed. Memory limit is 512MB of RAM.

So this definitely points to something else going on, but not 100% sure what. You're saying it OOM'd around batch 14 (144 images / 10), which implies there is some memory being held by the container between each batch and aligns with:

> it is not a leak but a steady increase in memory

Are you able to see the memory go down after the job completes? Or does docker not release the memory? 

[...]

PR #675, opened 2025-07-13
:recycle: Refactor thumbnail generation to trim memory consumption

aaronleopold:
This is largely explorative at the moment, trying to improve the situation reported in #671. I think this should, at the very least, improve the situation since I no longer spawn an unconstrained number of futures and defer to explicit batching. 

If this does wind up being the source of the problem, or improves the situation significantly, I'll port the same patterns to the scanner.


aaronleopold:
I am just going to merge this to `experimental` and see how it goes

PR #687, opened 2025-07-30
Fix image reader bugs

aaronleopold:
Another round of image reader bug fixes


aaronleopold:
I unfortunately don't have the time to really thoroughly put these changes through their paces, but I'm going to merge them to get them into `nightly` by the end of my night so (hopefully) others can help me out by testing them.

I briefly loaded up each reader and didn't encounter any really noticeable issues, but there are definitely quirks to work through over time: 
- The initial load of images in the reader foot don't seem to always compute the sizes after load, which makes a little stutter when it recomputes double-spread layouts (in the footer)
- The foot increases in height randomly when scrolling to the middle of a set

Issue #684, opened 2025-07-29
[BUG] v0.0.11 reader experience


retmas-gh:
Hello.

With the newest stable update v0.0.11 the reader experience has several bugs I encountered.

I have not used the issue template since I'm creating one issue around the topic, but if you would prefer it created separately, please say so.

The issues encountered:
1. Setting "Vertical scroll" option shows all images on one page as expected, but it does not allow scrolling.
2. The progress bar stays at fixed value (40% due to error below i guess); in the console I can observe following error:
```
Invalid prop `value` of value `40` supplied to `Progress`. The `value` prop must be:
  - a positive number
  - less than the value passed to `max` (or 100 if no `max` prop is set)
  - `null` or `undefined` if the progress is indeterminate.

Defaulting to `null`.
```
3. On the light theme, the top buttons, due to a shadow overlay, are not visible (they are silver or darker).

Additionally - this is something that didn't work previously but is related to the topic - the library settings set in
`Series`>`Settings`>`Reading` are not applied.

Kind regards.

aaronleopold:
Thanks for the write-up! This is really helpful, and it highlights just how important experimental/nightly testing is. 

The first two are really easy fixes, the third is slightly more involved since I'll likely have to consider a new theme value to get a good contrast color for all themes. I need to look into the last one more. I'll aim to have fixes in `nightly` by the end of the week

aaronleopold:
@retmas-gh These should be fixed in the latest `nightly` image which just got deployed. Would you be able to verify that the issues are corrected? There are no database migrations in `nightly` diffed against main, so it should be safe to quickly swap to it.

If the issue is resolved, I'll aim to push another tag over the weekend

retmas-gh:
Hello.

I'm sorry I did not test this beforehand - I was not subscribed to the Pull Requests notifications so I was not really aware there was a release coming with you being busy with the rewrite and all.

Thank you for the quick fixes; I've tested all aspects I reported myself and:

1. It is indeed possible now to scroll with the "Vertical Scroll" flow applied
2. The progress bar is no longer fixed but it fills (I presume) to the max of book_max_pages_count% (so with 10 max pages in the book the last page is at 10% of the whole progress bar, with 7 pages its 7% etc.); additionaly, the counter starts at 0 not 1 (the pages go from 0 to 9 out of 10 for example)
3. The buttons are now visible in all the themes
[...]

PR #692, opened 2025-08-01
Fix additional image reader bugs

aaronleopold:
Relates to the follow-up findings from https://github.com/stumpapp/stump/issues/684#issuecomment-3140953305

I haven't yet worked through all of the reported issues, but I'll leave this as draft and update the description once I have


aaronleopold:
I'm going to eagerly merge this to get the build going. I've tested it with 10 different comics and haven't been able to replicate the original issue, so I likely won't without burning more time than I otherwise have tonight

Issue #693, opened 2025-08-02
[FEATURE] Double page offset by one page


Arklaum:
## The Problem

Double page setting sometimes uses the wrong two pages. 

What I mean by this is look at a physical manga: we have the cover, then we open the cover and on the right is the back of the cover (blank) and then on the left is page 1. Sometimes digital versions do not include that blank page, so on such books when we use the double page option we get:

**Current behaviour**:
- the cover on its own
- then page 1 on the right and page 2 on the left

**But what should happen is**:
- the cover on its own
- then page 1 on its own
- then page 2 on the right and page 3 on the left

### Describe the solution you'd like
- A switch that puts page 1 by itself (off by default)

### Describe alternatives you've considered
- A switch 'cover is separate' (on by default, but when turned off will fix the issue too)

PR #702, opened 2025-08-09
✨ Add toggle to separate the page that follows the cover

Arklaum:
Solves #693 by adding a toggle that separates the second page. 

Useful for books where the title page is the second page and not the inside cover.


aaronleopold (approved):
Awesome, thank you! I'll merge shortly after CI does its thing

Issue #679, opened 2025-07-23
[FEATURE] Web Comic Reader Improvements


ghost:
Two things that I think could massively improve the web Comic Reader:
1. More image scaling zoom levels. Sometimes I want it zoomed in, but not so much that it fits the width. For example, something like 75%. Bonus if there's a custom zoom setting too.
2. Clicking on the left and right blank spaces outside of the image currently only flips the page in Height image scaling mode, but not in Width (or Original) image scaling mode.
  It also might be better if rather than clicking outside the image, you click on the left and right sides inside the  image, and clicking the center of the image brings up the interface/HUD thing.
4. Current page number out of total pages.
5. When flipping the page, it should automatically scroll back to the top of the image.

Aside from these, the web comic reader is actually pretty good!

aaronleopold:
> More image scaling zoom levels. Sometimes I want it zoomed in, but not so much that it fits the width

Constraining the width of the reader content (excluding controls) could probably achieve this. Is this about you would want? E.g., you could shrink the center content to be 75% and therefore re-scale the image accordingly

> Current page number out of total pages

This is coming in the next release, and is already available in `nightly`. See [here](https://github.com/stumpapp/stump/blob/develop/packages/browser/src/components/readers/imageBased/container/ReaderFooter.tsx#L168-L170) if you're curious, I basically just touched up the reader to better align with the look-and-feel of the mobile app

> When flipping the page, it should automatically scroll back to the top of the image

Is this just for original scaling mode? I think all the others contain the image within the dimensions of the window, but would have to double check. Regardless, I agree - if you had to scroll then changing page should pop you back

ghost:
> Constraining the width of the reader content (excluding controls) could probably achieve this. Is this about you would want? E.g., you could shrink the center content to be 75% and therefore re-scale the image accordingly

Yeah, that's basically it. Just some midway point between the current height and width scaling modes at the very least.

Bonus if you add a custom percentage option! :D 

> This is coming in the next release, and is already available in nightly.

Ah, cool!

> Is this just for original scaling mode?

It wasn't scrolling back to the top in Width image scaling most of the time, but on occasion it would. Seemed to happen randomely. I didn't try it in Original scaling because Original never looked different from Width image scaling, so I never used it. But I just did, and yeah, same problem with that one too.

hollisticated-horse:
I'd also add : 

- [ ] improve visibility of icons top-left top-right. Currently they are hard to see. 
[...]

Issue #606, opened 2025-02-18
[FEATURE] Web reading page zoom support


c608345:
I'd like zoom reading page.

hollisticated-horse:
This works really well in nightly. I would love to also see the improvements mentioned in #679.

PR #704, opened 2025-08-11
✨ Add ctrl + scroll to zoom and middle click to pan for paged reader

Arklaum:
Adds a zoom ability with ctrl + scroll when using paged reader. I would've liked to make no holding any modifier keys to zoom as default, since it would allow zooming and page turning while only using mouse, no need to touch keyboard, but I didn't want to mess with the current expected functionality of scrolling on a landscape screen when using scaling setting 'width'. 

Also adds hold then drag middle click to pan. 

Solves #606 and maybe issue 1 in #679 but I think that was about having a permanent 75% zoom and this resets every page turn, but it fulfils the same desire to see content closer.

Note that it doesn't change any behaviour for tap sides to navigate, so after zooming in, you still click the same places as before, doesn't matter where you've moved the images it won't move the click location.

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -20,6 +20,7 @@
 		"@dnd-kit/core": "^6.1.0",
 		"@dnd-kit/sortable": "^8.0.0",
 		"@hookform/resolvers": "^3.9.0",
+		"@panzoom/panzoom": "^4.6.0",
 		"@stump/client": "*",
 		"@stump/components": "*",
 		"@stump/i18n": "*",
--- a/packages/browser/src/components/readers/imageBased/paged/PagedReader.tsx
+++ b/packages/browser/src/components/readers/imageBased/paged/PagedReader.tsx
@@ -1,6 +1,7 @@
+import Panzoom from '@panzoom/panzoom'
 import type { Media } from '@stump/sdk'
 import clsx from 'clsx'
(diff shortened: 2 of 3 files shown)


Arklaum:
Hmm there's an issue on mobile that I need to fix.

Arklaum:
All fixed now

Arklaum:
There seems to be an issue on PC with left click to pan: the pan resets once you let go of the mouse button if you haven't changed any individual book settings. Seems related to the issue where library settings override the book settings.

I can change the behaviour to only happening when doing left click to open the overlay, by changing 

`onPageClick={() => setSettings({ showToolBar: !showToolBar && !panningDetected.current })}`

to

``onPageClick={() => { if (!panningDetected.current) { setSettings({ showToolBar: !showToolBar })}}}``

which is better than nothing. It doesn't happen at all if i remove `imageScaling` from line 130, but then when changing image scaling, the zoom won't reset.

Arklaum:
Okay all done!

aaronleopold:
> Okay all done!

Awesome, thank you! I'm just getting off work and will try to get to a review for this and your other PR by tonight

aaronleopold (approved):
I don't have time to run these changes locally myself, but I think they look good! I'll approve and merge once CI does its thing

PR #705, opened 2025-08-11
🐛 Fix inability to save after toggling 'Generate thumbnail images' off

Arklaum:
Enables clicking 'Save changes' after turning 'Generate thumbnail images' off. Previously you had to change a setting then toggle the switch and only then could you click save.


aaronleopold (approved):
Thanks!

PR #706, opened 2025-08-12
🐛 Fix click to flip page (SideBarControl) issues

Arklaum:
Fixed panzoom interfering with SideBarControl ‘active’ settings

Fixed inability to left click on SideBarControl when using image scaling ‘Width’ by observing if page set width changes. issue 2 in #679? (problem due to page set width calculation for fixSideNavigation returning wrong value after refreshing page / first time loading book into browser)

This same change also fixed the panzoom origin calculation on first time ever opening a book being wrong due to the same problem.

Fixed height of SideBarControl so that when using scaling ‘width’ on a landscape screen looking at a single portrait page it spans the whole height. (was also an issue with double pages, but it spanned nearly the full height already) It was kind of a work around, so it might not span the full height on super large images.  

Changed width of SideBarControl so that it peeks 3% of the viewing area width over the page even if fixSideNavigation is not true, and furthermore changed fixSideNavigation to be true for 86% rather than 80% of the viewing area since that means once you have 7% space left on e.g. the right hand side then the 10% fix kicks in (i.e. the click area will always be a minimum 10% of the viewing area) This was also mentioned in issue 2 of #679

### DIFF
--- a/packages/browser/src/components/readers/imageBased/paged/PagedReader.tsx
+++ b/packages/browser/src/components/readers/imageBased/paged/PagedReader.tsx
@@ -1,14 +1,14 @@
 import Panzoom from '@panzoom/panzoom'
 import type { Media } from '@stump/sdk'
 import clsx from 'clsx'
-import { memo, useCallback, useEffect, useMemo, useRef } from 'react'
+import { memo, useCallback, useEffect, useMemo, useRef, useState } from 'react'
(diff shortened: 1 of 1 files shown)


aaronleopold (approved):
You rock, thank you! I haven't had a moment to test it myself but the changes look good to me

PR #715, opened 2025-08-20
🐛 Fix flashing between pages

Arklaum:
When changing page on Firefox and Chrome, what happens very fast is: the current pages disappear, we're left with the background, then the new ones appear. This is made worse when the page is white and the background dark. Doesn't happen on Safari unless flicking through a book really fast (which is quite slow on Safari).

Not sure I like this fix, but I didn't want to remove the keys. This makes the next set of images render with zero opacity at the required size before turning the page, meaning they are ready to use immediately on page turn. Note this is only for the forwards direction, but I can easily add the other if needed.

Outcome in my testing: 

- Firefox: only flashes if flicking through a book really fast; normal reading pace should be fine
- Chrome: zero flashing ever even if flicking through a book really fast
- Safari: zero flashing ever even if flicking through a book really fast (again which is still slow)

---

By flicking through a book fast, I just mean holding the arrow key down.

### DIFF
--- a/packages/browser/src/components/readers/imageBased/paged/PageSet.tsx
+++ b/packages/browser/src/components/readers/imageBased/paged/PageSet.tsx
@@ -16,7 +16,7 @@ const PageSet = forwardRef<HTMLDivElement, Props>(
 	({ currentPage, getPageUrl, onPageClick }, ref) => {
 		const { setDimensions, book, pageSets } = useImageBaseReaderContext()
 		const {
-			bookPreferences: { imageScaling, brightness },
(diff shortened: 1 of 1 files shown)


Arklaum:
Now that sounds like a fix I will like! I'll certainly give it a look sometime.

aaronleopold:
> Now that sounds like a fix I will like! I'll certainly give it a look sometime.

I knocked out a rough implementation here if you're interested: https://github.com/stumpapp/stump/pull/720. It's on the big refactor branch, so it won't make its way to nightly for a little bit longer, though

PR #716, opened 2025-08-21
🐛 Fix inability to save after removing ignore rules

Arklaum:
Glob.


aaronleopold (approved):
Thank you! I'll merge this after the last PR wraps up its CI pipeline

Issue #660, opened 2025-05-15
[BUG] EPUB Font controls do not effect book contents


LordHerdier:
## Description of the Bug

(On nightly) the new font selection under the appearance menu does not effect the contents of the book. The font for the title of the book will change, but nothing else.

## Expected Behavior

Expected the font family for the book's content to change, not the font family for the book title.

## To Reproduce

1. Be on nightly/develop branch
2. Open an epub book
3. Click on theme/appearance controls and change the font family
4. Observe that the font did not change

## Screenshots

![Image](https://github.com/user-attachments/assets/0021eb3e-8693-4613-814c-eb5fdfcce4f4)

Notice the title is in OpenDyslexic, while the content is still the default font.

**Environment:**

- OS: [Debian 12]
- Device: [Desktop]

**Build Details:**

- Version: [nightly]
- Docker: [yes and no] (can confirm that it doesn't work in dev build or in docker)

**Additional context**

It doesn't look like EpubJS respects CSS quite the way that would be expected (possibly due to the iframes it's using?), so the current implementation of applying the theme in `packages/browser/src/components/readers/epub/themes.ts` directly to the reader container may not be the best course of action. 

From the [epubjs documentation](http://epubjs.org/documentation/0.3/#contentscss), it looks like you can pass CSS directly to the epub body, so that might be an option to pursue. 

I'll take a look at the code and see if I can get this resolved.

LordHerdier:
Okay so it looks like the issue _is_ with the iframe, but not the CSS like I thought. Turns out that we can apply system fonts directly with epubjs's addStylesheetRules, but our custom ones weren't loading due to the iframe not having access to the font file. Looks like it should work if we inject a style tag into the iframe's <head> tag and have it download the desired font from that.

Going to do a bit more testing..

PR #721, opened 2025-08-23
🐛 Fix: Apply fonts to books and 💄 dropdown boxes render the font

Arklaum:
I moved the fonts, and fixed typos inside the font css files, so the real changes were in the 6 files:

- EpubJsReader.tsx
- FontFamily.tsx
- FontSelect.tsx
- DefaultFontFamily.tsx
- ComboBox.tsx
- Overrides.css (i hope this one is okay)

I know there's another pull request for this fix #663, and I'm not sure what other features that one included so I don't know if this supersedes it (I think it does). I did try that fork but there seemed to be issues like colour of text and background not loading correctly, etc. Solves #660 I guess. 

I did not fix the flashing when changing line height / font / font size.

I made sure the dropdown boxes for font in 'reader settings' rendered in their own font and not the font set under 'appearance'.

Edit: I didn't update the documentation either. I'll do that now and send an update.


Arklaum:
Okay documentation has been updated

packages/browser/src/components/readers/epub/EpubJsReader.tsx
```diff
@@ -55,6 +55,20 @@ class SectionLengths {
...
+
+	const head = doc.head
+	if (!head) return
+
+	const link = doc.createElement('link')
+	link.rel = 'stylesheet'
+	link.id = 'stump-fonts-stylesheet'
+	link.href = '/assets/fonts/fonts.css'
+	head.appendChild(link)
+}
```
aaronleopold, line 70: Nice

aaronleopold (approved):
This looks great, thank you!

> I don't know if this supersedes it (I think it does) [the other PR]

I think it does, at the very least I can see the fonts being requested by the component locally and they seem to be working as expected. This is also a good bit leaner of a change, which is a plus.

> I did not fix the flashing when changing line height / font / font size

I think this is an artifact of epub.js, so I think it's acceptable

> I made sure the dropdown boxes for font in 'reader settings' rendered in their own font

I appreciate this!

Arklaum:
> I think this is an artifact of epub.js, so I think it's acceptable

Oh I see. I was testing out an implementation of foliate-js because of my Japanese book issues and it didn't flicker.

https://github.com/user-attachments/assets/ad671155-0cc6-46e7-ba44-cb022b5b5ab2

aaronleopold:
> > I think this is an artifact of epub.js, so I think it's acceptable
> 
> Oh I see. I was testing out an implementation of foliate-js because of my Japanese book issues and it didn't flicker.
> foliate-js.mov

My browser doesn't want to show me that video for whatever reason, but I'll trust you 😅 I haven't explored foliate-js, epub.js has been a headache to work with in the past. The mobile app will likely use readium since they have native toolkits, so I had loose plans to explore that for web at some point. If it's an area of interest you're welcome to try and port the reader to either, at the end of the day if it can support highlights/annotations, epubcfi-based progress tracking, etc, and provides a better experience, it's a win

I would ask that you base it into `breaking/sea-orm` though so conflicts are more manageable for that larger-scale of a change

PR #739, opened 2025-08-31
🐛 Fix epubjs flickering

Arklaum:
Fixes flickering when changing font size, line height or font family

Not sure if the lint is okay


aaronleopold (approved):
Awesome, thank you!

> Not sure if the lint is okay

Yeah it's totally fine, the epub.js types aren't the best

Issue #723, opened 2025-08-23
[BUG] Issues with Japanese epubs


Arklaum:
## Description of the Bug

For comparison:
- Books in English turn the page left to right, and lines are written left to right, and new lines go top to bottom
- Books in Japanese turn the page right to left, and lines are written top to bottom, and new lines go right to left. It's the same for Traditional Chinese books.

However for my Japanese books in stump:
-Lines read top to bottom but align themselves at the bottom, and new lines go right to left. 
-Or lines read left to right but justify themselves to the right, and new lines go top to bottom.

And the punctuation is on the wrong side of the text.

### Summary of the two issues:
1. `direction` is wrong (should be ltr, because text goes "forwards") 
2. `writing-mode` is wrong (should be vertical-rl)

### Issue 1 Comparison: Book 1 (Stump vs Apple Books)

| Stump | Apple Books |
| :---: | :---: |
| <img width="155" height="299" alt="Image" src="https://github.com/user-attachments/assets/63683673-c122-4e36-a20e-a4ba47a56dc3" /> | <img width="142" height="312" alt="Image" src="https://github.com/user-attachments/assets/65c62750-1545-4a51-a5c7-43ec87f671e0" /> |

### Issue 2 Comparison: Book 2

| Stump | Apple Books |
| :---: | :---: |
| <img width="225" height="108" alt="Image" src="https://github.com/user-attachments/assets/04a17429-d7a7-4a29-afd3-9bb639ca8159" /> | <img width="127" height="317" alt="Image" src="https://github.com/user-attachments/assets/5f497a67-424a-40c3-ac77-db5b6741fb89" /> |

**Build Details:**  Nightly 1d86817

---

See #740 for updates

Arklaum:
@aaronleopold Would you actually want a PR with the foliate-js reader? The library hasn't reached stable release + no official npm package yet (I cloned it to `browser/src/vendor/foliate-js`)

I have a working draft in my fork `sea-orm/foliate-js` if it's of interest (no library included)

aaronleopold:
> @aaronleopold Would you actually want a PR with the foliate-js reader? The library hasn't reached stable release + no official npm package yet (I cloned it to `browser/src/vendor/foliate-js`)
> 
> I have a working draft in my fork `sea-orm/foliate-js` if it's of interest (no library included)

[...]

PR #740, opened 2025-08-31
🐛 Fix Japanese epub text direction

Arklaum:
This fixes issue 1 in #723 by forcing the `direction` to be `ltr`.

Issue 2 is unsolvable with epubjs.

With this I'm 90% happy since all books are readable, so I'll stay off adding a different epub reader

### Comparison: Book 1 (Stump vs Apple Books)

| Stump New Behaviour | Stump Old Behaviour | Apple Books |
| :---: | :---: | :---: |
| New Book 1 | Image | Image |

### Comparison: Book 2

| Stump New Behaviour | Stump Old Behaviour | Apple Books |
| :---: | :---: | :---: |
| New Book 2 | Image | Image |


aaronleopold (approved):
I appreciate you, thanks again!

PR #741, opened 2025-09-01
🎨 Update and tidy fonts

Arklaum:
I wanted to update `Atkinson Hyperlegible` to `Atkinson Hyperlegible Next`. Ended up updating them all. I just downloaded the files and their the accompanying css code from https://github.com/fontsource/font-files. 10.5MB -> 4.2MB (even with more glyph extensions)

The only one I didn't do that way was updating to Charis 7.0, which I got from the official website.

Also fixes OpenDyslexic to actually use the italics and bold files.

---

Also, would it be okay if I added a Japanese font file? I can add a commit to this PR for it. It's 1.4MB for the one I'd like [Hina Mincho (Hina Serif)](https://github.com/satsuyako/Hina-Mincho) (OFL-1.1).

Also also I think default font should be per library rather than global (e.g. I have a JP book library which I'd like to set a JP font for, but for my EN book library I'd like to set a Latin font)


aaronleopold (approved):
> 10.5MB -> 4.2MB

Incredible 🙌 Thank you!

> Also, would it be okay if I added a Japanese font file?

Yeah that's totally fine

PR #742, opened 2025-09-02
🐛 Fix: Allow press and hold on mobile browser

Arklaum:
Allows press and hold on mobile browsers (only tested safari)

Allows press and hold even if mouse pointer goes off the button, useful when the font preview changes size and pushes the button off the mouse pointer (still stops after mouse up)

Made buttons bigger and have hovering/pressing down indicaton. I hope I used the right `variant`, I just copied `ghost` from the other buttons.


aaronleopold (approved):
This looks great, thank you!

PR #744, opened 2025-09-02
🐛 Fix scrolled mode

Arklaum:
Fixes scrolled mode for desktop browsers only ~and makes the useless horizontal scroll bar disappear (only for horizontal-tb books right now)~

When using a book that needs to scroll e.g. downwards, there is an annoying horizontal scroll bar because of the vertical scroll bar taking a tiny amount of space. I couldn't fix it while keeping books that scroll sideways working.

Also fixes an issue with images in books going half outside the browser window if using a tall screen.

---

One more note, I tried to change `readingDirection === 'rtl'` to `readingDirection === ReadingDirection.Rtl`, etc because of the red squigglies annoying me, but that breaks page turn direction because `'rtl'` is still the correct string to use. I'll look at it later.

### DIFF
--- a/packages/browser/src/components/readers/epub/EpubJsReader.tsx
+++ b/packages/browser/src/components/readers/epub/EpubJsReader.tsx
@@ -392,7 +392,7 @@ export default function EpubJsReader({ id, isIncognito }: EpubJsReaderProps) {
 	 * @param preferences The epub reader preferences
 	 */
 	const applyEpubPreferences = useCallback(
-		(rendition: Rendition, readingMode: ReadingMode, lang: string, pageFlipDirection: string) => {
+		(rendition: Rendition, lang: string, pageFlipDirection: string) => {
 			// ja should be ltr no matter what because text is always written "forwards"
 			const isJaWithPageFlipRtl =
 				(lang === 'ja' || lang === 'zh-TW' || lang === 'zh-HK') && pageFlipDirection === 'rtl'
@@ -419,14 +419,6 @@ export default function EpubJsReader({ id, isIncognito }: EpubJsReaderProps) {
 				rendition.themes.register('stump-light', {})
 				rendition.themes.select('stump-light')
 			}
-
-			// Set flow based on reading mode
-			if (readingMode === ReadingMode.ContinuousVertical) {
-				rendition.flow('scrolled')
-			} else {
-				// Default to paginated for 'paged' mode
-				rendition.flow('paginated')
-			}
 		},
 		[theme],
 	)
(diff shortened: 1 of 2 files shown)


Arklaum:
Just added an extra fix to make the arrow keys work, but that is by [making the sandbox insecure](https://github.com/futurepress/epub.js/blob/master/README.md#scripted-content). I will revert it if it's not a good idea.

Edit: I might have a fix actually?

Edit 2: No fix

Edit 3: also allow arrow keys to work even after clicking away (e.g. clicking on buttons)

aaronleopold:
> When using a book that needs to scroll e.g. downwards, there is an annoying horizontal scroll bar because of the vertical scroll bar taking a tiny amount of space. I couldn't fix it while keeping books that scroll sideways working.

If you can inject the styles on the container in the iframe which overflows, you might be able to do one of: 
- Fancy looking alternatives like how I did https://github.com/stumpapp/stump/blob/bb30e0cca4f7be656267f2b30d25bd36258141dd/packages/browser/src/AppLayout.tsx#L44
- Just control the `-webkit-scrollbar`/`-ms-overflow-style`/`scrollbar-width`

Since it's an accessibility concern, though, I'd bind it to the user preference that opts you in to hiding scrollbars.

> One more note, I tried to change `readingDirection === 'rtl'` to `readingDirection === ReadingDirection.Rtl`, etc because of the red squigglies annoying me, but that breaks page turn direction because `'rtl'` is still the correct string to use. I'll look at it later.

I'm not 100% sure where you're referring to here wrt the reader, but it sounds like we just need a basic mapper for epub.js which I think is acceptable

> Just added an extra fix to make the arrow keys work, but that is by making the sandbox insecure. I will revert it if it's not a good idea

Yeah I know it's been discussed on the Discord in the past. Since this is going to unstable I'm good with it for now, but I'm thinking maybe before stable bind it to a server-wide config

Arklaum:
I did a bit more digging and the sandbox is actually a Safari only issue. It was the fix in 1ce33e8 that was half the issue.

`rendition?.on('keydown', handleKeyDown)` works for only Chrome and Firefox when you're focused on the iframe, so 1ce33e8 made it always in focus after turning a page, but this breaks Safari page turning. This now acts like a 'fallback' page turner for if you click on the text.

Since then I've added a `window.addEventListener('keydown', handleKeyDown, { capture: true })` which works for all when you're not focused on the iframe, so specifically Safari will work as long as you don't click on the text. And page turning should now work for Chrome and Firefox if you click on e.g. the book title or after exiting the theme and options settings.

---

I'll look at the scrollbars stuff now and I'll put in another PR if I do anything

aaronleopold (approved):
You rock, thank you again!

PR #745, opened 2025-09-03
🐛 fix auto scaling for image based reader

Arklaum:
Fixes auto scaling mode for `breaking/sea-orm`

Also to me this div seems redundant? Or was it added to fix something because I just messed with it
https://github.com/stumpapp/stump/blob/2e968f91c706380352874a63134b3f0fc525ea92/packages/browser/src/components/readers/imageBased/paged/PageSet.tsx#L55-L59

### DIFF
--- a/packages/browser/src/components/readers/imageBased/paged/PageSet.tsx
+++ b/packages/browser/src/components/readers/imageBased/paged/PageSet.tsx
@@ -47,13 +47,10 @@ const PageSet = forwardRef<HTMLDivElement, Props>(
 				style={{
 					...styles[imageScaling.scaleToFit].imagesHolder,
 					filter: `brightness(${brightness * 100}%)`,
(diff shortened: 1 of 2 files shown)


aaronleopold:
> Also to me this div seems redundant? Or was it added to fix something

Both haha if I'm remembering right, I added it because the animated reader didn't center the set properly without it. You might be able to mess with the animated reader to get a fix in a less redundant way, though, but I took the lazy route 😛

Arklaum:
Hmm I can't seem to replicate it. I doubt what I changed will break anything so I'll leave this PR as is.

aaronleopold:
It was when running in the desktop app IIRC, I'll double check during my lunch today. It is also very explicitly labeled as an experimental reader, so I'm not too concerned about it breaking honestly

PR #748, opened 2025-09-04
💄 Scrollbar colours for Safari and Chrome and 🐛 fix dark epubjs theme

Arklaum:
Makes Safari and Chrome scrollbar (background) colours match the theme the same way Firefox does. I hope the implementation is okay. One annoyance is there was no setting to make it even darker when clicked and held (which Firefox does have for scrollbars and I quite like that), but it didn't work that way previously anyways.

Also a fix to make all dark themes use light text.

---

Note: Chrome does and soon Safari will support the standard [`scrollbar-color`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color) settings but it doesn't have colour change on hover so can't use that.


aaronleopold:
👋 I'll have time to give your two PRs a review tomorrow

aaronleopold (approved):
Looks good to me, thank you!

Issue #711, opened 2025-08-18
[FEATURE] Option to hide/show items in homepage per library


datascientist83:
**Describe the solution you'd like**

A toggle per library to remove/show items from said library to appear on the Home page (Continue reading, Recently added books,Recently added series)

aaronleopold:
I’ve added this to the migration project since this falls under the re-written arrangement features already modeled on that branch. See [here](https://github.com/stumpapp/stump/blob/breaking/sea-orm/crates/models/src/shared/arrangement.rs#L174-L179) if you’re curious

PR #750, opened 2025-09-04
:recycle: / :sparkles: (expo): A lot of mobile app changes

aaronleopold:
I went through the mobile app and changed a bunch of things, a few highlights:

- Redesign the browse screen (I think it's still kinda ugly though tbh, would love feedback)
- Adjust some of the dynamic headers, I had LOTS of issues when using the large header on iOS so nested screens will not have them
- Refactor the search screen to be stacked so that you can go "back" without losing your search. This did require me to eliminate the auto-debounced-search functionality but I think it is overall worth
- Added an overlay which will show the next in series if you try to swipe past the end of a book
   - This does not work on Android and I can't figure out why, so I'll come up with a different solution at another time
- Add to swap some filtered queries to offset pagination because of a bug in the cursor-based logic that I am too lazy to fix right now
- Add a small set of filters and order params for series and books within the browse screen
- Swapped a few of the native dropdown menus to non-native alternatives for Android. The native ones were just really ugly
- I did start to try and improve the ebook reader slightly with a menu and what not, but the callbacks (e.g., `onSingleTap`) don't work lol so I think it'll just have to wait until the swap to another reader

I'm thinking I'll push a build to TestFlight/Play before merging this, gather feedback, merge after


aaronleopold:
> I'm thinking I'll push a build to TestFlight/Play before merging this, gather feedback, merge after

jk there are some server changes which are required, so merging in a moment

Arklaum:
I've been trying out the app and the browse screen really looks like Apple Music! Also I will probably only ever click on the 'libraries' tab. I really don't mind what the design ends up being, I'm just being picky but I wish I could get to my books via the libraries quicker. For me it goes 1 tap the server, 2 tap browse, 3 tap libraries, 4 tap manga, 5 tap a series, 6 tap the volume, 7 tap Read. I'd like to just tap manga instantly the same way I do in the browser.

'Set as default server' didn't seem to save either.

I'm just gonna throw some ideas out:

- I think libraries should have a really darkened book cover with the name justified and aligned center in big letters so it's distinct from a book. 
- and maybe for series have a darkened book cover but the name is at the bottom (still inside) and with some text saying like (1/7 volumes read) or something to really highlight that it is a series.

---

I was having a look at some designs and I came across a "Netflix UI clone" I kinda liked and modified it a bit: https://jsfiddle.net/TempUsername120/5m6sjpx4/ (squeeze it to look like a phone).

In that design where it says 'Comics' and 'Manga' I'd imagine you would have personally set "quick links" to the specific libraries or the 'books' tab, 'smart lists' tab, or a series you're reading.

I think the "Jump Back In" bit should be like a carousel with the last 5 books you read, then you don't need to waste the space with "continue reading".

Also personally I almost never find anything in the "home" section of any self hosted media service useful. "Recently Added Books/Movies/TV Episodes" is always overwhelming for me. The two things I'd probably be most interested in:
- A "First Unread Volume from a Recently Updated Series" where, e.g. I've just added volume 5 and 6, so it shows volume 5 of that series. And maybe unlike what I did above have it be exclusive to series you already have some read books in.
- and a "Recently Added Series".

aaronleopold:
> I've been trying out the app and the browse screen really looks like Apple Music! 

Haha nice, I was actually inspired by [Manet](https://tilosoftware.io/manet) for that one! Maybe they were inspired by Apple Music, themselves 😄 I think my biggest beef with what I have here is maybe the icons? Maybe?

> I'm just being picky

I love being picky with software, if it makes for a better UI keep being picky

> I wish I could get to my books via the libraries quicker. For me it goes 1 tap the server, 2 tap browse, 3 tap libraries, 4 tap manga, 5 tap a series, 6 tap the volume, 7 tap Read. I'd like to just tap 'manga' instantly the same way I do in the browser.
> 
> 'Set as default server' didn't seem to save either.

Yeah so I think, even if 'set as default' _was_ working for you to remove the first click, there are things we can try to alleviate the remaining clicks: 

1. Stump tracks the last time you visited a library (on web, at least). I could look into porting that somewhere for mobile
2. I added support for favorite libraries, series, and media (individually). We can maybe utilize that for a truncated version of the (not implemented) Favorites tab on the home screen?
3. I haven't really built out the home arrangement features (e.g., https://github.com/stumpapp/stump/issues/711) but that would give folks options to say "I want this or that on my home screen"

Generally I'm all for customization and removing unnecessary clicks, so appreciate the ideas! I did like the example you shared! I think it might be difficult to put more data (e.g., `1/7 volumes read`) inside the images unless the images are rather large. I'm also learning that working with gradients on Android specifically is a little finicky. I'll try and have a play with it next time I work on it, though.

> In that design where it says 'Comics' and 'Manga' I'd imagine you would have personally set "quick links" to the specific libraries or the 'books' tab, 'smart lists' tab, or a series you're reading

I think favorites on the home screen could cover this?

 
> I think the "Jump Back In" bit should be like a carousel with the last 5 books you read, then you don't need to waste the space with "continue reading".

I love this, good idea. Thank you!

> A "First Unread Volume from a Recently Updated Series" where, e.g. I've just added volume 5 and 6, so it shows volume 5 of that series. And maybe unlike what I did above have it be exclusive to series you already have some read books in.

This kinda sounds like an 'On Deck' kinda view, right? I think having this makes sense

> and a "Recently Added Series".

That's an easy one, I'll be sure to at least add that soon, too

aaronleopold:
Not sure if you're setup for mobile development, but I took a stab at some of these things in https://github.com/stumpapp/stump/compare/sea-orm/more-mobile-improvements if you're curious and want to take a look

Arklaum:
> I think my biggest beef with what I have here is maybe the icons? Maybe?

What I see is:

1. Huge annoyance: the icons have different thicknesses (see Favourites and Series vs others)
2. Maybe add some colour
3. If you look at Apple Music, interestingly the lines start at the very right edge of the screen and stop at the left side of the text and don't go under the icons. Maybe I'm used to it but I like this "less cluttered" / "less divided" feeling? idk. And it's properly separated at the right edge because it goes all the way to the edge.
4. I also like chevrons, they kind of scream **_tap here_**. Again maybe I'm just used to them. 
5. I like feedback when tapping / holding before releasing, e.g. in Apple music the whole bar darkens, and in Manet the text darkens.
6. There is a line missing from under Smart Lists
7. I don't have much to say about the actual icons used, they mostly feel like rectangles with lines in them. Kind of indistinguishable at a glance? It might just be about getting used to them (plus adding a colour). Even the heart icon feels rectangular.
8. Not sure about Manet (I don't have music in Jellyfin) but in Apple Music the albums shown are specifically sorted by "Recently Added", unlike Stump where it is just alphabetical. Maybe for me personally if I had that I might not even want to add the "Recently Added Series" scroller thingy to my home page because I'd already have it.

| **Stump** | **Apple Music** | **Manet** |
| :---: | :---: | :---: |
| Image - | Image | Image - |

> Stump tracks the last time you visited a library

Huh I never noticed, what does it do with that info?


> working with gradients on Android specifically is a little finicky

:( I quite like the dark gradients with text on them because it uses 2/3rds of the space and looks cleaner (to me). Though I do like where on the website it packages the image with the underneath text as a single poster.

>>"quick links" to the specific libraries or the 'books' tab, 'smart lists' tab, or a series you're reading
>
> I think favorites on the home screen could cover this?

Maybe? What I wanted was just the names of the libraries with no images (1 to save space and 2 to not be confused with a book). I definitely think there should be an option for favourites on the home screen, but if I favourited a lot of things, it may not be usable for my use case of: I just want to easily tap to "Manga", "Comics" and "Novels".

Also, if there is a favourites horizontal scroller, how will it handle libraries, series and books differently, because they might all have the same image and be confusing.

> This kinda sounds like an 'On Deck' kinda view, right?

Oh yes I think it's similar. I thought I saw 'On Deck' somewhere, so I searched and it was in the iOS app Infuse (for Jellyfin) and it is similar to what I'm looking for (i.e. Doesn't show TV episodes with partial watch, only the next episodes where the previous was completed). The difference is I wanted to see it sorted by recently updated rather than recently watched. I'm not sure but if I wanted to read the next volume, I might like it in the "Jump Back In" section where if you've finished a book it'll use the next book. Or probably a dedicated "Your Next Read" which is the exact same but sorted by recently read.

I also like the "See All" button it has for the horizontal scroller thingies. Especially for On Deck it would work nicely as an overview of what I need to keep up with.

---

Oh another thing I wanted to mention was with the "Recently added/updated xyz" I would like to see the relative time added/updated (like in my example) otherwise it's like "overwhelming"? e.g. I don't intuitively know the sort order, and I can easily mentally block out things that are older than say 2 weeks. 

And I like seeing the volume number too (like in my example), not sure how you should get that info. It's because some series have long names so each book just has the same visible name, and with digital books covers it isn't that easy to see at a glance which volume it is (physical books on a shelf you just see the thin spine which has way fewer visual elements than a cover so its just easy to see the number / they're thin so its easy to count)

> Not sure if you're setup for mobile development, but I took a stab at some of these things

I will definitely get set up for it and take a look! Thank you so much!

PR #754, opened 2025-09-07
🐛 Fix inability to save library changes

Arklaum:
Fix inability to save library changes (e.g. update name, path) when thumbnail generation is enabled.

Fix no error message showing up when using already taken names / the path is a parent directory of a library. Hope I did this correctly.

Fix the spacing between the `Name` and `Path` components. ~Not sure what that div was for.~ Edit 2: oops figured it out. The spacing issue was with the vertical spacing when using a portrait screen. Just changed it to `md:gap-y-6` instead.

---

I didn't change the logic in `schema.ts`, I just made it readable to myself. But shouldn't there also be a check that we aren't creating a library that is a child to an existing library?

edit: I think I messed up an import sort 1 sec (ok fixed)


aaronleopold (approved):
Thank you!

> But shouldn't there also be a check that we aren't creating a library that is a child to an existing library?

That should be what this block is doing:

```ts
// return falsy value to indicate failure.
// If the path is a parent to any existing library -> fail
// If the path is not changing -> pass (override the fail)
(val) => {
	const isParent = existingLibraries.some((l) => l.path.startsWith(val))
	const isUnchanged = library?.path === val
	return !isParent || isUnchanged
},
```

But also, the backend will catch that too. This is more of a convenience than anything on the frontend so you don't go through the whole process only to have to backtrack to change the path.

Arklaum:
Like shouldn't we add
 ```ts
const isChild = existingLibraries.some((l) => val.startsWith(l.path))
 ```
and update the warning to say whatever about being a child

aaronleopold:
OH I understand, yeah that should be added

PR #774, opened 2025-09-24
:recycle: Reshuffle the metadata fields into somewhat proper order 

dj-bauer:
Just a small thing, that was bugging me about reshuffling the metadata order, which makes editing them easier for me

Now they are ordered in a more semantic way following the rough order
- Name
- Summary
- publishing date
- information about series and publisher
- Editors, colorists etc
- Mixed information
- IDs for other platforms as amazon, calibre etc.


aaronleopold (approved):
Thank you!

PR #775, opened 2025-09-25
:sparkles: (expo): POC Readium reader and :recycle: Refactor SeaORM initial migration

aaronleopold:
I'm hoping this will be the **very last** fully breaking change to the initial migrations file I make to the `unstable` image. I refactored it to be more manually programmatic since I realized that by using the actual entity modules I chance screwing up the initial just by changing the crate (which is bound to happen). So I'll leave this in draft while I do further testing.

The other big thing of note is that I have a scuffed but minimally functional Readium reader using native expo modules working for both Android and iOS. There are some features I started to build out around theming which I didn't finish, and I probably won't before this gets merged, but I'll come back and edit this description another time to outline what is changed.


aaronleopold:
Okay so the version of Readium I am using _really_ doesn't like Xcode 26 😬 so this might be longer than I had hoped

aaronleopold:
Okay, I'm calling this good enough! It definitely isn't a fully featured ebook reader _yet_, but it does the basics:

- Ability to select at least one alternative font (Android doesn't seem to like all the fonts I have working for iOS)
- Default 3 different themes (light, dark, sephia)
- It tracks your progress using readium locators (epubcfi not supported, so eventually web should move to readium too for consistency)
- Basic TOC navigation

There are some features like custom theming which I started and didn't finish

Issue #162, opened 2023-09-23
[FEATURE] Support PDF page-to-image configurable options


aaronleopold:
**Is your feature request related to a problem? Please describe.**

Right now PDF pages are blindly converted to a bitmap and then converted to a PNG image. Some PDFs render just fine with this default, but others turn out a bit wonky. 

**Describe the solution you'd like**

A user should be able to configure the PDF page-to-image options so they can control:

- Quality options (upscale, downscale)
- Target format (JPEG, PNG, etc)

**Additional context**

Look at the Alice In Wonderland PDF file on the demo site, the pages when using the image-based reader do not look good. I'm sure, at a minimum, https://github.com/stumpapp/stump/issues/156#issuecomment-1712655256 is also a factor.

aaronleopold:
> Currently PDF pages are rendered low quality by default and there's no knob to adjust that

That's what this issue is for 🙂 I haven't had time to prioritize building this out myself, but would happily accept contributions. Otherwise, I'll try to get to it sometime in 2025. 

The real issue, as I understand it based on conversations with folks more knowledgeable in PDF processing than myself, is that consistent quality for PDF page-to-image rendering is difficult to achieve. That said, it seems that [PdfRenderConfig](https://docs.rs/pdfium-render/latest/pdfium_render/prelude/struct.PdfRenderConfig.html) has some knobs we could probably hook into with config values to hopefully give more control and consistent quality.

The other potential solve would be to use something like PDF.js in the browser and not rely on PDF page processing to output an image.

kirawi:
Have you considered https://github.com/ArtifexSoftware/mupdf.js?

aaronleopold:
> Have you considered https://github.com/ArtifexSoftware/mupdf.js?

I haven't, but if it were to be considered there would have to be Rust bindings readily available since I don't necessarily want to make them myself 😅 

A quick search shows there is at least one, but not sure how viable it might be: https://crates.io/crates/mupdf

kirawi:
I've been playing with those bindings for creating a PDF viewer recently -- and as far as I can tell, there is only one issue with them which can cause incorrect behavior but otherwise it's mostly a copy of the Python PyMuPDF API. One neat thing I've read about it is that it can use some tricks to allow reading parts of a PDF file streamed from elsewhere: https://mupdf.readthedocs.io/en/latest/progressive-loading.html

aaronleopold:
@kirawi Sorry I'm a bit late, I totally missed this:

[...]

PR #777, opened 2025-09-25
Migrate PDF processing from pdf-rs + pdfium-render to MuPDF

hollisticated-horse:
- Replace pdf-rs and pdfium-render dependencies with MuPDF for PDF processing
- Add server-side page rendering cache for improved performance
- Fix transparency issues in layered PDF pages by setting alpha: false
- Add support for WebP and JPEG output formats in addition to PNG
- Implement configurable image format conversion (PNG → WebP/JPEG)
- Add in-memory cache for rendered pages with LRU-style eviction
- Harmonize code style and error handling throughout PDF processor

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,24 +14,18 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.19.0"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a76fd60b23679b7d19bd066031410fb7e458ccc5e958eb5c325888ce4baedc97"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
 dependencies = [
  "gimli",
 ]
 
 [[package]]
-name = "adler"
-version = "1.0.2"
+name = "adler2"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
-
-[[package]]
-name = "adler32"
-version = "1.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "aead"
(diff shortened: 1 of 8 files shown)


hollisticated-horse:
This should be correctly breaking/sea-orm with the mupdf commit cherrypicked on top. 

There might be a need to tidy up those commits. A lot of commented out tests were completely removed, which is not ideal. 

I'll await your opinion on this :grin:.

hollisticated-horse:
And it builds, and works.

aaronleopold:
> +68,272

I almost choked when I saw this haha Can you remove the `prisma.rs` which got checked in?

hollisticated-horse:
> I almost choked when I saw this haha Can you remove the `prisma.rs` which got checked in?

Oh no, sorrrrryyy. I removed it and squashed the commits for ease of reviewing.

Arklaum:
I wanted to ask, how do the licenses work because that's AGPL and I thought it would make the whole thing AGPL?

aaronleopold:
> I wanted to ask, how do the licenses work because that's AGPL and I thought it would make the whole thing AGPL? 

Oh good callout, I didn't notice the license. I would have to double check honestly, but that would be disqualifying (I'm not relicensing from MIT to AGPL).

Edit: So I am definitely not a lawyer, and after painstakingly re-reading APGL I still don't know for sure 😅 If I take the conditions listed [here](https://choosealicense.com/licenses/agpl-3.0/):

- Disclose source: Stump is open source, so source code is available
- License and copyright notice: Sounds like attribution?
- Network use is distribution: attribution?
- Same license: This one is tricky, it says "modifications must be released under the same license" but I am not making modifications. I don't know if simply using the library effectively means "modifying" a larger body of work, though.
- State changes: Same as above

All in all, probably a headache better avoided unless I hear otherwise from someone more knowledgeable than me

Edit again: The actual underlying `mupdf` [library](https://github.com/ArtifexSoftware/mupdf) is _also_ AGPL, which I assume is why this crate is too

Arklaum:
So I found this: https://www.gnu.org/licenses/gpl-faq.html#IfLibraryIsGPL which also includes AGPL

hollisticated-horse:
Ah, fudge. That's the first time I'm encounter a license problem. 

From your link @Arklaum, 

> **If a library is released under the GPL (not the LGPL), does that mean that any software which uses it has to be under the GPL or a GPL-compatible license? (#IfLibraryIsGPL)**
>
> Yes, because the program actually links to the library. As such, the terms of the GPL apply to the entire combination. The software modules that link with the library may be under various GPL compatible licenses, but the work as a whole must be licensed under the GPL. See also: What does it mean to say a license is “compatible with the GPL”? 

So the license is an issue. The muPDF repo mentions a commercial license and contacting sales, but that still does not handle the mupdf-rs etc. Well, my muPDF PR goes out of the window...

This is a prime example of a copyleft license at work :grin: . And I'm having second thoughts about some of my own projects licensing, since I usually license under AGPL. 

I'm investigating other solutions.

Arklaum:
What were the issues with the current one? I would think google's pdfium is decent so maybe it just needs some settings tweaked? e.g. boost the resolution and use JPEG to make sure its not too slow if those were the biggest issues.

Or there is this which doesn't require any underlying installs and could be an option once the performance is there (or maybe it is already good): https://github.com/LaurenzV/hayro/tree/master/hayro

aaronleopold:
> What were the issues with the current one? I would think google's pdfium is decent so maybe it just needs some settings tweaked? 

Some of the issues with pdfium are outlined in https://github.com/stumpapp/stump/issues/162, but IMO it could just be a configuration issue (or lack thereof). I haven't tinkered with it enough to confidently assert that opening up config options would correct it, though, in fairness. I'd be curious how much opening those configs up and adjusting the defaults helps

> Or there is this which doesn't require any underlying installs and could be an option once the performance is there (or maybe it is already good): https://github.com/LaurenzV/hayro/tree/master/hayro

I haven't seen this one, thanks! I won't have time to check it out right away (link to an [online demo](https://laurenzv.github.io/hayro/) though I'll try and poke at later), but generally I _love_ the idea of moving to an option that _doesn't_ require some external library. It seems insignificant of a maintenance burden to have to consider, but with all the other things tangled up in maintaining Stump it would be a small but welcome convenience to be able to remove the download/install steps of the dockerfile around pdfium

hollisticated-horse:
Opened a new PR fix pdfium fixes. Hopefull this get's the pdf reader back on track.

hollisticated-horse:
> Or there is this which doesn't require any underlying installs and could be an option once the performance is there (or maybe it is already good): https://github.com/LaurenzV/hayro/tree/master/hayro

I also asked the person behind [oxidize-pdf](https://github.com/bzsanti/oxidizePdf/issues/50#issuecomment-3345808162), who replied they were also working on a renderer, but it was ready yet. 

I'll have a look at hayro, see if I can make a prototype for stump.

PR #779, opened 2025-09-27
✨ (expo): Basic Thumbnail Ratio Selector

Arklaum:
### Overview:

- basic thumbnail ratio selector
- controls overlay gradient fade in and out + smoother header and footer animations
- "fix" carousel flickering (android issue)
- fix some headers being inside the status bar on first opening of a tab (android issue)
- fix bottom sheet snap points

## Details

1. For thumbnail images, `height` is now always calculated based on a given `width`. This preserved the design the best. 
    - I'll try to do something for the browser too if I feel like it. Would it be best to keep it local or make it a setting stored on the server that is used on every login?

3. I've given the bottom sheets a border so I can at least distinguish them from the content behind, but I still don't like the design. Also made them not go over the status bar areas (all but the filters one did), and specifically the filters bottom sheet sits just below the filters so you can switch filter easily.
    - There is a bug though when you open a bottom sheet too quickly after opening a new tab (e.g "Books") it will not disappear when you click on a different filter (not a new issue)

<details>
  <summary> Video: Bottom Sheets </summary>
  <table>
    <thead>
      <tr>
        <th align="center"><strong> Before </strong></th>
        <th align="center"><strong> After </strong></th>
      </tr>
    </thead>
    <tbody>
      <tr>
        <td align="center">

https://github.com/user-attachments/assets/3b180ed4-f692-4a2f-99aa-bb491d205bc1

 </td>
        <td align="center"> 

https://github.com/user-attachments/assets/9997463f-ec2b-4099-a57e-e24e847677bb

</td>
      </tr>
    </tbody>
  </table>
</details>

4. Thought I solved it but the corners issue is not fixed on android: https://github.com/orgs/stumpapp/projects/8?pane=issue&itemId=127980413

    - All the fixes I tried had issues. The underlying problem behind both the corners and flickering is that `<LinearGradient style={{borderRadius: 8}}/>` and  `<View style={{borderRadius: 8}}/>`  seem to have "different" corner rounding on android if the gradient is under BorderAndShadow. Maybe due to anti-aliasing using transparency, making the corners bleed through? It also makes the sides slightly thinner somehow. 

    - I've applied the best fix I had to `ReadingNow.tsx` to stop the flickering carousel. The gradient will cover the border (a new issue) on android but its better than flickering.

---

Side note: You have to use `import { Pressable } from 'react-native'` and not `import { Pressable } from 'react-native-gesture-handler'` if you have `LinearGradient` above `BorderAndShadow`. Otherwise the gradient disappears. Weird.

Also you can add `ios.infoPlist.UIDesignRequiresCompatibility: true` to `app.config.ts` to "fix" the Xcode 26 UI bugs, but I am looking into them.


aaronleopold (approved):
This looks awesome, thank you!

> I'll try to do something for the browser too if I feel like it. Would it be best to keep it local or make it a setting stored on the server that is used on every login?

I think local for now is probably fine. The one thing I could imagine someone asking for is having this kind of setting per-library, like if they have a mix of different formats with different aspect ratios. That would be a good bump in complexity to make that determination, though, so I'd say wait until the feature is asked for IMO 😅 

> There is a bug though when you open a bottom sheet too quickly after opening a new tab (e.g "Books") it will not disappear when you click on a different filter (not a new issue)

Maybe I'm too slow, I can't reproduce. I marked it down in the project, though, thanks!

>You have to use import { Pressable } from 'react-native' and not import { Pressable } from 'react-native-gesture-handler' if you have LinearGradient above BorderAndShadow. Otherwise the gradient disappears. Weird. 

Is this what you're referring to?

<details>
<summary>Maybe the thing</summary>

https://github.com/user-attachments/assets/7111c73f-ade8-48b4-9e02-612a7978a381
</details>

***

I think the borders on the sheet definitely helps, but yeah I find it hard to match the look-and-feel of Apple's native sheet since they do some cool transformations of the content it is overlayed on. I used those for the ebook reader settings modals exactly because of this.

> Also you can add ios.infoPlist.UIDesignRequiresCompatibility: true to app.config.ts to "fix" the Xcode 26 UI bugs, but I am looking into them.

I gave it a try and it didn't seem to do anything, but I didn't fully trash my existing `ios` build folder so maybe I need to do that before. I'll give it another try, thank you!

Arklaum:
> this kind of setting per-library, like if they have a mix of different formats with different aspect ratios. 

Yeah I thought about it a bit, like if you have a favourites row with different aspect ratios you might want to make the height the same instead of the width for horizontal scrolling so they all look even, but for a grid it will be a mess whatever you choose.

> Maybe I'm too slow, I can't reproduce.

Here's a video in case you want to see. Also I just tried on android and the bug seems to instead just close the sheet immediately after you tap it.

<details>
<summary> The bug vs expected behaviour </summary>

https://github.com/user-attachments/assets/ac398f75-f7ff-47e5-bef3-85d6c9ff647b

</details>

> Is this what you're referring to?
> Maybe the thing

I've seen that too, but it's a different not really important thing I wanted to just highlight. That bug though the gradient doesn't actually disappear, it just gets too much lighter? I might be able to make it look better with a bit of janky code.

> I find it hard to match the look-and-feel of Apple's native sheet

https://docs.expo.dev/versions/latest/sdk/ui/swift-ui/#bottomsheet soon :3

> but I didn't fully trash my existing ios build folder

Yeah I did and and it worked for me.

---

If you need to know, the gradients thing I was talking about is: if instead of `Pressable` from `react-native` you swap it with `Pressable` from `react-native-gesture-handler` the gradient just disappears permanently e.g. on the android code where it puts LinearGradient *above* BorderAndShadow. I didn't have to change the import in that file I edited but all the other files I saw import `Pressable` from `react-native-gesture-handler` instead

Also I added some ideas in the Mobile App UX Discussions, the recent one about thumbnail colours is something I would like opinions/approval to do before I touch rust.

PR #782, opened 2025-10-03
:arrow_up: (expo): Upgrade `expo` SDK to 54

aaronleopold:
The bulk of this PR is the upgrade to SDK 54 to enable using the native iOS 26 design patterns. There are also a few bug fixes present:

- Fix a bug where the `Up Next`, `Continue Reading`, and `Your Next Read` sections on the home screen were not the correct height when initially empty and then populated after progression made
- Fix iOS 26 header right interactive elements from filling extra space, an issue that cropped up after upgrade Xcode

I did start roughing out an idea for custom accent color selection but it is incomplete. I would not suggest using it, but if you would like to try then definitely make sure to add the Stump color to the palette before changing.


Arklaum:
This:
https://github.com/stumpapp/stump/blob/11322088446cc662d1b9fbbc3a69fef408cd4fe6/apps/expo/lib/constants.ts#L15-L17
should be this:
```ts
export const IS_IOS_24_PLUS = 
 	Platform.OS === 'ios' && 
 	(typeof Platform.Version === 'number' ? Platform.Version >= 24 : parseInt(Platform.Version) >= 24)
```

since `Number` doesn't work on e.g. iOS 26.0.1

also it could just be 

```ts
export const IS_IOS_24_PLUS = Platform.OS === 'ios' && parseInt(Platform.Version) >= 24
```

since it's always a string for iOS.

aaronleopold:
Thank you, @Arklaum! I don't think the changes here are perfect but I think it's good enough for a merge

PR #770, opened 2025-09-21
💄 (expo): UX Enhancements

Arklaum:
So this is a list of things I aim to put into this PR. I am using this space to track/show my progress. I am currently working on: Done!

## Tasks

- [x] 1. Different opacities when tapped ~(also try out squeeze on tap)~
- [x] 2. Squircles (iOS only)
- [x] 3. `expo-image` + `faster-image` -> `react-native-turbo-image`. (The double page reader does work properly too)
- [x] 4. Fix gradients + borders + drop shadows
- [x] 5. Fix "sharp" linear gradient inside reader
- [x] 6. `react-native-reanimated-carousel`: rounded corners + dark mode doesn't like 0.99
- [x] 7. Browse > Books: The back button is tiny
- [x] 8. Weird corner issues
- [x] 9. Add Android drop shadows
- [x] 10. Use full height for image based reader on iPad + use full height when zoomed in.

### Issues / Notes for each task (numbered)

- 2 **Important:** To use squircles you must add class name `squircle` whenever using rounded corners. Or add `style={{borderCurve: 'continuous', overflow: 'hidden'}}`. 
- 2 Adding style `borderCurve: 'continuous'` (or class name `squircle`) often requires removing `borderRadius` (`rounded-xx`), saving then adding it back to see the update.
- 2 There may be a few fixable visual bugs added, since I did a find and replace of `rounded-` to `squircle rounded-` where `squircle` applies the style `borderCurve: 'continuous'` **AND** `overflow: 'hidden'`. I found and fixed only one: where `rounded-xl` was added for seemingly no use, but the new overflow hidden style caused issues:

<details>
<summary> Example Bug before and after fixing (but look at that nice squircle!) </summary>

https://github.com/user-attachments/assets/da74a0e7-1d1d-43bb-9d6f-6535d7a90761

_I have since fixed the OPDS corner rounding amount too_

</details>

- 3 There's a few places, specifically ~`StackedEffectThumbnail.tsx`, `PublicationFeed.tsx`, `PublicationGroup.tsx` and~ `auth.tsx` that I have no idea how to find and test.
- 4 Android dark mode: flickering issues at the edge of images when swiping on carousel due to the gradients (existing issue and I couldn't fix it)
[...]

PR #786, opened 2025-10-05
🐛 (expo): Fix various RTL reader issues

Arklaum:
This contains various fixes for RTL mode, and of course make it work in the first place (it was a double inversion issue). I'm so happy to finally be able to read my books without the swipe direction confusing me.

It also contains a few visual improvements:
1. fix uneven grid spacing when the number of columns is larger than 3, e.g. on an iPad. Here is [Before](https://github.com/user-attachments/assets/8cdf48af-ff3a-4052-9e54-b1f11e4b5260) and [After](https://github.com/user-attachments/assets/ef43266a-15bc-41c7-9699-aabad26cf8d1) (best viewed by flipping between the two browser tabs). 
2. Remove vertical centering of thumbnails in a library. Side note: the Before in (1.) was actually not technically the Before image, [this is](https://github.com/user-attachments/assets/82094f94-e875-4313-8267-0e533316b90e) but it doesn't include removing the centering so it's hard to see the difference.
3. Made landscape images use the same height as other thumbnails in the footer gallery. Note: need to do that for the browser too.

I saw some TODO's that said swap to `FlashList` for performance improvements, but they don't have any `inverted` prop (which the RTL logic currently relies on and I don't feel like changing it unless I have to). I also saw another comment somewhere else about `LegendList` and those do have an `inverted` prop, so I might give that a go.

I also updated my little list to be easy to tick completed stuff: https://github.com/stumpapp/stump/pull/770#issuecomment-3325662355

I'm curious about if you look at each commit separately or the whole thing at once?


apps/expo/components/book/reader/image/Footer.tsx
```diff
@@ -128,9 +121,15 @@ export default function Footer() {
 
 	const getItemLayout = useCallback(
 		(_: ArrayLike<number[]> | null | undefined, index: number) => {
-			const totalOffset = pageSets
-				.slice(0, index)
-				.reduce((acc, set) => acc + calcSetContainerSize(set).width, 0)
+			const totalOffset =
+				7.5 +
```
aaronleopold, line 125: What is the 7.5 for?

apps/expo/components/book/reader/image/Footer.tsx
```diff
@@ -236,34 +235,49 @@ export default function Footer() {
...
+			} else return idx
+		},
+		[pageSets.length, readingDirection],
+	)
+
+	/**
+	 * A function that takes the slider value and returns the corresponding pageSet index
+	 * It uses the same logic as getSliderValue
+	 */
+	const getPageSetIndex = useCallback((value: number) => getSliderValue(value), [getSliderValue])
```
aaronleopold, line 254: Super unimportant, but I would say these should be "inverted" so that `getSliderValue` invokes `getPageSetIndex` (basically just swapping the names I guess)

apps/expo/components/book/reader/image/Footer.tsx
```diff
@@ -341,63 +359,69 @@ export default function Footer() {
...
 									onSuccess={({ nativeEvent }) => onImageLoaded(pageIdx, nativeEvent)}
 								/>
 							)
 						})}
 					</View>
 
-					<Text className="text-center">{pageSet.map((i) => i + 1).join('-')}</Text>
+					<Text className="text-center">
+						{pageSet
+							.sort((a, b) => a - b) // we always use (from left to right) the smaller then larger number even if using RTL (e.g. pages 3-4 and never 4-3)
```
aaronleopold, line 374: TIL, thank you

aaronleopold (approved):
As usual, this looks great! Thank you for consistently making this app better and better!

> I saw some TODO's that said swap to FlashList for performance improvements, but they don't have any inverted prop (which the RTL logic currently relies on and I don't feel like changing it unless I have to)

They do have guidance for migrating usages of `inverted` from FlashList v1 to v2: https://shopify.github.io/flash-list/docs/v2-migration#step-4-replace-inverted-lists. It does mention manually reversing the data set which is an unfortunate burden on us (but understandable from their perspective I'm sure haha).

> I also saw another comment somewhere else about LegendList and those do have an inverted prop, so I might give that a go.

I did try LegendList before really landing on FlashList and had lots of awkward recycler issues from what I remember, but especially on Android. I think they also released a v2 around the same time as FlashList, though, which I haven't tried yet

> I also updated my little list to be easy to tick completed stuff: https://github.com/stumpapp/stump/pull/770#issuecomment-3325662355

❤️ 

> I'm curious about if you look at each commit separately or the whole thing at once?

I usually give the commits a glance before digging in, but I mostly just look at the whole thing all at once. I think having clean, smaller commits like this makes it really convenient if I encounter something that is easier to process in a reduced scope (i.e., just the commit) but the review tools GitHub provides makes the review overall easier all at once

apps/expo/components/book/reader/image/Footer.tsx
Arklaum: Note: I didn't use the same function just for clarity. I know I could've used 

```ts
const getPageSetIndex = getSliderValue
```

or vice versa but then the input name being `idx` on both annoyed me. I've done the swap, so both would say `value`, which is okay, even though I only intended to use that word for slider value. Would you prefer that?

aaronleopold: No yeah I totally agree with your choice against just doing `const getPageSetIndex = getSliderValue`, this was just a super minor nit. I don't have a _super_ strong opinion either way, just something that stuck out in my brain

Arklaum:
> As usual

❤️ Aww I like those words

> They do have guidance for migrating usages of inverted from FlashList v1 to v2: https://shopify.github.io/flash-list/docs/v2-migration#step-4-replace-inverted-lists. It does mention manually reversing the data set which is an unfortunate burden on us (but understandable from their perspective I'm sure haha). 
>...
>I did try LegendList before really landing on FlashList and had lots of awkward recycler issues from what I remember, but especially on Android. I think they also released a v2 around the same time as FlashList, though, which I haven't tried yet

Well it might be okay then. When I tried removing the `inverted` prop and using inverted page sets there was weirdness and flickering when changing page. I'll give it another look, I'm sure it's fine and I'm just being dramatic.

aaronleopold:
> Well it might be okay then. When I tried removing the inverted prop and using inverted page sets there was weirdness and flickering when changing page. I'll give it another look, I'm sure it's fine and I'm just being dramatic.

Haha I'm sure it's not the latter. Fwiw I do also remember some weirdness and flickering when using FlastList for the reader, in fact I recall writing a comment _somewhere_ something to the effect of "this reader _really_ doesn't like flashlist" which in hindsight is really vague. But also that was v1, I haven't tried again since upgrading.

Also I'd be curious how much benefit using it would even provide. I'd love to measure it, if possible, but I imagine most of the benefit will come to those using Android devices. For whatever reason, react native often just feels more optimized on iOS unfortunately. For instance, I just loaded up my TPB Invincible Universe (660 pages) almost instantaneously with no stutter or lag.

***

I'll aim to merge this shortly 🙂

PR #790, opened 2025-10-07
:memo: Update docs for migration procedure

aaronleopold:
I didn't do a full audit of the docs website since that would be _quite_ a lot of reading, but I at the very least:

- Added a new section (accessible by clicking the announcement banner) to outline the migration process
- Updated mobile app docs to be more accurate to current state of things
- Fix permissions referenced

### DIFF
--- a/apps/docs/pages/guides/_meta.ts
+++ b/apps/docs/pages/guides/_meta.ts
@@ -14,4 +14,7 @@ export default {
 	cli: 'CLI',
 	desktop: 'Desktop App',
 	mobile: 'Mobile App',
+	'breaking-changes': {
+		display: 'hidden',
+	},
 } satisfies Meta
--- a/apps/docs/pages/guides/access-control/library-exclusions.mdx
+++ b/apps/docs/pages/guides/access-control/library-exclusions.mdx
@@ -7,8 +7,8 @@ A user can be explicitly excluded from seeing a library. This is useful for hidi
 ## Configure Exclusions
 
 <Callout emoji="🔐">
-	This functionality is gated behind the `library:manage` and `user:read` user permissions. To learn
-	more about permissions, see the [permissions](/guides/access-control/permissions) guide.
+	This functionality is gated behind the `MANAGE_LIBRARY` and `READ_USERS` user permissions. To
+	learn more about permissions, see the [permissions](/guides/access-control/permissions) guide.
 </Callout>
 
 <Steps>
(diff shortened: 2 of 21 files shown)


Arklaum:
https://github.com/stumpapp/stump/blob/eb21ef4a28516a65a28abb321595664ecceeaf2c/apps/docs/pages/guides/mobile/reading/image-based.mdx#L44-L51

So I didn't realise/notice there was another cache policy select inside the reader settings (I replaced the one that was in the debug options with those clear memory/disk cache buttons). It's actually non-functional since turbo-image doesn't support that level of fine grain control. I don't see how slow devices can benefit from the options, but if it's needed I can swap the image loader to something else? Otherwise I'll clean up the setting. 

But only turbo-image, expo-image and nitro-image have downscaling support which helps stop aliasing artifacts. And expo-image's won't work while using stretch mode which helps make inconsistencies between thumbnail ratios look better, e.g. many publishers using "1 : √2" (~1 : 1.414) books usually use either 1 : 1.404 or 1 : 1.424.

---

I would eventually want to use [nitro-image](https://github.com/mrousavy/react-native-nitro-image): Most image loaders pair SDWebImage (iOS) with Glide (Android), or pair Nuke (iOS) with Coil (Android), I assume because the latter are more modern, but in fact SDWebImage is faster than Nuke so nitro-image uses SDWebImage and Coil (I also saw this when playing with expo-image vs faster-image). It is also the only other image library with custom downscaling size. But it's new so doesn't have custom header support, cache management, or fade in on load.

aaronleopold:
> I don't see how slow devices can benefit from the options, but if it's needed I can swap the image loader to something else? Otherwise I'll clean up the setting.

Nah, I say let's just remove the feature entirely. I added it a while back on a hunch when someone reported higher memory usage on a lower powered Android tablet. I was hoping that being able to offload the cache to disk might save on runtime memory but I have no data to back up it actually worked

I'll make sure to remove those bits from the docs. Also side note thank for you giving it a look, I appreciate it!

> I would eventually want to use [nitro-image](https://github.com/mrousavy/react-native-nitro-image)... But it's new so doesn't have custom header support, cache management, or fade in on load

Once the library has the features we would need I'm totally okay giving it a try if you're interested in trying it out

Arklaum:
I would also say the thumbnail bit needs a bit of clarity, just to tell people what to use

https://github.com/stumpapp/stump/blob/eb21ef4a28516a65a28abb321595664ecceeaf2c/apps/docs/pages/guides/mobile/settings.mdx?plain=1#L23-L29

Maybe bad wording but something like:

```md
 - **1 : 1.6** - Taller, an uncommon but standard ebook aspect ratio.
 - **1 : 1.5** - Default, the standard aspect ratio for western books and American comics.
 - **1 : √2** - Shorter, another common aspect ratio for books and the standard aspect ratio for Japanese manga.
 ```

I chose those three because they're the top 3 ones, though I doubt `1 : 1.6` will get any use.

PR #793, opened 2025-10-08
🐛 (expo): Various small bug fixes

Arklaum:
This includes various small bug fixes, all named in each commit. Also I added React Native Reanimated's performance monitor to the reader as a debug toggle. It's super basic and I'm not sure if it's useful yet but it's something.

There is a small visual change in the reader, which is making double spread thumbnails look the same as landscape pages: 

<details>

<summary> Landscape and Double Page design </summary>
<table>
  <thead>
    <tr>
      <th align="center"><strong> Before </strong></th>
      <th align="center"><strong> After </strong></th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td align="center">

Before

</td>
      <td align="center"> 

After

</td>
    </tr>
  </tbody>
</table>

</details>

You can see the pages on the left side of the footer now look the same as the ones on the right. In my experience any landscape "page" has been a real double spread, so I think they should look the same. I think it looks better when scrolling too, but let me know your thoughts.

---

Side note, I was playing around with flashlist for the reader and footer, and I did get it to work but the issues were:
1. changing reading direction will cause flickering (since the data changes you see the page on the opposite side of the book, and then you make it scroll back to the correct page)
2. For the reader you have to wrap the flashlist in a view with explicit height and width so the pagination works.
3. I did actually get it to work without that flickering issue by instead: when RTL use `scaleX: -1` over the flashlist and again over each page set, but the `<Zoomable>` controls for panning became inverted, but it might have an easy fix of just modifying the library itself with some inverted panning settings.

There is the same flickering issue (with flatlist too) when changing from double page mode to single page, but thankfully it's hidden by the bottom sheet, though the same issue is not hidden when changing orientation. Surely there must be a better way to do double pages so this doesn't happen, but I'm not sure what.

I was also wrong when I said legendlist had `inverted` prop, it doesn't, I misread the title of a docs page. Also in one of the commits I removed a `View` that was wrapping a `BottomSheet.Modal` and that does nothing (I think?), except it fixes an issue if you use flashlist or legendlist (it stops the page sets showing only on the bottom half of the screen).

### DIFF
--- a/apps/expo/app/server/[id]/(tabs)/search/[query].tsx
+++ b/apps/expo/app/server/[id]/(tabs)/search/[query].tsx
@@ -3,7 +3,7 @@ import { graphql, PaginationInfo } from '@stump/graphql'
 import { useQueries, useQueryClient } from '@tanstack/react-query'
 import { Link, useLocalSearchParams } from 'expo-router'
 import { useCallback } from 'react'
-import { FlatList, Platform, View } from 'react-native'
+import { FlatList, View } from 'react-native'
 import { ScrollView } from 'react-native-gesture-handler'
 import { SafeAreaView } from 'react-native-safe-area-context'
 import { useActiveServer } from '~/components/activeServer'
@@ -155,10 +155,7 @@ export default function Screen() {
 	}
 
 	return (
-		<SafeAreaView
-			style={{ flex: 1 }}
-			edges={Platform.OS === 'ios' ? ['top', 'left', 'right'] : ['left', 'right']}
-		>
+		<SafeAreaView style={{ flex: 1 }} edges={['left', 'right']}>
 			<ScrollView
 				className="flex-1 bg-background py-4 tablet:py-7"
 				contentInsetAdjustmentBehavior="automatic"


Arklaum:
>  I'm curious how v1 implemented the inverted behavior and if it would be possible to replicate that.

I tried `scaleX: -1` since that's what flashlist v1 did https://shopify.github.io/flash-list/docs/1.x/usage#inverted and flatlist currently does. Since `scaleX: -1` over the flashlist worked to invert the pan direction, and panning is currently fine with flatlist, I'm sure there's something I missed in the second inversion that should invert it back.

edit: the solution was right in front of me, I literally said 
> wrap the flashlist in a view with explicit height and width so the pagination works. 

I just had to wrap the Zoomable with explicit height and width and the panning works

aaronleopold:
> I just had to wrap the Zoomable with explicit height and width and the panning works

Nice! Happy you got it working. I'll be curious to hear about your observations after you tinker more and see if it meaningfully improves the reader

***

I'm going to merge this in shortly, thanks again!

PR #803, opened 2025-10-15
🐛 fix android long press on server

Arklaum:
Stops android long pressing on server opening the server (since it's supposed to just bring up the context menu)

Also just a bit more rounding of stuff and removing a weird extra horizontal line in ReaderSettings

### DIFF
--- a/apps/expo/app/(tabs)/settings/usage/[id].tsx
+++ b/apps/expo/app/(tabs)/settings/usage/[id].tsx
@@ -60,10 +60,7 @@ export default function Screen() {
 	if (isLoading) return null
 
 	return (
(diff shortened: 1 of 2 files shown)


apps/expo/components/savedServer/SavedServerListItem.tsx
```diff
@@ -50,6 +50,7 @@ export default function SavedServerListItem({ server, onEdit, onDelete, forceOPD
 								},
 							})
 						}
+						onLongPress={() => {}}
```
aaronleopold: Ha, nice. Love that this was the fix

PR #805, opened 2025-10-17
🐛 (expo): fix reading timer memory leak

Arklaum:
This fixes the reading timer by not using `useStopwatch`.

Side note: I was wrong when I said the reading timer updated the reader store every second, so the reading timer memory issue is unrelated to the reader store one, and was more fundamental to `useStopwatch` (hence its removal). So with this I guess there is still the performance monitor and the reader store, but the latter is minor.

### DIFF
--- a/apps/expo/stores/reader.ts
+++ b/apps/expo/stores/reader.ts
@@ -1,8 +1,6 @@
 import { BookPreferences as IBookPreferences } from '@stump/client'
 import { ReadingDirection, ReadingImageScaleFit, ReadingMode } from '@stump/graphql'
-import dayjs from 'dayjs'
-import { useCallback, useEffect, useMemo, useState } from 'react'
-import { useStopwatch } from 'react-timer-hook'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { create } from 'zustand'
 import { createJSONStorage, persist } from 'zustand/middleware'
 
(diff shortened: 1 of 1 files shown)


aaronleopold (approved):
This looks great, thank you!!

Arklaum:
edit: Ignore my prev unedited message, let me check any changes and if they're needed I'll add it to any next PR. 

Also, the browser uses the same old logic, but doesn't have the memory issues. Should it be changed there anyways?

And there seems to be an existing issue with the read time:
1. When there is a read time in the book overview (`books/[id]/index.tsx`) it is different to the one in the reader, and when I hit reset timer it will show 0 seconds but then if you dismiss the controls overlay then bring it back up it shows the time that was in the overview
2. When there is no read time in the book overview that's when hitting reset actually resets to zero (literally the same issue as 1)

aaronleopold:
> Also, the browser uses the same old logic, but doesn't have the memory issues. Should it be changed there anyways?

I'd say let's leave it until the other issues are fully resolved, then it could be ported over

> When there is a read time in the book overview (books/[id]/index.tsx) it is different to the one in the reader

I think the overview uses relative/humanized time, but would have to double check, if that's what you mean? If it is totally different though then yeah that's not expected.

Otherwise, I should have time tomorrow to try and take a closer look after wrapping up some of the other fixes I've been working on locally. If you get to it first though that's totally fine, I'll be curious what you find

PR #808, opened 2025-10-22
✨ Basic browser thumbnail ratio selector 

Arklaum:
So this includes:
- a thumbnail ratio selector for the browser (a user preference). I call it basic because I have a few ideas to make it better: e.g. custom ratios, analysis to pick the best, a screen to show a sample book with the ratio on it, etc.
- language support for the browser footer read time (using `luxon` instead of `dayjs`) + minor formatting change (e.g. '1 minute 0 seconds' instead of '60 seconds')
- a fix for the image size getter (landscape pages were paired with portrait pages when using double paged mode on `Auto` and `Width` scaling)
- update the browser footer double spread / landscape pages to look like the recent changes to the expo app reader footer
- fix the alignment issues with Virtuoso when the different height image (the current page set) goes out of the viewport, and specifically on Safari when new images load in

Is thumbnail stuff I've so far done okay? Also I definitely missed some places. And I haven't touched `backfill.sql` either (I assume no need?).

Also in `ReaderFooter.tsx` we don't need `List` anymore but I wrote a comment there about `'flex'` not working:
https://github.com/stumpapp/stump/blob/1f4c8f16caaf18ce0038eb8c2903154d559d31d4/packages/browser/src/components/readers/imageBased/container/ReaderFooter.tsx#L226-L235
~I'll remove the component later.~ edit: I have just removed the component 

It's nearly ready, but I wanted feedback in case the thumbnail ratio bit needs some extra stuff.

### DIFF
--- a/packages/browser/src/components/readers/imageBased/container/ReaderFooter.tsx
+++ b/packages/browser/src/components/readers/imageBased/container/ReaderFooter.tsx
@@ -56,9 +56,11 @@ export default function ReaderFooter() {
 		const duration = Duration.fromObject({ seconds: elapsedSeconds }).reconfigure({ locale })
 
 		let formattedDuration
-		if (elapsedSeconds <= 59) formattedDuration = duration.shiftTo('seconds')
-		else if (elapsedSeconds <= 3599) formattedDuration = duration.shiftTo('minutes', 'seconds')
-		else {
+		if (elapsedSeconds <= 59) {
+			formattedDuration = duration.shiftTo('seconds')
+		} else if (elapsedSeconds <= 3599) {
+			formattedDuration = duration.shiftTo('minutes', 'seconds')
+		} else {
 			const hms = duration.shiftTo('hours', 'minutes', 'seconds')
 			formattedDuration = Duration.fromObject({ hours: hms.hours, minutes: hms.minutes })
 		}


aaronleopold:
Hey! I should have time over the weekend to give this a proper look 🙂

packages/browser/src/components/readers/imageBased/container/ReaderFooter.tsx
```diff
@@ -36,83 +39,104 @@ export default function ReaderFooter() {
...
-				.format(`H [hour${elapsedSeconds >= 7200 ? 's' : ''}] m [minutes]`)
+		const duration = Duration.fromObject({ seconds: elapsedSeconds }).reconfigure({ locale })
+
+		let formattedDuration
+		if (elapsedSeconds <= 59) formattedDuration = duration.shiftTo('seconds')
+		else if (elapsedSeconds <= 3599) formattedDuration = duration.shiftTo('minutes', 'seconds')
+		else {
+			const hms = duration.shiftTo('hours', 'minutes', 'seconds')
+			formattedDuration = Duration.fromObject({ hours: hms.hours, minutes: hms.minutes })
 		}
```
aaronleopold: Can we use curly braces here? Whenever there is more than one conditional block and/or blocks that don't just return values I find it more readable and less error-prone

aaronleopold (commented):
> It's nearly ready, but I wanted feedback in case the thumbnail ratio bit needs some extra stuff

I think this looks great! Thank you for all of these fixes alongside the feature!

*** 

> language support for the browser footer read time (using `luxon` instead of `dayjs`)

TIL about `luxon`, nice

> Is thumbnail stuff I've so far done okay? Also I definitely missed some places

Yeah I don't see any issues with what you have so far, I haven't been able to run the changes locally but the changes themselves look good to me. I'll make time tomorrow to give it a run locally. 

And I figure some places will go missed, it's easy to do with a larger-scale change like this. I'm hoping to have time to finally promote the `breaking/sea-orm` branch to `experimental`, and we'll have at least a month before I'd be comfortable moving it to `nightly`. This is to say, got lots of time.

> And I haven't touched backfill.sql either (I assume no need?)

There shouldn't be a need, the `backfill.sql` only cares about taking data that existed _before_ the breaking changes and converting as much as it can into the new database. Since this didn't exist, it's totally fine (and makes sense) to omit.

***

I know you mentioned wanting feedback so I didn't explicitly "approve" the PR. Just give me a ping when you'd like me to take another look or merge

Arklaum:
It should be ready now, since I believe I've updated all the places that need `thumbnailRatio` since my initial comment, so hopefully it's all done!

There may or may not be more "hidden" issues like the one with the horizontal list height in `HorizontalCardList.tsx` where there was no explicit dependence on the aspect ratio, so we will see.

PR #811, opened 2025-10-28
:sparkles: (expo): WIP offline reading support

aaronleopold:
This PR adds really experimental support for offline reading and progress syncing:

- Add SQLite to manage more complex relations wrt downloaded files, progression, series/library references, etc
- Refactor the two primary readers to not make online-always or server-scoped assumptions (e.g., always pulling an SDK instance, using active server context, etc)
- Implement a new native expo module to support streaming pages from zip archives
   - No RAR support ~or PDF support~ for now
- Build out entire downloads UI
   - Basic list with a few sorting options
   - Basic (single action, for now) selection state
   - Port a couple of the curated sections found in the server home page (e.g., large format reading now + currently reading)
      - This is optional
   - Add sync action that will attempt to sync any unsynced progress to corresponding server(s)

This PR isn't quite ready, but I'm super excited to get it out the door. There is a lot of testing to still be done, and some missing bits around space management I'd ideally add before merge. Also, frankly, the Android UI isn't as good as iOS and I'd like to improve it a bit.

### DIFF
--- a/core/src/filesystem/scanner/utils.rs
+++ b/core/src/filesystem/scanner/utils.rs
@@ -459,7 +459,7 @@ pub(crate) async fn safely_insert_series(
 		// metadata entry vs killing the entire series creation process over a single bad entry
 		if let Some(mut meta) = metadata {
 			meta.series_id = Set(created_series.id.clone());
-			if let Err(error) = meta.insert(conn).await {
+			if let Err(error) = meta.insert(&txn).await {
 				tracing::error!(?error, "Failed to insert series metadata");
 			}
 		}


Arklaum:
So I tried this and it's beautiful! I have a few layout/UI issues from my brief look but I'll write about them later after a more thorough look. Also those SF cloud symbols "may only be used to refer to Apple's iCloud service" says the SF Symbols app.

I have a new Performance Monitor (with RAM and CPU usage) and a FlashListed reader+footer (which I have definitively determined is better than FlatLists), would you like them here, or I can wait? There will be merge conflicts in ImageBasedReader and stuff and I can deal with it

aaronleopold:
> So I tried this and it's beautiful! I have a few layout/UI issues from my brief look but I'll write about them later after a more thorough look

Yay! Happy you like it! I'd love all the feedback once you give it a good look, I trust your judgement 

> Also those SF cloud symbols "may only be used to refer to Apple's iCloud service" says the SF Symbols app.

Part of me is annoyed lol but I guess I understand wanting to keep your system icons consistent wrt their intent and usage. Thank you for calling it out! I spent a good chunk of time looking through the symbols app and totally missed that message. I'll make sure to swap them out before merge. If you have any suggestions while poking around let me know

> I have a new Performance Monitor (with RAM and CPU usage) and a FlashListed reader+footer...  would you like them here, or I can wait?

I'm good with it going here if that makes it easier to deal with conflicts, this won't be merged until the earliest this weekend anyways

Arklaum:
So I really like the `DownloadRowItem` components, they look really nice and I actually have a feature request for a Reading Timeline / Journal that I'm writing up that could do with a similar looking item. The feedback is an extremely ios point of view, so I might say things impractical for android or need adjusting.

## The Downloads Overview
### Things I'm more sure about
1. This is probably my strongest issue. In the sort menu, it should be much clearer what the sorting option actually does. I absolutely despise the words "ascending" and "descending" even though they are ubiquitous, because I always have to tap and then check to see it's what I wanted. But Apple doesn't seem to use them in Apple Music and Podcasts, for example if you look at [the menu button in Apple Music](https://github.com/user-attachments/assets/254f820f-cc90-4a34-b42e-5cdacb039876), the subtitle indicates the sort direction, and tapping flips the sort direction, so we'd do this:
    ```ts
    <NativeDropdownMenu.CheckboxItem
	    value={sortConfig.option === 'NAME'}
	    key="sortByName"
	    onSelect={() => {
		    // if already 'NAME', swap the sort direction
		    if (sortConfig.option === 'NAME') {
			    setSortConfig({
				    option: 'NAME',
				    direction: sortConfig.direction === 'ASC' ? 'DESC' : 'ASC',
			    })
		    }
		    // else always first enable the most sensible sort direction
		    else {
			    setSortConfig({ option: 'NAME', direction: 'ASC' })
		    }
	    }}
    >
	    <NativeDropdownMenu.ItemTitle>Name</NativeDropdownMenu.ItemTitle>
	    {sortConfig.option === 'NAME' && (
		    <NativeDropdownMenu.ItemSubtitle>
			    {sortConfig.direction === 'ASC' ? 'A-Z' : 'Z-A'}
		    </NativeDropdownMenu.ItemSubtitle>
	    )}
	    <NativeDropdownMenu.ItemIcon ios={{ name: 'character' }} />
    </NativeDropdownMenu.CheckboxItem>


    ```
    And there is this behaviour where the first tap always uses a default sensible option, I assume for muscle memory (e.g you want `'Z-A'` for Name (Apple uses "Title" not "Name" but whatever) after changing from Recent you will always tap once, reopen the menu and tap again), and best to keep this first tap behaviour consistent with other apps.

    And for 'Recent', I had no idea what that meant until I saw the new docs update say it's "download date". I think it should either be "Recently Downloaded" or "Date Downloaded" (Apple Podcasts uses this name) and Instead of `'A-Z'` it's `'Newest First'` and `'Oldest First'` (that's what Apple does in Music and Podcasts).

    Small side rant about "ascending" and "descending" and other things. For me, if I wrote the alphabet on a vertical line, 'A' would be at the top and I'd "descend" down to get to 'Z', but "ascending" means the opposite. If you think of A=1 and Z=26 with ascending meaning smallest to largest then that makes sense, but I don't think that way. These terms are everywhere with that meaning so what can I do. It reminds me of video game inverted controls, flight sticks, and MacBook trackpads. Like if you put a joystick on the back of a characters head, to make them look up and right, you'd move the stick down and left, but that's not the convention. And similarly MacBook trackpad swipe gesture scroll directions are inverted compared to windows laptops. Another side comment but it's also why I have issues with the [LTR-RTL icon used in the reader menu](https://github.com/user-attachments/assets/efabcc01-8f61-4c9f-afcc-df72a76c7679), arrow left currently represents RTL but I'm often thinking in terms of swipe direction / the direction I move my hands after grabbing a physical page to turn it, and those actions for an RTL book are from the left side to the right side, which if I drew as an arrow, would be an arrow pointing right. I wish it were an open book with a folded left or right corner as the icon, that feels unambiguous to me but not sure what's actually possible.

2. In the menu button, the order should be Select, Attempt Sync and then Hide Curated, though I have more thoughts on the menu later.
3. The carousel should be edge to edge (it's been moved to the right due to padding)
4. The`DownloadRowItem` uses `resize={width * 1.5}` but should be `resize={(width / 2) * 1.5}` since that's the actual width of the image + use a 1.5x higher res image for extra sharpness. I could change the component to instead accept a `size={{width, height}}` and `downscaling={{enabled, scale}}` if that helps keep things more manageable.
5. Also in `BookSearchItem` I used a `borderRadius=4` because those are small thumbnails, and `DownloadRowItem` has similar sized thumbnails so should too. I think there are 3 different borderRadius values in use, one for small `4` medium `8` and large `12`.

### Things I'm less sure about
6. The page numbers + size should be sitting directly above the progress bar and not hanging directly below the title. Though for items without progress bars, I guess they'd need one.
7. the page we're currently on, and maybe [the % number at the end of the progress bar like this](https://github.com/user-attachments/assets/bca6261c-1ca6-4ecf-8f6a-d57143db8c0a) or above.
8. Would rounding 1dp for the size be fine? I mean like 260.1MB instead of 260.11MB. I know epubs are gonna be small though.
9. Related to (5) I think there's currently like 5 different thumbnail widths in use, and `DownloadRowItem` uses a width of 67.5 which makes it a 6th thumbnail width. Nothing wrong with that but `BookSearchItem` has width of 75 and looks similar so maybe a change is warranted in one or the other, but there really isn't any problem here.
10. I think 'Hide Curated' should instead be chances to an 'Edit Layout' or 'Edit Sections' button. It isn't an important action so should be separated. Maybe it opens a bottom sheet with buttons and drop down boxes, I know the iOS Files app does that for its "View Options" button that contains settings. Or it could be an ["Edit Sections" like in Apple Music](https://github.com/user-attachments/assets/dbe5418a-113a-42c5-b116-d120ba0a8784).
11. The sort button uses a filter icon, which is fine I guess. Apple used to use it in Apple Music and Books to house filters _and_ sort options but they've since moved [sort options into the menu button](https://github.com/user-attachments/assets/f030a3a3-ea01-4508-92ba-3e2cd435e0ab). Though the Photo's app still uses it for sorting since there's no menu button. I don't think there's a good replacement icon either.
12. Related to (11) I always see the [filter button next to the menu button](https://github.com/user-attachments/assets/f4e5e1b7-6926-4824-965f-72c3f89ef804). I'm still thinking about whether this should be used since there will just be a blank space above the title. Side comment: what's annoying is that we can't get the [large header titles in line with the buttons like this](https://github.com/user-attachments/assets/0ff36796-fa55-4168-b7b0-0036f2c4b31f). Apple seems to do that whenever there's no back button. I think that would require custom headers for those screens.
13. In the settings page, the name 'Curated Downloads' sounds like automatic smart downloads, which could be a cool feature but it's not that feature. Especially since it's a toggle. If settings were segmented into submenus like "Appearance" for theme, accent, thumbnail ratio, etc, and then under "Downloads" the name could be "Show Curated Section", but maybe "Show Curated Downloads" could be fine for now.
14. I'm not even sure the 'Curated Downloads' setting should be shown in the Settings page. A lot of the settings pages in other self hosted apps feel like a flood of information and toggles that I have no idea of their use, and would be better served nearby what they change for added context and separation. For this case, like what described in (10). 

## The Downloads Select Screen
Here is the [Apple Books Edit Menu](https://github.com/user-attachments/assets/5e14c659-965e-4ebd-baa6-a3be283cc514), I reference it multiple times.
### Things I'm more sure about
15. The top left icon that means select all is super ambiguous. You shouldn't use ambiguous icons, I heard that in an Apple Developer video somewhere (also it's the same icon as add to Photo Album). Apple books just says "Select All" and "Deselect All" and that's perfectly fine.
16. The "Done" button should replace the top right menu button, not just be alongside a new menu button with completely new actions in it. https://developer.apple.com/design/human-interface-guidelines/toolbars#Actions says the button should be separate + tinted.
17. So with that change either the delete button and share are like in the Apple Books Edit Menu, or could just be a `ellipsis` button in the bottom right, or if there are more actions to add, you could look at the iOS files app which has a toolbar with multiple actions including delete and share.
### Things I'm less sure about
18. The selected item animation is nice, I'm still thinking about the actual selected item look. I only have one suggestion right now (19). Also [expo-symbols](https://docs.expo.dev/versions/latest/sdk/symbols/) supports animating the SF symbols so could be something to look into.
19. The space on the right between the border and the progress bar + cloud looks a bit odd. I think {marginHorizontal: -6, marginVertical: -4} plus shifting the cloud and shortening the progress bar slightly left \*might\* look better + increased border radius, I think this would work well with (5) but I'd probably have to play with it a bit to see what looks better + make sure it looks good outside the edit menu.
20. I'm also thinking about the location of the tick, whether it should be somewhere in the thumbnail or centered or somewhere else, I'm not sure yet.

## The Book Screen (Server)
21. I'm thinking about the download button but not sure about anything yet.
22. It might be good to use the local file if it's already downloaded to save mobile data for example.


Hopefully I didn't mistype any numbers. I might have forgotten some things too. I'm sure I'll notice again eventually if so.

aaronleopold:
I really appreciate the thoughtful feedback! I'll try to address everything but please call me out if I miss anything:

> I actually have a feature request for a Reading Timeline / Journal

:eyes: I'm intrigued

> The feedback is an extremely ios point of view, so I might say things impractical for android or need adjusting

That's totally fine, expo is making things easier to adapt across platforms anyway, e.g. like what I've been doing with the `.ios.tsx` variants for some components.

> In the sort menu, it should be much clearer what the sorting option actually does... if you look at [the menu button in Apple Music](https://github.com/user-attachments/assets/254f820f-cc90-4a34-b42e-5cdacb039876), the subtitle indicates the sort direction, and tapping flips the sort direction

I am on board with this change, there are a few things I would want to consider though:

- Tapping most options will close the menu. So if we merge ordering with the field that means if someone wanted to change both field and direction they'd have to open the menu twice -> `tap Series > reopen menu > tap Series again to change direction`
- I've started using `@expo/ui` for some dropdowns to slowly phase out `zeego` but it is missing some features. E.g. I still haven't figured out how to get subtitles, and the checkbox item doesn't like system icons for some reason last I checked. So I'd either need to refactor to use `zeego` for this menu or determine if we can stay with `@expo/ui` for it

> Small side rant about "ascending" and "descending" and other things...

That's totally fair, I'm very happy moving towards more semantic descriptions

> Another side comment but it's also why I have issues with the [LTR-RTL icon used in the reader menu](https://github.com/user-attachments/assets/efabcc01-8f61-4c9f-afcc-df72a76c7679)... I wish it were an open book with a folded left or right corner as the icon

Same, I remember I spent at least an hour on that one when I first added it but just didn't like any of the sf symbols for it lol

> 6. The page numbers + size should be sitting directly above the progress bar and not hanging directly below the title. Though for items without progress bars, I guess they'd need one.

This is why they landed where they are, mostly to avoid inconsistency between items with and without progress

> 8. Would rounding 1dp for the size be fine? I mean like 260.1MB instead of 260.11MB. I know epubs are gonna be small though.

Yeah I think that's totally fine, it could even be dynamic based on size (e.g. <1MB = 2dp, else 1dp)

> 9. Related to (5) I think there's currently like 5 different thumbnail widths in use

Good point, I definitely think there is a benefit to standardizing these more for consistency throughout the app. I'll try and make a note in the project board

> 10. I think 'Hide Curated' should instead be chances to an 'Edit Layout' or 'Edit Sections' button... Or it could be an ["Edit Sections" like in Apple Music](https://github.com/user-attachments/assets/dbe5418a-113a-42c5-b116-d120ba0a8784)

Yeah that's valid, I think since there is really only one section (two-ish, I guess) right now I just made it a button. I couldn't think of any other curated things to add, you could hypothetically create a port of the on deck section but that was all I could think of. But I do like the idea of an edit sections menu as we add more things

> 11. The sort button uses a filter icon, which is fine I guess. Apple used to use it in Apple Music and Books to house filters _and_ sort options but they've since moved [sort options into the menu button](https://github.com/user-attachments/assets/f030a3a3-ea01-4508-92ba-3e2cd435e0ab)... I don't think there's a good replacement icon either.

I agree, I don't think there is a great alternative icon for this, and I actually don't personally like that unified menu that your shared Apple moved to.

> 12. Related to (11)... what's annoying is that we can't get the large header titles in line with the buttons like this.

Funny enough, that is the direction I first tried to take. I am waiting on some features to be [merged/released](https://github.com/software-mansion/react-native-screens/issues/2990#issuecomment-3448692775). Honestly though I haven't checked back in since I subscribed to the thread, so maybe it's possible now, but I also didn't check whether the header alignment would also be made possible by that work (probably not).

> 13. In the settings page, the name 'Curated Downloads' sounds like automatic smart downloads, which could be a cool feature but it's not that feature... 14. I'm not even sure the 'Curated Downloads' setting should be shown in the Settings page. A lot of the settings pages in other self hosted apps feel like a flood of information and toggles that I have no idea of their use, and would be better served nearby what they change for added context and separation.

I think I agree! I'd say it can just be removed from settings and the toggle (or future changes to make it _not_ a toggle) can just be colocated with the actual content

> Here is the [Apple Books Edit Menu](https://github.com/user-attachments/assets/5e14c659-965e-4ebd-baa6-a3be283cc514), I reference it multiple times

So I actually tried to get somewhere in this direction before landing on where we are, and while I see you touched on some things I'll mention below I'll note a few bumps that steered the design:

- I couldn't find any reference to replacing the native tabs for an accessory like that, but all the other apps with some kind of selection state I used for inspo did just that
- Awkward to make it work for Android, too
- The `Select All` text kept getting clipped when I would try to add it to the `leftHeader` slot
- The aforementioned [unreleased features](<(https://github.com/software-mansion/react-native-screens/issues/2990#issuecomment-3448692775)>) that would allow me to make them separate buttons

All this to say, I totally agree with 15-17 and would love to get there eventually

> 18. ...Also [expo-symbols](https://docs.expo.dev/versions/latest/sdk/symbols/) supports animating the SF symbols so could be something to look into

:eyes: I haven't seen this one before, or just don't remember if I have lol but yeah I like the idea of animating the checkmark

> 19. The space on the right between the border and the progress bar + cloud looks a bit odd

Yeah I see what you mean, I'm happy to iterate on this one until it looks a bit better for sure. Once I have a moment I can try out your suggested changes

> 20. I'm also thinking about the location of the tick, whether it should be somewhere in the thumbnail or centered or somewhere else, I'm not sure yet.

I did try left-aligning it but hate the content shift when selecting/deselecting items, so landed on center. I think thumbnail placement could also work, too, though

> 22. It might be good to use the local file if it's already downloaded to save mobile data for example.

This is how PDFs and EPUBs will work (for PDFs if you tlgge the native reader). I think it's a bit more complicated for the comics, I would have to think on how the best way to approach that is so that you can benefit from immediate progress tracking while not requestign images from the server, or if it would even be worth the effort

---

A few things I didn't respond to because I just changed/fixed:

3, 4, 5, 14 (although I did respond)

A few things I didn't respond to because I just fully agree and didn't have anything overly meaningful to add:

7, 2

---

~Also my focus shifted a bit to trying to figure out why the `eas` build keeps failing for iOS. It does not like the `ZIPFoundation` dependency~ It isn't a nice fix but it should be fixed now. I have a PR up that hopefully the Readium fork of the library accepts that should resolve it without the plugin I added

Arklaum:
> 👀 I'm intrigued

I'll add it as an issue at some point soon so I can get opinions.

> [downloads sort menu] Tapping most options will close the menu. So if we merge ordering with the field that means if someone wanted to change both field and direction they'd have to open the menu twice

But right now changing both field and direction requires opening twice? I believe this this new case would reduce the number of taps needed (on avg) since the most common changes will be one opening, where right now it is two. So for this new menu it would depend which field you want, since the most "sensible" sort option is always the first tap, and it's inverse is the second tap. Speaking just for me, I know this by muscle memory already from many other apps: tap once for the sensible default (which is always `'A-Z'`or `'Newest First'`), open the menu and tap again to invert.

Again, just speaking for me, I think of `'A-Z'` and `'Newest First'` being a pair, because they're what you'd generally use for "Name" and "Date Downloaded". But in fact they're opposite "directions" which is hard for me to parse. So changing from "Name A-Z" to "Date Downloaded Newest First" would be one tap (which in fact changes both field and direction at once) compared to right now where it is two taps.


> I've started using @expo/ui for some dropdowns to slowly phase out zeego but it is missing some features. E.g. I still haven't figured out how to get subtitles

This one is still zeego, but yeah I noticed "Go to Series" and "Go to Library" look super weird because the titles aren't as subtitles right now. Hopefully these features are added soon. I also noticed the icons are on the left compared earlier iOS versions, which makes the `arrow.up.right` icon look a bit worse for those two buttons. oh well.

> I think since there is really only one section (two-ish, I guess) right now I just made it a button... But I do like the idea of an edit sections menu as we add more things

I think once there is customisation in the Home page, you'd just use the same kind of "Edit Sections" design even if downloads only ever has one section, just for consistency.

> I totally agree with 15-17 and would love to get there eventually

Yay glad to know that

> I did try left-aligning it but hate the content shift when selecting/deselecting items, so landed on center. I think thumbnail placement could also work, too, though

[Apple Books does this](https://github.com/user-attachments/assets/04eb0f3a-21c9-4c73-a56d-c251d5034f56) for lists if you like the left side tick, so no content shift? But I was more imagining something _on_ the thumbnail in the top left, or centered might be nice too, but I'd have to check if those really do look good or not. So one such idea is: we could have some existing circle in the top left of the thumbnail and the animation can firstly fill the circle and then form the tick.

aaronleopold:
> But right now changing both field and direction requires opening twice?

Hahaha yeah you're totally right, I don't know what I was thinking then. Also totally right re:

> This one is still zeego

I think I was conflating it with the action menu component I was working on earlier.

> So for this new menu it would depend which field you want, since the most "sensible" sort option is always the first tap, and it's inverse is the second tap

Okay yeah I understand what you're saying, I agree

*** 

> I think once there is customisation in the Home page, you'd just use the same kind of "Edit Sections" design even if downloads only ever has one section

Yeah nice I agree

> [Apple Books does this](https://github.com/user-attachments/assets/04eb0f3a-21c9-4c73-a56d-c251d5034f56) for lists if you like the left side tick, so no content shift?

My beef with that kinda layout is to avoid having the content shift dynamically you just permantly reserve space on the left for the checkbox, which I wouldn't like personally

> So one such idea is: we could have some existing circle in the top left of the thumbnail and the animation can firstly fill the circle and then form the tick

Yeah I like that one too

Arklaum:
I like the "A → Z" with the arrow that's super neat! Also the select overlay needs updating for the edge to edge.

Also also you should add `maintainVisibleContentPosition={{ disabled: true }}` to the `FlashList` to stop it from jumping around when updating the sort, because it's trying to follow a specific item. And for the `FlashList` maybe a reorder animation like the one at the end of this `FlatList` example video: https://docs.swmansion.com/react-native-reanimated/docs/layout-animations/list-layout-animations/ would be nice.

One more thing I noticed: when you select everything manually, the "Select All" icon doesn't update to the "Deselect All" icon. And in other apps if everything is selected and you untick one thing, then "Deselect All" becomes "Select All".

---

How ready is the new PDF reader? I've experienced a few issues (like RTL and Paged not working) so I don't want to highlight things before they are implemented. And I was wondering why not for example have used something like this https://github.com/wonday/react-native-pdf. I'm just trying to learn what comes with the current implementation and what I'd generally have to do to modify it.

And unrelated but I accidentally turned the brightness all the way down in the epub reader and got stuck with a black screen, and using Literata didn't set the font on the book itself, only the preview. And it would be nice if the status bar re-showed itself when the overlay is opened.

aaronleopold:
> Also the select overlay needs updating for the edge to edge... Also also you should add `maintainVisibleContentPosition={{ disabled: true }}` to the `FlashList` to stop it from jumping around when updating the sort

Thank you!!

>  And for the `FlashList` maybe a reorder animation like the one at the end of this `FlatList` example video: https://docs.swmansion.com/react-native-reanimated/docs/layout-animations/list-layout-animations/ would be nice.

Oooh yeah I totally agree, I can try and have a go at it once I have a moment this week

> One more thing I noticed: when you select everything manually, the "Select All" icon doesn't update to the "Deselect All" icon. And in other apps if everything is selected and you untick one thing, then "Deselect All" becomes "Select All".

Yeah the implementation isn't quite there yet. I'm still working through the kinks, basically I started drafting the solution to be more general since I know there are other areas of the app I see selection state management is needed, and the selected all state is a little off. I think I just need to have the store intake the count of the full set

> How ready is the new PDF reader? I've experienced a few issues (like RTL and Paged not working) so I don't want to highlight things before they are implemented.

It is definitely not ready for prime time, I also noticed some of the settings aren't correctly mapping to the native modules

> I was wondering why not for example have used something like this https://github.com/wonday/react-native-pdf.

Yeah good question! I liked being able to use the same locator format to support things like bookmarks and annotations (even though the backend currently still relies on `epubcfi` that will change). Other than that, part of me was just testing the waters around to see how viable Readium is for PDF in addition to ebooks.

If it came down to it, I'd happily ditch my custom native view implementation provided the alternative improves general usability or experience. One less thing to maintain 🙂 

> I'm just trying to learn what comes with the current implementation and what I'd generally have to do to modify it

I've mostly just poured through the toolkit docs for each language ([Kotlin](https://github.com/readium/kotlin-toolkit/tree/develop/docs), [Swift](https://github.com/readium/swift-toolkit/tree/develop/docs)) and used the `TestApp` in each repo as a frame of reference.

> And unrelated but I accidentally turned the brightness all the way down in the epub reader and got stuck with a black screen

Noted, we can probably set a lower limit on that setting for sure. I don't think we need to support 0% brightness lol

> using Literata didn't set the font on the book itself, only the preview

Hm, I maybe declared the font [incorrectly here](https://github.com/stumpapp/stump/blob/expo/downloads-and-sqlite/apps/expo/modules/readium/ios/EPUBView.swift#L256-L269) but am not sure

> And it would be nice if the status bar re-showed itself when the overlay is opened.

Added to project board!

aaronleopold:
@Arklaum I saw you had another PR (you are awesome) so I wanted to give you a heads up that I plan on merging this in later today as-is, although I know there are still rough edges to be ironed out. Sorry ahead for the conflicts that likely will crop up when I squash this in!

PR #812, opened 2025-10-30
⚡️/✨ (expo): Flashlist the reader, vertical support, and new performance monitor

Arklaum:
So this includes:
1. FlashLists for the reader and footer
2. Vertical reader support
3. A new performance monitor with RAM and CPU usage

[An image of (2) and (3)](https://github.com/user-attachments/assets/cbd22239-583d-45cb-96a1-b59f660674b1)

For the FlashLists, I got out my old iphone 6s to try some release builds on and it was definitely better. On my 900MB 400 page manga, compared to FlatList:
- there was almost no stuttering when swiping fairly quickly where FlatList had so much more
- didn't crash where FlatList sometimes did (no idea why)
- and the footer items loaded in faster when scrolling fast
- \*Maybe\* a marginally lower CPU usage? but definitely a steadier UI FPS

For the new performance monitor, I used https://github.com/hannojg/react-native-performance-stats. It's unmaintained but small. I had to patch it to work again on ios + added a bug fix for fast refresh on ios from one of it's PRs + modified the android cpu usage to work in release builds. RAM is probably the most useful part though.

I swapped the drop down menu to use 'Paged' then 'Scroll (Horizontal)' then 'Scroll (Vertical)' instead of 'Paged' then 'Scroll (Vertical)' then 'Scroll (Horizontal)'  since that's how it was in `ReadingModeSelect.tsx` but I have two competing thoughts which are: "it should be sorted by frequency of use" and "it should be sorted by similarity". I don't really have a preference so I can change it in `ReadingModeSelect.tsx` that's preferable.

---

Some memory usage details: In a development build, the new perf monitor seemed to cause memory to go up by 40MB per hour (previous one was 2.5GB per hour, which is a bit suspicious since 40MB * 60 = 2400MB and the prev one updated at the refresh rate whereas the new one is every second). But I also tested a release build and didn't see any memory issues there so I was curious and:

- I added it to a pre reading timer memory leak fix and saw that one existed but was slower. It seemed to cause memory to increase by a few MB, stop for a a bit, then repeat.

- I did the same for the previous performance monitor issue, and it was also slower but increasing continuously.

- I thought maybe this new one might just be slower too, so I left it for 12 hours and it went from 190MB to 207MB usage so ~1.5MB per hour. And on android memory seems to sawtooth (go up over a few minutes by about 10MB then jump back down) but over 8 hours measured after the jumps down it was ~1MB per hour.

I probably should be using external tools but that's no fun

---

I don't know if it's relevant anymore since it's an old ios version but there were two headers stacked on top of each other on my iphone 6s (e.g. two back buttons, one the native ios 15 back button and the other the `<ChevronBackLink />` just below it).

I also had to delete all downloads and redownload after rebasing onto the new branch.

### DIFF
--- a/apps/expo/components/book/reader/image/Footer.tsx
+++ b/apps/expo/components/book/reader/image/Footer.tsx
@@ -45,7 +45,13 @@ export default function Footer() {
 	} = useImageBasedReader()
 	const elapsedSeconds = useBookReadTime(book.id)
 	const {
-		preferences: { footerControls = 'slider', trackElapsedTime, readingDirection, readingMode },
+		preferences: {
+			footerControls = 'slider',
+			trackElapsedTime,
+			readingDirection,
+			readingMode,
+			doublePageBehavior,
+		},
 	} = useBookPreferences({ book, serverId })
 
 	const galleryRef = useRef<FlashListRef<number[]>>(null)
@@ -139,10 +145,11 @@ export default function Footer() {
 	)
 
 	const visibilityChanged = usePrevious(visible) !== visible
+	const doublePageBehaviorChanged = usePrevious(doublePageBehavior) !== doublePageBehavior
 	useEffect(() => {
 		if (footerControls !== 'images') return
 
-		if (visible && visibilityChanged) {
+		if (visible && (visibilityChanged || doublePageBehaviorChanged)) {
 			const idx = pageSets.findIndex((set) => set.includes(currentPage - 1))
 			if (idx === -1) return
 			galleryRef.current?.scrollToIndex({
@@ -152,7 +159,7 @@ export default function Footer() {
 				viewOffset: -3, // account for half of the gap between two adjacent page sets
 			})
 		}
-	}, [footerControls, currentPage, visible, visibilityChanged, pageSets])
+	}, [footerControls, currentPage, visible, visibilityChanged, pageSets, doublePageBehaviorChanged])
 
 	const formatDuration = useCallback(() => {
 		if (elapsedSeconds <= 60) {


aaronleopold:
I'm excited to take a look at this! I'll try to look after work today, but it might not be until the weekend. I wanted to respond to some things though:

> I also had to delete all downloads and redownload after rebasing onto the new branch.

You'll have to delete downloads for each development build you make, since the container ID changes which affects things like `FileSystem.documentDirectory`. It shouldn't be an issue outside development builds, though, I think the container ID stabilizes. I can't say for sure it would be the same between a TestFlight and App Store version, though.

> I got out my old iphone 6s to try some release builds on

I love this, and:

> I don't know if it's relevant anymore since it's an old ios version but there were two headers stacked on top of each other on my iphone 6s

I noticed something similar when I spun up an older version of iOS in a sim to investigate a bug someone reported. I noted it in the mobile project but don't know how many people are still using iOS 15 to gauge priority 

> It's unmaintained but small. I had to patch it to work again on ios + added a bug fix for fast refresh on ios from one of it's PRs + modified the android cpu usage to work in release builds. RAM is probably the most useful part though

It might be worth forking it to this org and applying your fixes there, but patch works perfectly for now

> I have two competing thoughts which are: "it should be sorted by frequency of use" and "it should be sorted by similarity". I don't really have a preference so I can change it in ReadingModeSelect.tsx that's preferable.

I think the first is subjective, everyone will have different frequencies of use (e.g., I almost never use anything besides ltr paged). fwiw I think the order you changed it to is perfect for my brain, and it is following alphabetical ordering

Arklaum:
> I think the first is subjective, everyone will have different frequencies of use

Individually that's true but I meant like in aggregate, I can fairly confidently say the frequency of use will be 'Paged' then 'Scroll (Vertical)' then 'Scroll (Horizontal)'. I know for example Apple's guidelines say place frequently used items first, but also group similar items next together. I changed it to that order because 'Paged' and 'Scroll (Horizontal)' group well together (both horizontal), and the order is semi-sorted by frequency, which is fine. 

Minor things but I like obsessing over them! By "not having a preference" I actually meant I didn't mind changing it, not that I didn't have an order I most liked. This stuff about order is something I will bring up again later about the downloads section but way more important there, and with some examples too.

aaronleopold:
> I know for example Apple's guidelines say place frequently used items first

TIL, I should really give HIG a proper read someday

> I changed it to that order because 'Paged' and 'Scroll (Horizontal)' group well together (both horizontal), and the order is semi-sorted by frequency, which is fine

I think this makes perfect sense, I'm personally happy with it staying as you have it here

> Minor things but I like obsessing over them!

I feel similarly and I appreciate meeting others wired in that way! I can also be, separately, massively indecisive with some things so when stuck on a design/UX problem I will happily lean on others I'd trust in that area with stronger opinions. Otherwise, I break out [the wheel](https://spinthewheel.io/)

> This stuff about order is something I will bring up again later about the downloads section but way more important there

Nice, yeah like I said in another thread somewhere I'd love alllll the feedback I can get

PR #822, opened 2025-11-14
:zap: WIP SQL pool monitoring and resource guarding

aaronleopold:
This PR drafts some changes I was trying out to better protect against resource exhaustion after some discussions on Discord. The changes started to get a little stale, so I am making a PR to remind myself to see them through.


aaronleopold:
Closing this, I no longer feel this is the correct approach. A few reasons:

- While investigating some of the ingestion issues (PIDs, time, etc) I investigated more how sqlx manages connection pools for sqlite and this would kinda just be duplicating what it already does internally and therefore be a useless abstraction
- My next thought was to perhaps just give the server its own pool, rather than sharing it with core processing, to isolate held connections. I think that is viable, but stopping exploration for the next point
- Someone with a very sizable library has been helping test the current unstable and found that adjusting the busy timeout to be 30s resolved those pesky acquire connection errors. I don't think it literally fixes it, just extends the timeout, but is an easier correction to make for now

PR #839, opened 2025-12-04
♻️ (expo): New Perf Monitor Library

Arklaum:
Change to use a new library `react-native-performance-toolkit`.

I couldn't do `yarn add` for some reason, I had to manually type into `package.json`. If I were to delete the lock file then `yarn add` does work again, but I didn't want to do that without your input.


Arklaum:
I get error `Invariant Violation: expected workspace package to exist for "send"`.

aaronleopold:
> I get error `Invariant Violation: expected workspace package to exist for "send"`.

Yeah okay, I see exactly the same. A fresh lockfile also fixed it for me, too. I'll inspect the lockfile tomorrow to see if there is something yarn doesn't like but honestly just generating a fresh one is probably fine. I'll base it to the unstable branch if so, just in case

PR #860, opened 2025-12-22
:bug: Fix OPDS pagination infinite loop caused by `next_page()` bug

asonas:
Fix a bug in `OffsetPagination::next_page()` where the method returns the current page instead of the next page when using 1-based pagination (the default). This caused OPDS 2.0 clients to enter an infinite loop when following pagination links.

## Problem

When `zero_based` is `false` (the default, 1-based pagination), the `next_page()` method returned `self.page` instead of `self.page + 1`. This resulted in the OPDS 2.0 `next` link always pointing to the current page (e.g., `?page=1` -> `?page=1`).

## Solution

Simplified the method to always return `self.page + 1`, as the next page calculation does not depend on whether pagination is 0-based or 1-based.

## Changes

- Fixed `OffsetPagination::next_page()` to correctly return the next page number
- Added unit tests for `next_page()` covering both 0-based and 1-based pagination scenarios

### DIFF
--- a/crates/graphql/src/pagination.rs
+++ b/crates/graphql/src/pagination.rs
@@ -84,11 +84,7 @@ impl OffsetPagination {
 	}
 
 	pub fn next_page(&self) -> u64 {
-		if self.zero_based.unwrap_or(false) {
-			self.page + 1
-		} else {
-			self.page
-		}
+		self.page + 1
 	}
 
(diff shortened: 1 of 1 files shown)


aaronleopold (approved):
This looks great, thank you for also adding tests!

PR #862, opened 2025-12-23
Add a secure library feature that allows user to create "secure" library for sensitive media content

oovz:
Secure Libraries provide client-side encryption for sensitive content. All media files, thumbnails, and catalog metadata are encrypted at rest using AES-256-GCM. The server **never** stores decryption keys — content is decrypted only in the browser after the user unlocks it.

### What Secure Libraries Protect Against

- **Server compromise**: Even with full database and filesystem access, an attacker cannot read encrypted content without user passwords
- **Unauthorized access**: Users without explicit access grants cannot view or even detect the existence of secure library content
- **Data at rest exposure**: All media files are encrypted on disk; original plaintext is deleted after encryption

### What Secure Libraries Do NOT Protect Against

- **Client-side attacks**: If an attacker compromises a user's browser session while unlocked, they can access decrypted content
- **Weak passwords**: Encryption keys are derived from user passwords; weak passwords weaken protection
- **Screen capture/recording**: Decrypted content is visible in the browser and can be captured

### Key Concepts

| Term | Description |
|------|-------------|
| **SMK** (System Master Key) | Generated once during system setup; required to create secure libraries and grant access. Store securely — it cannot be recovered. |
| **LMK** (Library Master Key) | Derived from SMK for each library; used to encrypt content. Never stored on server. |
| **Keypair** | X25519 key pair generated for each user at first login; used to wrap LMK for per-user access grants. |

### High-Level Flow

1. **Server owner** runs `stump_server system setup` to generate and save the SMK
2. **Create secure library** via UI or API, providing the SMK
3. **Scan** the library to encrypt all content (original files are deleted)
4. **Grant access** to users (each grant wraps the LMK with the user's public key)
5. **Users unlock** by entering their password, which decrypts their private key and unwraps the LMK
6. **Browse and read** normally — decryption happens transparently in the browser

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ Stump is a free and open source comics, manga and digital book server with OPDS 
 
 - [Roadmap 🗺](#roadmap-)
 - [Getting Started 🚀](#getting-started-)
+- [Secure Libraries 🔐](#secure-libraries-)
 - [Developer Guide 💻](#developer-guide-)
   - [Where to start?](#where-to-start)
 - [Project Structure 📦](#project-structure-)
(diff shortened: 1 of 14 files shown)


oovz:
implementation is done to web client and it works with this new secure library feature. desktop/mobile clients are not changed.
OPDS support may never work because of the nature of client-side decryption at-rest.

aaronleopold:
Thank you for the effort here, but I don't feel that this feature aligns with Stump at the moment. The vast majority of people don't need client-side encryption for their digital book libraries, and IMO the added complexity (e.g., unrecoverable SMK, permanent file deletion) creates more risk than benefit for people using Stump. It also adds significant complexity to the codebase that I don't necessarily want to commit to maintaining right now, at least not unless I see a very strong demand for it. I'm curious if there a reason disk encryption at the OS level doesn't work for you?

Also, generally speaking, large features like this should be discussed in an issue/discussion first before being implemented based into `main` with 34k additions. Even if I wasn't in the middle of a very massive rewrite effort that would conflict heavily with these changes, it is a lot to take in all at once. I hate shooting down folks who put in the effort to contribute, which is why I outlined [guidelines for contributing](https://github.com/stumpapp/stump/blob/main/.github/CONTRIBUTING.md) to the repo to prevent this sort of thing.

I'm going to close this PR for now, but feel free to open an issue/discussion if you'd like to discuss further.

PR #867, opened 2025-12-27
WIP: :sparkles: (expo): Support bookmarking and theme customizations

aaronleopold:
This is a heavy WIP, we'll see if I can clean it up enough to get it in before the new year


Arklaum:
I had some small thoughts about the ebook reader. We could use [this](https://docs.expo.dev/versions/latest/sdk/brightness/) to set the brightness instead of it being fake? And the `minimumTrackTintColor` for the slider should probably be another colour so that you can actually see the brightness bar.

And with the new sheets, we could have like `detents={[0.5]}` and `dimmed={false}`, then the actual book's text itself seems like the perfect preview window?

aaronleopold:
@Arklaum I'm sorry, I totally missed this message in my inbox!

> We could use [this](https://docs.expo.dev/versions/latest/sdk/brightness/) to set the brightness instead of it being fake?

Yeah I'm not opposed to that, originally I didn't go that route (and instead only applied it to the readium webview) so the settings don't become too dim.

> And the `minimumTrackTintColor` for the slider should probably be another colour so that you can actually see the brightness bar

I'll add a couple new colors to the config for sliders

> And with the new sheets, we could have like `detents={[0.5]}` and `dimmed={false}`, then the actual book's text itself seems like the perfect preview window?

I'll give it a try and see how it feels

Arklaum:
> Yeah I'm not opposed to that, originally I didn't go that route (and instead only applied it to the readium webview) so the settings don't become too dim.

I'm personally used to the actual device dimming. Apple Books and Google Play Books do it that way, but the Kindle app doesn't. 

Stepper buttons that are separate like this look nice and also might be less likely to mess up pressing too:

<details>
<summary> Stepper buttons Before and After </summary>

<table>
<thead>
<tr>
<th align="center"><strong>Before</strong></th>
<th align="center"><strong>&nbsp;After&nbsp;&nbsp;</strong></th>
</tr>
</thead>
<tbody>
<tr>
<td align="center">
IMG_4173
</td>
<td align="center"> 
IMG_4172
</td>
</tr>
</tbody>
</table>

</details>

Would you like them?

And for the theme selector, if there were smaller previews you could have empty theme as a button for adding, and long press for a menu to delete or edit:

<details>
<summary> Example </summary>

IMG_4172

</details>

aaronleopold:
Please feel free to push to this PR as you'd like! I really appreciate you spinning this up locally and taking a look!

***

Wrt the dimming, I don't have a strong lean either way tbh. So, if you feel more strongly with it I am totally good with that change.

> Stepper buttons that are separate like this look nice and also might be less likely to mess up pressing too... Would you like them?

Yeah those look significantly better, eventually I was just going to use expo ui's stepper components (in main, not released last I checked) but those are great.

> And for the theme selector, if there were smaller previews you could have empty theme as a button for adding, and long press for a menu to delete or edit

I love interaction design you suggested here a LOT more wrt the dashed border one to add another. I think a long press menu generally works a lot better on iOS, and often behaves a bit wonkier on Android, but I'm very good to try it out

Arklaum:
So I added the changes! 

For the brightness slider, it works as expected on ios but I can't test android properly. On android studio, the system brightness slider seems to move in different proportions compared to when I move the theme sheet brightness slider, which is strange. I have no idea about real devices though.

aaronleopold:
Thank you!! I won't have time to really check them out until tomorrow or Thursday. For the brightness stuff on Android, I'll boot up my old Pixel 2 and see if it functions okay on it

aaronleopold:
I had a little time during lunch to check it out and think it's great! I fixed some funk I noticed on Android, mostly around the dropdown menus rendering above the sheet, and added the delete/duplicate options as suggested for that long press menu. 

I didn't have time to test on an actual Android device for the brightness, but couldn't actually get the iOS one to work in the sim (might be expected for sim, haven't checked docs yet I was pressed on time). I'll try again once I have a moment, though

Arklaum:
> added the delete/duplicate options as suggested for that long press menu

Oh I forgot to mention that in my previous message. I was gonna say these would be good and specifically duplicate so you could use a current option as a new starting point to make small modifications and compare what you like. Thank you! I did notice before we were unable to delete the default themes, but now they can be?

I had another idea: in the create theme sheet we could add like a repository of curated themes. A grid of the small square theme previews with some nice curated options. It could have the defaults (if they are left deletable), plus other curated niche options. 

> couldn't actually get the iOS one to work in the sim (might be expected for sim, haven't checked docs yet I was pressed on time)

That's probably the case since it does work on my phone.

aaronleopold:
> I did notice before we were unable to delete the default themes, but now they can be?

Yeah, while I was porting things over to the dropdown I realized that the delete protection is keyed off by name, which is a bit silly on my part since someone could have just created a custom `Dark` and be unable to delete it. There isn't a strong reason in my head for enforcing people to keep the default ones, as long as there is always at least one theme I figured that would be enough. If you disagree or think I'm missing something, though, let me know. We can also just add static IDs for the default ones and add the protection back

> I had another idea: in the create theme sheet we could add like a repository of curated themes

I love things like this! I'm not 100% sure where they could live, though. Could maybe just have a repository in the org for storing them? Unless you imagined it living in users' server instances, directly?

Arklaum:
Where they could live? I was thinking I'd just hand pick a dozen nice colour combinations, give them names, and they'd be an array or whatever in `CustomizeTheme.tsx`

aaronleopold:
Ha, yeah I just misunderstood how you meant it. When you said repository of themes I assumed a user-curated one, like a shared library of sorts. I'm totally good with:

> just hand pick a dozen nice colour combinations, give them names, and they'd be an array or whatever in CustomizeTheme.tsx

aaronleopold:
One thing I have been trying to think through wrt bookmarks (and honestly annotations, once I add them) is how to sync things properly between the server and mobile app. It feels a good enough amount more complex than normal read progression (especially since there is effectively a single, containing entity which houses active progress), and so I may wind up merging these features in before having that sync in place. I originally thought it would be easy enough to sync up those which don't have a `serverBookmarkId`, but if I were to just toggle a bookmark on/off while offline reading then it would sync back up and potentially duplicate. Comparing locators isn't 100% since different screen sizes can yield different objects. Mostly just sharing in case it is of interest

Arklaum:
I've added 16 premade themes! The themes in each row are loosely related. Three themes are the existing ones, but "Sepia" now refers to a different set of colours, and Sepia's old colours are now called "Parchment". Should the EpubThemesStore version be incremented so the name doesn't refer to the old colours anymore?

aaronleopold:
Awesome! Thank you! I'll pull it down shortly and give it a spin. 

> Should the EpubThemesStore version be incremented so the name doesn't refer to the old colours anymore?

Yeah I think that makes sense

aaronleopold:
I definitely don't think the bookmark stuff is perfect, but I think this is good enough for initial testing so I will merge this in shortly and try to push a build out sometime today/tomorrow. 

I didn't have time to install a dev build on a physical Android device, but since it is just brightness in question for that I think that is fine enough for now too

PR #870, opened 2025-12-29
:lipstick: Refresh web thumbnail and card components

aaronleopold:
This PR aims to bring some of the patterns used in the mobile app wrt thumbnails and thumbnail cards to the web app. It isn't 100% yet, but I think it is getting close


aaronleopold:
I'm sure I missed some spots and this isn't 100% perfect, but I think I'll call it here and merge this in shortly

PR #871, opened 2025-12-30
💄 (expo): Softer footer animation and option to disable dismiss gesture

Arklaum:
So here we have two things: 
1. The softer footer animation. Could also be tried on web too.

2. An option to disable the dismiss gesture in the reader. 
    - I put off the change since the settings sheet didn’t show with just the addition of `presentation: 'fullScreenModal'`, but luckily the new sheet library works fine with it out of the box so I changed it over.
    - It doesn't help on android either, and I can't find an easy way to disable swipe back. On android it doesn't seem to be as big of a problem because the first swipe that registers as a "dismiss gesture" just showed the system bars and only a second swipe while those are visible takes you back (on iOS the first gesture takes you back).

### DIFF
--- a/apps/expo/app/(tabs)/settings/index.tsx
+++ b/apps/expo/app/(tabs)/settings/index.tsx
@@ -1,4 +1,4 @@
-import { View } from 'react-native'
+import { Platform, View } from 'react-native'
 import { ScrollView } from 'react-native-gesture-handler'
 
 import { ContactInformation, SupportInformation } from '~/components/appSettings'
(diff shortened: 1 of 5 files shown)


aaronleopold (approved):
Thank you!

***

> Could also be tried on web too

Yeah, aligning makes sense

> On android it doesn't seem to be as big of a problem because the first swipe that registers as a "dismiss gesture" just showed the system bars and only a second swipe while those are visible takes you back

Yeah I think that is acceptable for now

Issue #757, opened 2025-09-07
[BUG] Condition where library doesn't enter setup config


Arklaum:
## Description of the Bug

If the cursor is focused on the path or name text box (i.e. you clicked on it) when creating a library and you hit return, it immediately creates the library with no other setup options being presented.

I'll probably look into fixing it later after I finish the save issue PR

PR #873, opened 2026-01-01
🐛 Fix premature library creation

Arklaum:
Solves #757. Would you prefer this in nightly?

I thought at first it was because of this:

https://github.com/stumpapp/stump/blob/ec3447292ee9f5cd99f925c7b4d941f16291c04a/packages/browser/src/scenes/createLibrary/CreateLibraryForm.tsx#L171-L172

because making it conditional fixed it for the first step (i.e. for the Name and Path inputs) but didn't fix it for the other steps (e.g. the Ignore rules input box and the thumbnail generation scale factor input box, etc).

Also I didn't see the issue in the note but it doesn't really matter

### DIFF
--- a/packages/browser/src/scenes/createLibrary/CreateLibraryForm.tsx
+++ b/packages/browser/src/scenes/createLibrary/CreateLibraryForm.tsx
@@ -159,6 +159,15 @@ export default function CreateLibraryForm({ existingLibraries, onSubmit, isLoadi
 		}
 	}
 
+	/**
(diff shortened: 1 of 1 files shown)


aaronleopold (approved):
Thank you!

> Would you prefer this in nightly?

Yeah for a small fixes like this we can just base into `nightly`, but I'm lazy right now so I'll bring it up with the other commits in the next day or so

> Also I didn't see the issue in the note but it doesn't really matter

It's possible it was resolved through some update over time, it was 2 years ago so

PR #876, opened 2026-01-05
🐛 (expo): Fix empty list message

Arklaum:
I missed some bits in my last minute renaming of `ListEmptyMessage`, so I've fixed that. And a fix for if you load the app without stump features enabled then enable it the server card will have tiny height.

---

Also, do you think the settings should look more in line with the other lists [like this](https://github.com/user-attachments/assets/5c93672f-299b-4cec-9747-c8c9fd702885)?

### DIFF
--- a/apps/expo/app/(tabs)/index/index.tsx
+++ b/apps/expo/app/(tabs)/index/index.tsx
@@ -136,7 +136,7 @@ export default function Screen() {
 
 			{!isCleanSlate && (
 				<ScrollView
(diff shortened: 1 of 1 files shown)


aaronleopold (approved):
This looks great, thank you!

> Also, do you think the settings should look more in line with the other lists [like this](https://github.com/user-attachments/assets/5c93672f-299b-4cec-9747-c8c9fd702885)?

Yeah, I think it makes sense to unify the patterns throughout

PR #890, opened 2026-01-11
🐛 (expo): Add disable dismiss gesture option to downloads

Arklaum:
I missed this bit and was so annoyed when it happened


aaronleopold (approved):
Good catch! If you can merge (think you have permissions) feel free, otherwise I'll be by a computer later this evening. I'll try and push a tiny build tomorrow so you can have it

PR #892, opened 2026-01-11
🐛 Fix stale form ui

### LINKED ISSUE #891: [BUG] Unable to Change User Permissions
## Description of the Bug

I would like to allow users to download books, but the check box in the user settings page does not respond to checking the box - remains empty.

## Expected Behavior

I am able to change user permissions.

## To Reproduce

1. Newly add user, not default admin.
2. Go to user settings and try to change their permissions
3. observe

## Screenshots

I can, though it would be an empty check box.

## Debugging Information

All two users are OIDC auth'd.  Admin and user created by selecting Log In with OIDC.  Neither have stump passwords.  In the `unstable` build for OIDC, I was able to adjust permissions.  This has been my first attempt for `experimental`.

**Environment:**

- OS: OMV 7, running Docker
- Device: MacOS / Safari, iOS Safari - both do not seem to be able to select the box.

**Build Details:**

- Version: Semantic version
[v0.0.12](https://github.com/stumpapp/stump/releases/tag/v0.0.12)
Build channel
Experimental
Exact commit
[ab8a887](https://github.com/stumpapp/stump/commit/ab8a887)
Build date
January 10, 2026 7:08 PM
- Docker: Yes

**Additional context**

I was lead to this as I was trying to get the new user to be able to use the iOS app.  When trying to load a book in the app, the book does not seem to load (Maybe an entirely different issue) but I recall reading something along the lines that the book cannot stream, instead needs to download - I was going to see if allowing downloads resolved the iOS issue for the user.

No issues with admin user.

Arklaum:
This should resolve #891

The places where the UI was stale were where `useFormContext` and `form.watch` were both used since React Compiler doesn't see `form` from `useFormContext` as something to skip auto-memoisation for. The other places that had `useForm`should now be auto-memoised too. I found that [this extension](https://marketplace.visualstudio.com/items?itemName=blazejkustra.react-compiler-marker) was quite neat since it marks which components are auto-memoised.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -6,6 +6,7 @@
 		"Gruntfuggly.todo-tree",
 		"mike-co.import-sorter",
 		"aaron-bond.better-comments",
-		"ryanluker.vscode-coverage-gutters"
+		"ryanluker.vscode-coverage-gutters",
+		"blazejkustra.react-compiler-marker"
 	]
 }


aaronleopold:
Thank you! I just stepped away from my computer but will give this a review later today. I wanted to quickly respond to:

> I found that [this extension](https://marketplace.visualstudio.com/items?itemName=blazejkustra.react-compiler-marker) was quite neat since it marks which components are auto-memoised.

Feel free to add it to the suggested list of extensions in the `.vscode` folder, if you'd like

Issue #885, opened 2026-01-08
[BUG] Mobile - A collection of the android mobile application


skridofly:
# Status on 09.01.26
## Book Overview
- The Chapter name in the Book overview still sometimes displays the path to the current chapterHTML (e.g. text/part0006.html). Sometimes properly displays the real chapter name
- The extended Description is consistently cut off early
- "Ackerman & Shirazi" gets turned into "Ackerman &amp; Shirazi" solely in the Series Position Indicator
## EPUB Reader
- Changing the Font Size, only reloads font size in preview section. The EPUB itself needs to be reentered
- The EPUB will never autocomplete, even after reaching the last page in the book
- Unable to jump to Chapters in the Contents Menu
    - Progress isn't saved after unsuccessful Chapter jump
- After tapping somewhere near the menu bar on top, the reader will be jumped to position chapter one (happened once so not fully sure)
- Sometimes after having the app open in the background for a long time, it changes maybe the font, so the reader position stays on the same page, but the content is shrunk and as such on a different position
## Downloads
- The Chapter list for downloaded EPUBs is empty
- The downloaded EPUBs will only sync the progress from server to download after having the downloaded opened ones
- Downloading EPUBs from the same server of two entries will overwrite each other, instead of creating two downloaded EPUBs
## Performance bugs/Random behavior
- Slow loading times when entering specific paths quickly after start up
- Loop of the back gesture. I.e. the back gesture in Series View load Series View again
- First describe in [this](https://discord.com/channels/972593831172272148/1332439367016513656/1454954617267028176) discord message

aaronleopold:
> The extended Description is consistently cut off early

You should be able to long press to launch a little sheet that shows the full content. Is this not working? Or perhaps just not clear and therefore not ideal UX?

<details>

<summary>E.g. iOS</summary>

![image](https://github.com/user-attachments/assets/3e34f91f-3691-44b7-829d-ea6726f1f601)

</details>

skridofly:
> You should be able to long press to launch a little sheet that shows the full content. Is this not working? Or perhaps just not clear and therefore not ideal UX?

This is the extended version. The overview shows a cut off version as planned but when clicking on that one, the section that pops up is still cut by like 5 words or so

Edit: I just noticed an update has been pushed which fixes this behavior

aaronleopold:
> The Chapter list for downloaded EPUBs is empty
> Changing the Font Size, only reloads font size in preview section. The EPUB itself needs to be reentered

[...]

PR #893, opened 2026-01-13
🐛 (expo): fix default stump server entry

Arklaum:
Should fix issue 1 of #885. If you had opds disabled on a stump server it wouldn't auto enter on startup. 

I personally don't think if you have a stump server you need to be able to have it's opds version open as default, but I could add that selection option.


aaronleopold:
> I personally don't think if you have a stump server you need to be able to have it's opds version open as default, but I could add that selection option

I agree, ideally for a given default server it should be:

1. If stump server -> `/server/id`
2. If OPDS-only server -> `/opds/id`

PR #894, opened 2026-01-15
Mobile app fixes and updates

aaronleopold:
A collection of changes made while mostly working through #885, primarily bug fixes but I will have a change that improves the progression sync for downloads


aaronleopold:
I'll merge these in as-is and then just create more focused PRs for additional fixes

PR #897, opened 2026-01-18
🐛 (expo): fix premade themes grid on tablets

Arklaum:
The grid didn't look great on tablets (didn't have 4 columns). Now it should look nicer.


aaronleopold (approved):
Thank you!

PR #910, opened 2026-01-25
💄 (expo): Show status bar with controls

Arklaum:
Makes the status bar (time, battery , etc.) show when the controls overlays are shown. 

Not great on the epub reader yet, but I have already tested logic for dealing with the theme colours and commented it out. What would be the best way to check it's specifically the epub reader we are using? I could use the readium `locator` or `book` from the epub location store, and when it's undefined then we're not reading epub books? Or store something like `isReadingEbook` and have it set in the `read.tsx` files? 

Also would complete this: https://github.com/orgs/stumpapp/projects/8?pane=issue&itemId=136966004 (I can't move it myself)

### DIFF
--- a/apps/expo/app/_layout.tsx
+++ b/apps/expo/app/_layout.tsx
@@ -3,6 +3,7 @@ import '~/global.css'
 import { DarkTheme, DefaultTheme, Theme, ThemeProvider } from '@react-navigation/native'
 import { PortalHost } from '@rn-primitives/portal'
 import * as Sentry from '@sentry/react-native'
+import { getColor, to } from 'colorjs.io/fn'
 import dayjs from 'dayjs'
 import duration from 'dayjs/plugin/duration'
 import relativeTime from 'dayjs/plugin/relativeTime'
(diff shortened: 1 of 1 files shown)


aaronleopold:
> What would be the best way to check it's specifically the epub reader we are using? I could use the readium `locator` or book from the epub location store... Or store something like `isReadingEbook` and have it set in the `read.tsx` files?

I think either is viable imo, I'd say a dedicated flag is more readable but would be totally good with either

> Also would complete this: https://github.com/orgs/stumpapp/projects/8?pane=issue&itemId=136966004 (I can't move it myself)

Let me know if you still can't move it

Arklaum:
I went with this: 

https://github.com/stumpapp/stump/blob/34b813025875dad69e49e67c68c8984c9ab1fb42/apps/expo/app/_layout.tsx#L87

because it was the easiest. Using `state.locator` actually didn't work because exiting the reader doesn't make it go back to being `undefined`.

But properly setting and storing `isReadingEbook` should maybe probably be done like when the nested controls are added.

PR #914, opened 2026-01-28
:sparkles: (expo): Support imports and series downloading

### LINKED ISSUE #909: [FEATURE] a download series button in the mobile app
It’s rather time consuming to download large amounts of comics by hand.

I’d like for a download series button to be implemented in the same menu as the scan series and overview buttons.

aaronleopold:
Resolves #909

This PR is not ready, there is quite a bit of testing I want to do still (especially on Android). There are a few primary changes worth highlighting:

1. I've added a `+native-intent` file to route imports accordingly, which TLDR; enables you (in addition to the entitlements changes) to e.g. go to your file browser and open a supported book in the app. Doing so will import it into the local library
2. The `Downloads` tab has been renamed to `Library`/`Local Library` since it isn't strictly downloads anymore
3. All explicit download actions are now queued, which is fairly common I think. Implicit downloads (e.g., trying to read an ebook without a download present) skip the queue.
4. You may now download an entire series. I should note that you can always hit the download button (if you have the permissions), even if you already have the series downloaded. I didn't think querying for all IDs just to disable the button was worthwhile.

The queue button isn't perfect, but should be fine for now. I'll aim to get this in sometime over the weekend (ideally)

### DIFF
--- a/crates/graphql/schema.graphql
+++ b/crates/graphql/schema.graphql
@@ -2624,6 +2624,7 @@ input UpdateUserPreferencesInput {
 	enableAlphabetSelect: Boolean!
 }
 
+"A multipart file upload"
 scalar Upload
 
 input UploadBooksInput {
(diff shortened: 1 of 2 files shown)


aaronleopold:
I've done some testing on Android and am comfortable enough merging this in. I couldn't test the import functionalities directly on Android, though, so some bugs are definitely expected.

PR #921, opened 2026-01-31
💄 (expo):  Update table of contents

Arklaum:
Updates the table of contents design to look nicer. It also now has the synthetic page numbers for each chapter, and a liquid glass "scroll to current chapter" button that appears if the current chapter is out of view.

Also updates the header to grab the chapter name for chapters that are broken into multiple resources (such as those where images are supposed to be in the middle of a chapter)

### DIFF
--- a/apps/expo/components/book/reader/epub/LocationsSheetContent.tsx
+++ b/apps/expo/components/book/reader/epub/LocationsSheetContent.tsx
@@ -305,7 +305,7 @@ const TableOfContentsListItem = ({
 						>
 							<Text
 								className={cn(
-									'py-4 text-base',
+									'flex-1 py-4 text-base',
 									currentChapterActive && 'font-bold',
 									isChild && 'text-foreground-muted',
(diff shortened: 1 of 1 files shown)


apps/expo/components/book/reader/epub/LocationsSheetContent.tsx
```diff
@@ -169,39 +268,145 @@ const TableOfContentsListItem = ({ item }: { item: TableOfContentsItem }) => {
...
+							</Text>
+							<Text
+								className={cn(
+									'py-4 text-base text-foreground-muted',
+									currentChapterActive && 'font-bold',
+								)}
+								style={currentChapterActive && { color: textColor }}
+							>
+								{item.position || 'Not Available'}
+							</Text>
```
aaronleopold: I think we could probably add a `shrink-0` here, I noticed if a chapter is long enough in length the number gets pushed out

PR #929, opened 2026-02-07
:globe_with_meridians: Update translations and support locales

aaronleopold:


aaronleopold:
I've tried adding `en-gb` and `fr-ca` support to the project but it seems to kill the existing French translations. I'll try to see if perhaps I am missing a setting somewhere, or reach out to support for clarification

PR #941, opened 2026-02-21
:recycle: Unify timestamps and improve localized formats

aaronleopold:
This is a sweeping change to replace `dayjs` with `date-fns` and improve localized time formatting throughout the apps


aaronleopold:
I don't have time to check every spot on the web side of things for the updated formatting but going to merge this in

Issue #120, opened 2023-04-17
[FEATURE] Book club features


aaronleopold:
I think it would be really cute to add a suite of features aimed at facilitating book clubs

**Describe the solution you'd like**

- [x] User can create clubs
    - [x] Basic on/off RBAC, where a server admin may turn off this feature for a user. I think default in this case might be off.
- [ ] A user may invite another user to a club, where they are added **upon acceptance**
    -  NOTE: this requires a whole notification system 😬 
- [x] A club (entity) has:
    - [x] name + optional description
    - [x] members
    - [x] a book schedule
        - [x] configurable (but reasonably defaulted) interval between books (e.g. 1-2 month per book)
        - [x] book list with start/end dates for each (default interval, allow overriding)
- [ ] Clubs have a homepage
    - [ ] Current book (and next book) featured
    - [ ] User progress is displayed for each member of the club
        - [ ] Allow this to be turned off if a user would prefer others don't know progress
    - [ ] Basic chat board for each book that gets archived after each book rotation in schedule (**maybe**)
    - [ ] Admins of club have option to create/update book schedule

DieselTech:
When the book club session is closed or ended, the users profile could show have many clubs they have participated in. I guess that also brings up a whole new feature centered around profile cards. But would be great in a social reading context. I feel like if there is a book club option that it would start to have a lot of users added to servers just for that.

![image](https://github.com/stumpapp/stump/assets/30128380/6db9472d-3832-441e-a39f-d3703c3689b3)

aaronleopold:
> I guess that also brings up a whole new feature centered around profile cards. But would be great in a social reading context.

Yeah a whole new set of "public" user endpoints would have to be created to support this sort of thing. I like it though! Definitely something to keep in mind when I'm (finally, hopefully soon) able to give this feature some time. Once I finish implementing RBAC/ABAC Stump will be in a much better place for this kind of thing. One of my favorites though so super excited

PR #942, opened 2026-02-21
:sparkles: Overhaul book club backend and build out initial mobile app integrations

aaronleopold:
Relates to https://github.com/stumpapp/stump/issues/120

This is a big one, I've refactored a few of the core book club features in the backend and built out an entire (barebones) UI for the mobile app.

For the backend changes:

- I removed the explicit `book_club_schedule`
  - I originally thought that automated phases and lists of books would be neat but the overhead of implementing that kind of thing was too high. I'll consider automations in the future, but I really wanted to get this somewhere more than just an idea so punted it for later
- I removed the one-to-one relationship between book club books and discussions, so a club can create any amount of discussions for a book
- I added support for non-book-bound discussions, which are basically just pinned discussions for the club.
  - A `General` will be created by default for clubs
- I replaced the message like-based system with reactions
- I added support for uploading custom emojis for the server, if uploads are enabled (which can be used for reactions in book club messages)
- I removed the ability to add URL-based user avatars and just made it an upload-only system. For OIDC, if the provider has an avatar URL it will be downloaded and set upon registration

For the mobile app additions:

- I added a new preference for enabling/disabling book club features, which will hide the tab
  - If your user doesn't have access to book clubs in the server, though, the tab will be hidden regardless of this setting
- Add a very minimal screen for viewing book clubs
- Add a very minimal form for creating a book club
- Add an at-a-glance screen for book clubs. This is somewhat minimal for now but will change as the feature is built out more
  - See moderator(s)
  - Pinned discussions + discussions for current book (if set)
  - See current book (if set)
  - Link to previous discussions/books
- Add ability to search server for book to set as current book
- Add a barebones but functional chat interface:
  - Support inline replies
  - Support reactions
  - Support threads

The UI shape is there but do not expect to really use this feature yet. I am still keeping the dev-only flag on the web frontend for now. Things I didn't quite work on as part of this but will be coming:

- Voting system for selecting the next book
- Live chat updates, chat isn't live right now so requires manual refreshes (obviously not ideal)
- More moderation/management tools
- Invitation flow UI
- Basic settings flow


aaronleopold:
I'll be testing the migration a bit more to make sure nothing breaks and then I'll get this in shortly after

Edit: Tested a few migration paths and nothing seems to break, so should be good enough for experimental

Issue #718, opened 2025-08-22
[FEATURE] Library and File Explorer (Files tab) Improvements


konokodo:
**Is your feature request related to a problem?**

As someone who is reliant on the File Explorer (Files tab) feature, I have come across a couple of pain points:
- There is currently no option to set a default library tab (e.g. Series/Books/Files)
- There is no thumbnails on folders on File Explorer (Files tab)
- There is no option to only display archives and folders on File Explorer (Files tab)
- Requiring double click to navigate on folders and archives on File Explorer (Files tab)
- There is no File Explorer (Files tab) on OPDS

**Describe the solution you'd like**

- I would like an option for default library tab, like an option to set Files tab as default in particular.

- _File Explorer (Files tab)_ - **Thumbnail view**
I would like for the Files tab to have the same thumbnail view as Series and Books tab, for better usability and consistency across all tabs. 
When you select Series priority as a Library Pattern, the server generates thumbnails each folder. I believe these thumbnails should be used on the Files tab as well.
So, I propose for the Files tab to have the same Grid view (with big thumbnails) like in Series and Books tab, in addition to the current smaller grid view on Files tab without thumbnails as a separate view option for a total of 3 views (table, big grid, and small grid).

- _File Explorer (Files tab)_ - **Hide non-archive files**
I would like an option on the Files tab to hide files that are not zip/rar/cbz/cbr and other archives.

- _File Explorer (Files tab)_ - **Single click to navigate**
I would like an option or new defaults on the Files tab of only requiring a single click when navigating to a folder or opening an archive.

- A File Explorer (Files tab) view for a Library on OPDS would be much appreciated.

**Additional context**

I really like that Stump has a File Explorer (Files tab) as a core feature early on that is used as an alternative to the preset Library Patterns. I believe this feature is one of the things that makes Stump unique and less opinionated compared to other mainstream comic servers.

I am aware that the File Explorer feature exist to catch edge cases that cannot be covered by the two existing library patterns. Hence I have only suggested features that already exists on other parts of the reader and a couple of toggles, except for OPDS support which probably is a big undertaking but would be nice to have.

aaronleopold:
Hey! I really appreciate your feedback! I pretty much agree with all of it, but I do have some questions/comments for clarification:

> I would like an option for default library tab

Do you imagine this as config on the library or a personal preference? I.e. does it affect all users with access to the library?

[...]

PR #954, opened 2026-03-03
✨ Improve OPDS v1.2 book browsing and search (#718)

nitou2504:
## Summary

Addresses the OPDS portion of #718.

Hi! I've been using Stump for a while and it's amazing on performance for large libraries — noticeably faster than alternatives like Booklore. However, the only missing features or issues for my use case and as in the issue #718 (using Koreader that has OPDS v1.2 support only) were:

1. **Search only matched series** — so searching for a book title or author returned nothing. I know a workaround would be to have per author folders, but I have some libraries that aren'r organized like that :( .
2. **Book descriptions weren't showing** — the summary/description from metadata wasn't being exposed in the OPDS feed. So Book information only showed the size and file type.
3. **All books showed "Unknown author"** — even when the metadata had writer information, it wasn't being passed to the OPDS entries, so when downloading a book via OPDS they all were "Unknown author - filename.epub"

### What this PR does

**New endpoints:**
- `/books` — browse all books directly with pagination and search.
- `/books/latest` — recently added books.
- `/search/feed` — unified search that looks across libraries, series, and books (by title, summary, and author) instead of just series names.

**Metadata fixes:**
- Book descriptions now appear via the `<summary>` element.
- Authors are now pulled from metadata writers, so books no longer show as "Unknown author".
- `<content type="html">` added per OPDS spec.

I used Claude to help identify the relevant files and understand the existing implementation, then did the changes and testing myself.

### How I tested

- Built a local Docker image and ran it alongside the nightly container with the same data/config.
- Compared OPDS responses between both via `curl` to verify nothing broke.
- Tested on an actual KOReader device against a large ebook library.

### DIFF
--- a/core/src/opds/v1_2/feed.rs
+++ b/core/src/opds/v1_2/feed.rs
@@ -264,6 +264,7 @@ mod tests {
 			None,
 			None,
 			None,
+			None,
 		);
 
 		let feed = OpdsFeed::new(


aaronleopold:
I'll aim to try and review this by the weekend, thank you!

nitou2504:
Thanks for your great work !

Issue #945, opened 2026-02-24
[FEATURE] Option to disable chapter thumbnails and display a text list for series with no cover art


sumerian-circles:
### Is your feature request related to a problem?

Many manga, manhua, and webtoons do not have individual cover images for each chapter. So when the reader displays chapter thumbnails taking a page as the thumbnail it looks very unpleaseant

### Describe the solution you'd like

It would be cool if there was a preference setting (per series or per library) where you can switch the thumbnails off for chapters in a series and have it be a list with just chapter names and maybe some buttons (mark chapter as read,favorite ,and file info). The series thumbnail itself should stay as is. So basically changing the UI for series with no cover art to look more like what you would see on a manga site 

### Describe alternatives you've considered

I tried the list view on the web reader but it also includes thumbnails and looks a bit jarring like it dosent fit with the rest of the UI in my opinion. 

### Where would this feature be implemented?

- [x] Web App
- [x] Mobile App
- [ ] Server
- [x] Documentation

### Additional context

<img width="1634" height="1490" alt="Image" src="https://github.com/user-attachments/assets/8d8d21f8-6ab3-48b2-802c-1f9aa1441d3b" />
Some thing like this image but with the modern look of stump. Pretty much a "manga" style UI for specific series or a libraries so if you have a seperate library for manga/comics you can keep the original look witch looks good for comic libraries and change to the list view for manga libraries.
when you chose this prefrence you should no longer have to go trough the pre-reader UI (where it show chapter info and metadata) since most manga to my knowledge dont have per chapter metadata worth displaying. So when you click a chapter it just send you to the reader.

Extra things that would be cool:
chapters you mark as read get tinted to signify they were read
You have a button continue/start reading in the series UI or a sligthly modified home page entry (if you chose this preference it should just be the series title and thumbnail with the chapter name and page number in the home page entry also the series should be excluded from "your next read" since it would make the home page cluttered with multiples of the same thumbnail)

aaronleopold:
There are a few feature requests in here, at least based on my read of your ask, so to make sure I capture it all:

1. Make grid vs list display options less "global" and tied to specific libraries or library types (e.g., either track layouts per library or per library type). Note that library types do not exist today, but eventually will as part of the metadata fetching effort.
2. An option (per library or library type) to skip the book overview page and just go to the reader
3. Some indication of progression/completion in list layout
4. Continue/start reading button at series level view

I wasn't 100% sure what you meant by this bit though:

[...]

PR #960, opened 2026-03-12
:sparkles:/:lipstick: Library-bound layout preferences and unify UI's

aaronleopold:
Relates to #945

This PR does a few primary things:

- Add new `skip_book_overview` preference for libraries
- Redesign web overview page headers
- Decrease "globalness" of layout preferences
   - Grid vs list layouts are now scoped primarily to a library and the modality being viewed. E.g., instead of books vs series it is now `libraryId-books` or `libraryId-series` etc
- Support more control options for list view (e.g., columns and ordering) 
- Reorganize (mobile) book overview screen
- Support markdown descriptions on mobile

### DIFF
--- a/packages/graphql/src/client/graphql.ts
+++ b/packages/graphql/src/client/graphql.ts
@@ -1005,8 +1005,18 @@ export type Library = {
 };
 
 
+export type LibraryGenresArgs = {
+  sort?: InputMaybe<OrderDirection>;
+};
+
+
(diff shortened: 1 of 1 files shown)


aaronleopold:
I gave this some additional testing and review time since it is larger than ideal, but I think this is good to go once CI is green. Thanks @Arklaum for the improvements/fixes!

PR #972, opened 2026-03-15
:bug: (expo): Fix OPDS v2 pagination issues

### LINKED ISSUE #961: [BUG]  Mobile App OPDS 2.0 Pagination Incomplete?
### Description of the Bug

I've been playing around, mostly with Codex, and it seems like the pagination support in the mobile app is not all there?  I do see one particular page in the Codex feed, the "All Issues" just off the top, and it looks like the app is doing an on-demand paging during scrolling. I haven't gotten to the end, as it's a bit slow, but presumably it'll get through all the 200-ish pages of 100 items each.

There are other feeds where there's there no indication of any paging that might be happening.  For example, Codex has a folders view, and I can drill down to a page for a folder with a few hundred subfolders ("subsections"), and it displays exactly 100 items, which is the items-per-page size from the server.  

I think the OPDS is well-formed with "first", "previous", "next", and "last" links?

```
{
    "metadata": {
        "title": "series",
        "subtitle": "desc",
        "modified": "2026-03-08T16:32:08.423000-07:00",
        "itemsPerPage": 100,
        "currentPage": 2,
        "numberOfItems": 246
    },
    "links": [
        {
            "href": "/codex/opds/v2.0/f/4/2?orderBy=date&orderReverse=True&topGroup=f",
            "rel": [
                "self",
                "up"
            ],
            "type": "application/opds+json"
        },
        {
            "href": "/codex/opds/auth/v1",
            "rel": "http://opds-spec.org/auth/document",
            "type": "application/opds-authentication+json"
        },
        {
            "href": "/codex/opds/v2.0/s/0/1{?query}&topGroup=s",
            "rel": "search",
            "type": "application/opds+json",
            "templated": true
        },
        {
            "href": "/codex/opds/v2.0/",
            "rel": "start",
            "type": "application/opds+json"
        },
        {
            "href": "/codex/opds/v2.0/f/0/1?orderBy=date&orderReverse=True&topGroup=f",
            "rel": "top",
            "type": "application/opds+json"
        },
        {
            "href": "/codex/opds/v2.0/f/4/1?orderBy=date&orderReverse=True&topGroup=f",
            "rel": [
                "first",
                "previous"
            ],
            "type": "application/opds+json"
        },
        {
            "href": "/codex/opds/v2.0/f/4/3?orderBy=date&orderReverse=True&topGroup=f",
            "rel": [
                "last",
                "next"
            ],
            "type": "application/opds+json"
        }
    ],
```

Probably not top priority right now, but wanted get this on the board for now.

I feel like paging for a mobile app is a tough thing to get right, as well, as you've got pages from the server that don't match the screen size, and the user might have to either scroll and page, or you hide it all with the scroll-to-page method, which leads to weird buffering if the user wants to scroll quickly to the end.  Another approach might be to do only paging, where app re-pages and there's no scrolling for the user, _just_ paging, since the app is pre-caching to make the new virtual pages fit the screen.  That approach could allow for the classic top nav buttons with fast jumps.

Thanks!

### Expected Behavior

Paging of large lists will be handled by the app.

### Steps To Reproduce

With Codex (and maybe other servers) that has a large library with lists that are larger than the servers page size (which is 100 on Codex), navigate to one of those lists, and see that you can't always access the whole list.

- [x] Web App
- [ ] Mobile App
- [ ] Server

### Server & Client Details

Recent Codex version 1.9.24

Current versions of the Stump mobile app on iOS.

### Additional context

_No response_

aaronleopold:
Should resolve #961

I made sure the pagination resolving logic checks for `rel` values that are arrays which should fix the problem, but I'll make sure to test more thoroughly with Codex before I am comfortable merging this as the fix.

### DIFF
--- a/apps/expo/components/opds/PublicationGroup.tsx
+++ b/apps/expo/components/opds/PublicationGroup.tsx
@@ -125,7 +125,7 @@ export default function PublicationGroup({
 	return (
 		<View>
 			<View className="flex flex-row items-center justify-between px-4 pb-3">
-				<ListLabel>{metadata.title || 'Publications'}</ListLabel>
+				<ListLabel className="ios:px-4 px-2">{metadata.title || 'Publications'}</ListLabel>
 
 				{selfURL && <FeedSelfURL url={selfURL} />}
 			</View>


aaronleopold:
I've done some testing and I _think_ this resolves the issue, but I haven't been able to fully replicate the flow outlined in the ticket which exposes a screen full of paged navigation links. I'll tentatively merge and get feedback when I am able to push another build

PR #974, opened 2026-03-18
:wrench: Increase GraphQL `limit_depth` to 15

bct:
Right now the graphql playground displays an error:

    Query is nested too deep.

And is unable to display documentation.

image

The `limit_depth` call was added in https://github.com/stumpapp/stump/commit/c5d9de9eb80e5d7ff8971e1f713f7b8e82e0d678 , and the playground seems have been broken since then.

### DIFF
--- a/crates/graphql/src/schema.rs
+++ b/crates/graphql/src/schema.rs
@@ -28,7 +28,7 @@ pub async fn build_schema(ctx: CoreContext) -> AppSchema {
 	)
 	// AccessRole is used in a serialized json for SmartList so we need to register it manually
 	.register_output_type::<AccessRole>()
-	.limit_depth(10)
+	.limit_depth(15)
 	.data(ctx);
 
 	add_data_loaders(schema_builder, conn).finish()


bct:
Original commit that added the limit: https://github.com/stumpapp/stump/commit/9e43dfccde9982b6947bffaa1aaac6c2780327ce

I chose the value 15 arbitrarily, I didn't attempt to find a lower value that works.

bct:
GraphiQL produces an introspection query like this:

```
query IntrospectionQuery {
  ...snipped some normal looking stuff...

  fragment TypeRef on __Type {
    kind
    name
    ofType {
      kind
      name
      ofType {
        kind
        name
        ofType {
          kind
          name
          ofType {
            kind
            name
            ofType {
              kind
              name
              ofType {
                kind
                name
                ofType {
                  kind
                  name
                  ofType {
                    kind
                    name
                    ofType {
                      kind
                      name
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

Which is presumably what's failing. idk if that's normal or what the cause might be.
I imagine a non-playground client wouldn't run into this.

aaronleopold:
Ah, okay. Yeah, that'll do it

> I imagine a non-playground client wouldn't run into this.

Yeah, I wouldn't think so

PR #988, opened 2026-03-29
💄 (expo): Add various confirmation alerts

Arklaum:
Here we have the following:
- Added confirmation alerts to the book menu (delete read history, clear progress, mark as read).
- Also I changed some of the alerts that existed throughout the app (e.g. the server delete and delete local library alerts) to use the same Alert API. There is still one from `'react-native-dialog'`, but `/components/ui/dialog.tsx` is unused now.
- And I've updated and cleaned up the long value card rows to look better I think.


apps/expo/components/ui/card.tsx
```diff
@@ -138,22 +144,132 @@ function Stat({ label, value, suffix }: StatProps) {
...
+					{isExpandable && !expand && (
+						<View className="absolute bottom-0 right-0 z-10 items-center justify-center px-1">
+							<Text
+								style={{ color: accentColor || colors.fill.brand.DEFAULT }}
+								className="font-medium"
+							>
+								See more
+							</Text>
+						</View>
+					)}
```
aaronleopold: I don't see a way to close it once expanded

Arklaum:
> Did you interact meaningfully with the book club features or did you mostly just go through to make the same changes you made elsewhere?

I only made a book club to change the alerts. I won't personally be using the feature, so if at any time you want specific thoughts on something just ask and I'll try it out, but I won't be able to give much feedback from any long term use perspective.

PR #1032, opened 2026-04-02
:passport_control: Restrict default user permissions

### LINKED ISSUE #917: [FEATURE] Option to prevent user from changing their own password

I would like to set up a 'public' user. Ideally this would mean someone couldn't go in and change the password of the account to lock everyone else out.

I would like to prevent a specific user from changing their own password.

Maybe having a user with no password required?

- [x] Web App
- [ ] Mobile App
- [x] Server
- [ ] Documentation

### Additional context

_No response_

aaronleopold:
Resolves #917


aaronleopold:
I was hoping to have time to test this a bit more, but won't have the time this weekend. I'll go ahead and merge it in and let the experimental testing process do its thing

PR #1034, opened 2026-04-02
🔨 Use the same Rust version from the Nix flake

Kernald:
The Nix flake was previously using its own Rust version, floating depending on whatever nixpkgs currently had. This re-uses the version from rust-toolchain.toml instead, making sure there's no inconsistencies.


aaronleopold (approved):
Thank you!

PR #1045, opened 2026-04-03

aaronleopold:


retmas-gh:
Since I'm not sure I can edit the original file on Crowdin for the source language I will just mention here that the following strings should be fixed :) :
-  `Stump features are optional and can be turn them off if you do not need them`
- `There was an error deleting all all downloads from {{serverName}}`

aaronleopold:
Ha, thank you! At least this is your signal a human actually writes these 😀

I'll fix later when I wake up

PR #1046, opened 2026-04-04
Basic Kobo sync implementation

### LINKED ISSUE #361: [FEATURE] Kobo Sync and Send to Kindle
Currently, OPDS is the only way to distribute content to readers, and I am unaware that any commercial reader supports it. On the other hand, [caliber-web](https://github.com/janeczku/calibre-web) implemented a way to send files via email to Kindle devices and even hosted the Kobo API to trick a device into syncing with the library like it was Kobo's servers.

*EDIT*: An interesting library to look at, would be [kobink](https://github.com/potatoeggy/kobink). It could be integrated directly into the code, as a module, or put in a separate docker container to handle the syncing of ebooks with a Kobo device.

Make a similar feature set available for stump! While I do like Clibre-web, I do not like how Calibre manages the data in the background and how the user cannot rename things as he wants (database issues and so on). That is why I am using Stump now, but QoL features like this are currently missing.

**Additional context**
I am aware, that other features and bugs are higher on the priority list, but it would be cool to discuss possible solutions to this Problem! I was maybe thinking of taking the implementation of calibre-web, which consists of the [kobo.py](https://github.com/janeczku/calibre-web/blob/master/cps/kobo.py), [kobo_auth.py](https://github.com/janeczku/calibre-web/blob/master/cps/kobo_auth.py), and [kobo_sync_status](https://github.com/janeczku/calibre-web/blob/master/cps/kobo_sync_status.py), and implementing a library of its own, so it can be imported into other projects like stump. I will see how much I can do on my own, but getting help from someone experienced would be great!

bct:
This PR has something like a minimum viable feature set:
- syncing available epubs to the Kobo, with cover images and most metadata
- paginated & incremental sync
- downloading epubs to the Kobo
- user-facing documentation

Notable features that it does not include:
- syncing read progress in either direction
- converting EPUBs to KEPUBs. the Kobo can read EPUBs, but the experience is worse - page turns are slower, progress is not computed as accurately, etc.
- proxying to the Kobo store. if you point your device at Stump, you will not be able to sync books from the Kobo store.
- selecting which books should be synced to the Kobo. all EPUBs available to an API key will be synced to the Kobo.


apps/server/src/routers/kobo/sync.rs
```diff
@@ -0,0 +1,692 @@
...
+			item_count = sync_items.len(),
+			"Generated a page of Kobo sync items"
+		);
+
+		Ok(sync_items)
+	}
+}
+
+#[cfg(test)]
+mod tests {
```
bct, line 291: I did not find any integration tests that use the database, so I threw something together, including functions for generating data. Maybe there's a reason that there aren't tests like this, they turned out to be pretty ugly.

crates/models/src/entity/kobo_sync_session.rs
```diff
@@ -0,0 +1,77 @@
...
+// TODO: pluralize?
+#[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
+#[sea_orm(table_name = "kobo_sync")]
+pub struct Model {
+	#[sea_orm(primary_key, auto_increment = false, column_type = "Text")]
+	pub id: String,
+	#[sea_orm(column_type = "Text")]
+	pub user_id: String,
+	#[sea_orm(column_type = "Json")]
+	pub media_ids: MediaIds,
```
bct, line 19: Kobo syncs are paginated (a single sync session is split over multiple requests) and incremental (each sync session only returns data that has changed since the last completed sync).

There are several types of updates that the server can send to the device:
- book added
- book removed
- book changed
- reading state changed

One thing to be careful about is lost updates - if we skip sending a "book added" update, that book will never show up on the device.

The approach I've taken is:
- when a new sync session is initiated, store the IDs of media that has changed since the last sync session
- as each page of the sync is retrieved, load the page of media and determine the type of update (added / changed) based on the media's timestamps

This is fairly simple. Explicitly recording what is being synced should make syncs robust even when the library is changing during the sync process. Maybe it's too simple - I'm not sure that it can support "book removed" updates.

Other approaches would probably involve more explicit tracking of what has been synced to each device. For example, Komga determines what has changed between syncs by storing the IDs of the entire library with each sync and diffing them.

aaronleopold:
I am excited to give this a review! I'm hoping to have some time later today or tomorrow. Thank you for all of your work!

apps/server/src/routers/kobo/sync.rs
aaronleopold: > Maybe there's a reason that there aren't tests like this, they turned out to be pretty ugly

It is a bit of a chore, honestly. Testing in Rust isn't as smooth or convenient as other languages (mostly thinking of TypeScript), and since I do it as part of my job it isn't something I necessarily am itching to do for fun. Especially because of this, I really appreciate you adding some in

apps/server/src/routers/kobo/router.rs
```diff
@@ -0,0 +1,345 @@
...
+				.insert("x-kobo-sync", HeaderValue::from_static("continue"));
+		}
+
+		match self.sync_token.try_to_header_value() {
+			Ok(sync_token) => {
+				response
+					.headers_mut()
+					.insert("x-kobo-synctoken", sync_token);
+			},
+			Err(e) => tracing::error!(?e, "Failed to produce Kobo sync token"),
```
aaronleopold: Should this be a true failure scenario? Or is it acceptable from Kobo's perspective to respond without a sync token?

apps/server/src/routers/kobo/router.rs
```diff
@@ -0,0 +1,345 @@
...
+	let result = get_media_thumbnail_by_id(&ctx, &req.user(), book_id).await?;
+
+	// the Kobo only supports JPEGs, and doesn't need large thumbnails.
+	let jpeg_buffer = tokio::task::block_in_place(|| {
+		let converted = GenericImageProcessor::generate(
+			&result.data,
+			ImageProcessorOptions {
+				format: SupportedImageFormat::Jpeg,
+				// TODO: ImageResizeMethod::FitWithin?
+				// (similar implementation to ScaledDimensionResize)
```
aaronleopold: Happy to help with this, if needed

apps/server/src/routers/kobo/router.rs
```diff
@@ -0,0 +1,345 @@
...
+}
+
+async fn book_download(
+	State(ctx): State<AppState>,
+	Extension(req): Extension<AuthContext>,
+	Path(KoboAPIKeyAndBookId { book_id, .. }): Path<KoboAPIKeyAndBookId>,
+	headers: HeaderMap,
+) -> APIResult<impl IntoResponse> {
+	// TODO: is this reasonable? would it ever be useful to have kobo sync permission without
+	// download file?
```
aaronleopold: I think there are similar awkward points with OPDS, since not all formats are streamable (e.g., ebooks) and so not having the permission effectively means you cannot use OPDS for those. IMO it is acceptable, the granularity of permissions makes admins have more control (even if they are logically a bit silly)

apps/server/src/routers/kobo/sync_token.rs
```diff
@@ -0,0 +1,88 @@
...
+use axum::http::HeaderValue;
+use base64::{engine::general_purpose::STANDARD_NO_PAD as BASE64, Engine};
+use std::str::Utf8Error;
+
+use reqwest::header::{InvalidHeaderValue, ToStrError};
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SyncToken {
```
aaronleopold: It might be useful to make these regular comments in `SyncToken` doc comments, i.e. using `///` instead of `//`. That way it shows up in hints for editors supporting it

apps/server/src/routers/kobo/sync.rs
```diff
@@ -0,0 +1,692 @@
...
+
+impl KoboSync {
+	async fn find(db: &DatabaseConnection, user: &AuthUser, id: String) -> Option<Self> {
+		kobo_sync::Entity::find_by_id(id)
+			.one(db)
+			.await
+			.ok()?
+			.and_then(|m| {
+				if m.user_id != user.id {
+					tracing::warn!("Attempted to use another user's Kobo sync token");
```
aaronleopold: We should probably add the user's id to the log

apps/server/src/routers/kobo/sync.rs
```diff
@@ -0,0 +1,692 @@
...
+			session.model.id.clone(),
+			&session.model.media_ids.0,
+			offset,
+			limit,
+			session.model.previous_sync_at,
+		))
+	}
+}
+
+pub struct SyncPage<'a> {
```
aaronleopold, line 187: Same remark as before re: the struct field comments could be made doc comments

apps/server/src/routers/kobo/sync.rs
```diff
@@ -0,0 +1,692 @@
...
+		let reading_sessions = reading_session::Entity::find()
+			.filter(reading_session::Column::UserId.eq(self.user.id.clone()))
+			.filter(reading_session::Column::MediaId.is_in(&self.media_ids))
+			.all(self.db)
+			.await?;
+
+		// TODO: this is not the correct way to do this!
+		// we should be able to do this in a single query.
+		// also, we need to properly handle cases where there are multiple sessions for a single
+		// piece of media.
```
aaronleopold: There should never be multiple, active reading sessions for a single media per user. Since you filtered by `user_id`, and there is a unique constraint on `media_id`+`user_id`, I think we are safe from this case

apps/server/src/routers/kobo/sync.rs
```diff
@@ -0,0 +1,692 @@
...
+		for stmt in tables {
+			db.execute(db.get_database_backend().build(&stmt)).await?;
+		}
+
+		Ok(())
+	}
+
+	// note that None here means "use some default", not necessarily "set the value to None".
+	// that may make it impossible to set some values to None. I'm not sure how to avoid that while
+	// keeping good ergonomics.
```
aaronleopold: You could just manually implement `Default`, instead. That way you don't have to wrangle with `None`s unless explicitly testing for it

apps/server/src/routers/kobo/sync.rs
```diff
@@ -0,0 +1,692 @@
...
+			serde_json::json!({}),
+			None,
+			3,
+		)
+		.await
+		.expect("failed to initiate sync");
+
+		assert_eq!(vec!["book-1", "book-2", "book-3"], first_page.media_ids);
+		assert_eq!(3, first_page.sync_token.next_offset);
+		assert_eq!(true, first_page.should_continue);
```
aaronleopold: You can do `assert!(first_page.should_continue);`, as well

apps/server/src/routers/kobo/sync.rs
```diff
@@ -0,0 +1,692 @@
...
+			&user,
+			Some("kobo-1"),
+			serde_json::json!({}),
+			Some(&first_page.sync_token),
+			3,
+		)
+		.await
+		.expect("failed to continue sync");
+		assert_eq!(vec!["book-4", "book-5"], second_page.media_ids);
+		assert_eq!(false, second_page.should_continue);
```
aaronleopold: Similarly, `assert!(!second_page.should_continue);`

core/src/kobo/entity.rs
```diff
@@ -0,0 +1,160 @@
...
+			external_ids: vec![],
+			genre: DUMMY_UUID.to_string(),
+			is_eligible_for_kobo_love: false,
+			is_internet_archive: false,
+			is_pre_order: false,
+			is_social_enabled: true,
+			isbn: m.metadata.clone().and_then(|mm| mm.identifier_isbn),
+			language: "en".to_string(),
+			phonetic_pronunciations: Empty {},
+			publication_date: None, // TODO
```
aaronleopold: You could eventually construct manually by merging `month`, `day`, and `year` from metadata into a fully-resolved date (or just providing what exists, if not all three). I do something similar somewhere in the code, I think in the frontend somewhere.

crates/models/src/entity/kobo_sync.rs
```diff
@@ -0,0 +1,63 @@
+use sea_orm::FromJsonQueryResult;
+use sea_orm::{entity::prelude::*, prelude::async_trait::async_trait, ActiveValue};
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, FromJsonQueryResult)]
+pub struct MediaIds(pub Vec<String>);
+
+// TODO: pluralize?
```
aaronleopold: If matching what is most common in Stump, yeah it would be `kobo_syncs` for the actual table name but `kobo_sync` for the model

crates/models/src/entity/kobo_sync_session.rs
aaronleopold: > Maybe it's too simple - I'm not sure that it can support "book removed" updates.

Just to dump an idea, instead of tracking media ids alone you could do something like:

```rust
pub struct SyncChange {
    pub media_id: String,
    pub change_type: ChangeType, (added/modified/removed)
}
```

And then account for the chagne types separately in `begin_new_sync`. Although, Stump doesn't use soft deletes quite that much (see https://github.com/stumpapp/stump/issues/336).

Also, just to make sure I understand how this works: in a world where we don't support book removed updates, the worst that could happen is someone deletes a book from Stump and it would just stay on their device until they do a full resync?

***

As far as general approach, what you have feels rational to me (granted I have not used Kobo sync features before). The way you described how Komga approaches it, while definitely heavy handed, does feel a bit more robust if we _need_ to account for removals.

aaronleopold (commented):
I think this is a fantastic start to the feature! Your work is very easy to read through and follow, which I appreciate a lot.

I didn't see any noteworthy Rust crimes, either, clippy is pretty good about catching those. I know there were a lot of TODOs, but I tried not to focus them. Feel free to poke in any areas you were hoping to get more feedback/general eyes/discussion on.

apps/server/src/routers/kobo/router.rs
bct: Good idea. Responding without a sync token wouldn't be a disaster, but this should never happen, and returning an error keeps things simple.

apps/server/src/routers/kobo/sync.rs
bct: Ah! I saw the `has_many` on the relation and got confused. Thanks for clarifying, that makes this a lot easier.

bct: Are there cases where a book might have `finished_reading_session`s but no `reading_session`?

aaronleopold: Yeah that scenario is very likely. A `finished_reading_session` is just a historical record of an instance where you completed the book. A `reading_session` is an active session that is maintained only while actively reading, and after completion gets converted into a `finished_reading_session` and deleted. 

It's been brought up before that it isn't the most intuitive, but that's how it works today at least

crates/models/src/entity/kobo_sync_session.rs
bct: > Also, just to make sure I understand how this works: in a world where we don't support book removed updates, the worst that could happen is someone deletes a book from Stump and it would just stay on their device until they do a full resync?

It would stay there until they manually deleted it from the device - I don't think there's a way to do a full resync except to delete everything on the device.

Another aspect of this is that if you delete something from the device it notifies the server of that deletion. If we were syncing from e.g. a smart list we could delete that media from the list, and then sync that deletion to _other_ devices syncing from the list.

In any case I think it's fine to ignore this for now. There are multiple options for modifying this approach in the future.

aaronleopold (approved):
I unfortunately haven't had much free time to really do an independent learning for specifics on how Kobo sync works, but your code is logical and looks great to me so I am going to approve this PR.

I really appreciate some of the cleanup and consolidation refactors you did, and all of the testing utils you added, _and_ the documentation updates. Thank you!

> Kobo can read EPUBs, but the experience is worse - page turns are slower, progress is not computed as accurately, etc

Good to know, I am curious what would be involved. Seems like it wouldn't be terrible, I took a really quick glance and found this [neat tool](https://github.com/Glydric/Rulibre/), which looks to call out to [kepubify](https://github.com/pgaskin/kepubify).

***

I'll aim to merge this later this evening, thank you again!

bct:
> Good to know, I am curious what would be involved. Seems like it wouldn't be terrible, I took a really quick glance and found this neat tool, which looks to call out to kepubify.

Yeah, I _think_ it's just a call to kepubify. This also seems to be a prerequisite for syncing read progress - progress is tracked using sentence-level IDs that are embedded in the kepub.

aaronleopold:
> progress is tracked using sentence-level IDs that are embedded in the kepub.

Interesting, initially I started to think through what support for `kepub` would look like throughout Stump (e.g., the readers and internal progress tracking for web and mobile apps) but then realized that I imagine folks wanting kobo sync don't necessarily care about reading within the Stump ecosystem and will just read on their kobo. I appreciate the context!

PR #1060, opened 2026-04-05
:wrench: (expo): Change configured orientation

aaronleopold:
Should allow the app to go in either orientation, but I haven't tested it yet


aaronleopold:
I tested and it does seem to work, so I am going to merge this in. Note that not all the screens look great on landscape, but that can be fixed another time

PR #1063, opened 2026-04-06
🌐 (expo): Epub locale strings and locale debug option

Arklaum:
I've added some locale strings for the epub settings, an option to display the locale strings (which would be way more useful on web right now), and changes the `'book'` vs `'global'` picker to a toggle that says "Override global settings".


apps/expo/components/book/reader/epub/controls/ReadingProgression.tsx
```diff
@@ -3,30 +3,37 @@ import { useShallow } from 'zustand/react/shallow'
...
+
 	return (
-		<Card.Row label="Reading Direction">
+		<Card.Row label={t(getKey('label'))}>
 			<Picker
 				value={store.readingDirection}
 				// @ts-expect-error PickerOption type mismatch
-				options={READING_DIRECTION_OPTIONS}
+				// FIXME: epub settings use 'ltr' and 'rtl'. image reader settings use 'LTR' and 'RTL'
+				// causes no text to show up in epub reader if last changed on image reader
```
aaronleopold, line 30: Good callout, thank you. I think the native enums just need to align with the server gql ones

apps/expo/components/book/reader/image/ImageReaderSettingsSheet.tsx
```diff
@@ -0,0 +1,90 @@
...
+			detents={[0.5, 1]}
+			grabber
+			scrollable
+			backgroundColor={IS_IOS_24_PLUS ? undefined : colors.background.DEFAULT}
+			grabberOptions={{ color: colors.sheet.grabber }}
+			insetAdjustment="automatic"
+			style={{
+				// add a slight tint if global settings are not used
+				// this visually helps to confirm the switch did something if no settings change
+				backgroundColor: overrideGlobalSettings ? backgroundColor : undefined,
```
aaronleopold, line 48: Nice touch

aaronleopold (approved):
I won't have time to try it out locally, but the changes themselves look good to me! Thank you for the few fixes, as well

apps/expo/components/book/reader/image/ImageReaderSettingsSheet.tsx
Arklaum: It is quite nice, but I completely forgot about animating it. I'll try to add that at some point

apps/expo/components/book/reader/epub/controls/ReadingProgression.tsx
Arklaum: I just noticed I didn't write my comment that well, by "no text" I meant no text in that specific epub setting picker saying "Left to right" or "Right to left"

aaronleopold: You're totally good, that is exactly as I understood what you wrote to mean

PR #1075, opened 2026-04-10
➖ Replace `once_cell::Lazy` with `std::sync::LazyLock`

DanikVitek:
Remove the dependency on `once_cell` by replacing its uses with the `std` equivalents


aaronleopold (approved):
Thank you!

PR #1078, opened 2026-04-12
✨ Add type-to-confirm dialog for library deletion

### LINKED ISSUE #1074: [FEATURE] Stronger confirmation prompt to delete a library

I'm an idiot and just deleted my library while wanting to clean up missing files, again.

It would be nice if the library deletion dialog (and maybe other deletion dialogs) would ask e.g. to type the name of the library to confirm. It's the third time I go to run the "Clean library" process after reorganising files and delete the library instead.

Not being an idiot, but unfortunately that doesn't seem to be an option.

- [x] Web App
- [ ] Mobile App
- [ ] Server
- [ ] Documentation

### Additional context

Happy to contribute a PR on that one!

Kernald:
## Summary

This adds a reusable TypeToConfirmModal component, requiring to type the name of an entity before the confirm button gets enabled, and then uses it for library deletion. This fixes #1074 but also establishes a pattern that can be reused if we eventually add series/book removal.

## Screenshots

image

### DIFF
--- a/packages/browser/src/components/library/DeleteLibraryConfirmation.tsx
+++ b/packages/browser/src/components/library/DeleteLibraryConfirmation.tsx
@@ -78,13 +78,11 @@ export default function DeleteLibraryConfirmation({
 		}
 	}, [error])
 
-	const i18nValues = { type: 'library' }
-
 	return (
 		<TypeToConfirmModal
-			title={t(getKey('title'), i18nValues)}
-			description={t(getKey('description'), i18nValues)}
-			confirmText={t(getKey('confirm'), i18nValues)}
+			title={t(getKey('title'))}
+			description={t(getKey('description'))}
+			confirmText={t(getKey('confirm'))}
 			confirmVariant="danger"
 			isOpen={isOpen}
 			onClose={onClose}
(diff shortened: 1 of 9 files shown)


packages/i18n/src/context.ts
```diff
@@ -4,7 +4,7 @@ import { AllowedLocale } from './config'
 
 export type LocaleContextProps = {
 	locale: AllowedLocale
-	t: (key: string) => string
+	t: (key: string, options?: Record<string, string>) => string
```
aaronleopold: I don't actually see this being used anywhere, although I think I understand the basic idea was to have this handle the replacements internally if specified? Maybe the i18n library handles this for us, automatically? I didn't realize, if so

aaronleopold (approved):
This is great, thank you! I just had one question (that maybe is just solved by the library itself, which would be awesome)

packages/i18n/src/context.ts
Kernald: This might be a leftover of me trying to get interpolation working and not actually needed - I'll give it another look later today. The root of the issue was that the missing `fallbackNS` when initialising the i18n library, and my locale being `en-AU` rather than `en-US`, which I figured out later.

Kernald: Removed, it's indeed not necessary :-)

Kernald: Ugh CI seems to disagree with that, but it seemed to work locally?

aaronleopold: Ah, I see, it's just a type error not runtime that's why it worked. It just doesn't like that you kept those `i18nValues` in the `t(..)` calls in `DeleteLibraryConfirmation.tsx` after removing the second arg from the function signature in context. I don't mind making it happy tomorrow when I have a moment

PR #1082, opened 2026-04-13
:wrench: Add optional env var to configure directory for Stump.log

henrywhitaker3:
Adds a new env var `STUMP_LOG_DIR` which updates the directory that the `Stump.log` file is stored in. If not set, it defaults to the current behaviour/config dir


aaronleopold (approved):
Thanks!

PR #1084, opened 2026-04-13
🔨 Expand t() to take an optional record for interpolation

Kernald:
This allows keys to contain variables replaced automatically by i18next, rather than doing manual replacing or composition.

### DIFF
--- a/packages/i18n/src/context.ts
+++ b/packages/i18n/src/context.ts
@@ -4,7 +4,7 @@ import { AllowedLocale } from './config'
 
 export type LocaleContextProps = {
 	locale: AllowedLocale
-	t: (key: string) => string
+	t: (key: string, options?: Record<string, string>) => string
 }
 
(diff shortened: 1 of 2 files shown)


packages/i18n/src/context.ts
```diff
@@ -4,7 +4,7 @@ import { AllowedLocale } from './config'
 
 export type LocaleContextProps = {
 	locale: AllowedLocale
-	t: (key: string) => string
+	t: (key: string, options?: Record<string, string>) => string
```
aaronleopold: I think based on the docs you shared yesterday, this could technically be `Record<string, unknown>`. I think fine as-is, though

aaronleopold (approved):
Thank you!

packages/i18n/src/context.ts
Kernald: Replaced

PR #1090, opened 2026-04-16
🔨 Clean up Android development environment

Kernald:
This does a couple things:
 - Update the Nix flake to versions that match what's expected from other dependencies (Expo mainly), and add a few missing env variables in that flake
 - Add a direct dependency on `expo-dev-client`, making the first run a lot smoother - `yarn android` now does the right thing entirely

### DIFF
--- a/apps/expo/package.json
+++ b/apps/expo/package.json
@@ -64,6 +64,7 @@
 		"expo-build-properties": "~1.0.10",
 		"expo-constants": "~18.0.12",
 		"expo-crypto": "~15.0.8",
+		"expo-dev-client": "~6.0.20",
 		"expo-file-system": "~19.0.21",
 		"expo-font": "~14.0.10",
 		"expo-glass-effect": "^0.1.8",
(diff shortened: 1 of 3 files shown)


apps/expo/package.json
```diff
@@ -64,6 +64,7 @@
 		"expo-build-properties": "~1.0.10",
 		"expo-constants": "~18.0.12",
 		"expo-crypto": "~15.0.8",
+		"expo-dev-client": "~6.0.20",
```
aaronleopold, line 67: The current version looks to be `55.0.27`, I'd say let's update to a newer version. `6.0.20` was published [4 months ago](https://www.npmjs.com/package/expo-dev-client/v/6.0.20), I think they (expo) just reconfiged all the expo libraries to stay more aligned with SDK version. We didn't miss 49 major versions 😄

aaronleopold: Ah, actually I see this is based into `nightly` which doesn't have the upgrade to SDK 55 changes. So I think this is good as-is and when I merge into that branch I'll update it there accordingly.

aaronleopold (approved):
Thank you!

PR #1091, opened 2026-04-16
🐛 (expo): Fix various image reader bugs

Arklaum:
The issues with the image reader that have been fixed are:
- image paging, where the sides of the screen and the sides of the page did not properly align when different types of pages existed (it stopped like half way through each page after a landscape page). I don't think `getItemType` does anything for performance anyways
- on tapping to change page, if zoomed in and recently panned and the panning momentum hasn't yet stopped, the page resets to the wrong place
- and updated the image library which has since fixed the placeholder weirdness


aaronleopold (approved):
Thank you!

PR #1096, opened 2026-04-18
🐛 Sort dynamically added tags alphabetically in TagSelect

Kernald:
Tags created inline via the "Add" button were appended to the end of the options list instead of being sorted into position.


aaronleopold (approved):
Thank you! And thank you for adding tests

PR #1110, opened 2026-04-22
:sparkles: Intake tags from ComicInfo.xml during scans

### LINKED ISSUE #651: Create tags from ComicInfo.xml
### Discussed in https://github.com/stumpapp/stump/discussions/650

<div type='discussions-op-text'>

<sup>Originally posted by **SPKR4DED** April 23, 2025</sup>
Could you add tag reading/filtering via ComicInfo.xml. </div>

season179:
## Summary

Closes #651.

Library scans now pick up the `<Tags>` element from ComicInfo.xml (and the equivalent `tag` / `tags` entries from Calibre OPF metadata) and link them to the media as first-class `Tag` entities via the `media_tags` junction table, so they appear in the UI alongside manually-assigned tags and can be used for filtering the same way.

### Behavior

- **Additive**: linking only *adds* missing tags — a rescan will never remove a tag the user assigned manually through the UI.
- **Idempotent**: re-scanning an unchanged file is a no-op; existing links are detected and skipped.
- **Comma-split**: a value like `"epic, high-fantasy"` becomes two tags, matching how the UI already stores them.
- **Reuses existing tags**: if a tag with the same name already exists, we link to it instead of creating a duplicate.

### Where it plugs in

- `ProcessedMediaMetadata` gains an `Option<Vec<String>>` `tags` field. The XML deserializer uses the existing `string_list_deserializer`; the HashMap (OPF) path splits comma-separated values on the way in.
- `BuiltMedia` carries the extracted tag names alongside the `media` / `media_metadata` active models. They're intentionally kept out of `media_metadata` since tags are their own entity.
- The scanner's `create_media` / `update_media` paths, plus the `Custom { regen_meta: true }` visit branch, call a new `ensure_tags_linked` helper inside the same DB transaction as the media/metadata write.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p stump_core --all-targets -- -D warnings`
- [x] `cargo test -p stump_core --lib` (187 passed, including new tests)
- [x] Unit tests: ComicInfo `<Tags>` parsing (populated + empty element), HashMap→metadata comma-split path
- [x] Integration tests (in-memory SQLite, exercising `create_media` / `update_media` end-to-end):
  - [x] Scan of a book with ComicInfo tags lands them as first-class `tag` rows linked via `media_tags`
  - [x] A tag assigned through the UI survives a rescan (additive semantics verified)
  - [x] Rescanning an unchanged book is idempotent — no duplicate `tag` rows, no duplicate `media_tags` rows
  - [x] A tag shared across books is reused, not duplicated

### DIFF
--- a/core/src/filesystem/media/builder.rs
+++ b/core/src/filesystem/media/builder.rs
@@ -31,6 +31,10 @@ pub struct MediaBuilder {
 pub struct BuiltMedia {
 	pub media: media::ActiveModel,
 	pub metadata: Option<media_metadata::ActiveModel>,
+	/// Tag names extracted from the file's metadata (e.g. ComicInfo.xml `<Tags>`).
+	/// Applied additively to `media_tags` during create/update so user-assigned tags
+	/// are never removed by a rescan.
+	pub tags: Vec<String>,
 }
 
 impl BuiltMedia {
@@ -72,6 +76,7 @@ impl MediaBuilder {
 				media_id: Set(Some(media.media.id.clone())),
 				..meta
 			}),
+			tags: generated.tags,
 		})
 	}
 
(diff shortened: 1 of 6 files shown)


aaronleopold (approved):
This looks great, thank you! And I appreciate the tests

Issue #1062, opened 2026-04-06
Feedback for Metadata Fetching


Ammoton:
### 1. Make it possible to fetch the metadata again for a specific entry.
This might be a separate feature request, but I would unify the current metadata editing and the fetching for a specific entry into one menu accessible via the three-dots menu (maybe under manage). There should be an edit button that opens a view where you can edit the title, author, etc., and where you can fetch metadata with a comparison view. For the metadata fetching part, I imagine something similar to Jellyfin's identification (Request 5), where you provide a name (Request 2) or ID and then get a list of different entries. I find Audiobookshelf to be good inspiration for this. (The metadata at the bottom of the page should be displayed beside the book and only be editable through the new editing view.)

### 2. Define search name
There should be the possibility to define a search name for an entry where no match was found. Maybe there could be logic for when there is no title in the metadata to pull the title from the filename, where the user or a predefined list converts the filename to a title. For example, if the title has `v02`, it could be converted to `Volume 2` for the search, or `(Digital)` could be removed.

### 3. See entries with no metadata match
In the metadata tab in the library settings, there should be a section for entries that have no match at all.

### 4. Make filename the title if no title is present in comparison view
I found it confusing at first when there was only "new metadata" and I assumed it was correct, but it was not. I would put the filename in the title field so it’s easier to compare the entry with the match.

### 5. Identify entry with a view of many titles at once
Similar to Audiobookshelf's match or Jellyfin's identification feature where you get a list of matches and can select the right one. In the view where entries with scores that are too low are shown, you should be presented with the best match; if that is not it, you should have the option to "Identify" it via a list view of many possible matches.

PR #1114, opened 2026-04-24
:sparkles: Add validation for metadata provider token

aaronleopold:
This change does primarily two things:

1. Sets up a reusable pattern by which the provider form can issue test requests to validate the token entered is valid
2. Implements the above for Hardcover

It will also detect when you enter `Bearer ` and inform you to remove it. Kinda relates to #1062, in conversations with the author of that issue.

### DIFF
--- a/packages/browser/src/scenes/settings/server/metadataIntegrations/providers/CreateProviderDialog.tsx
+++ b/packages/browser/src/scenes/settings/server/metadataIntegrations/providers/CreateProviderDialog.tsx
@@ -96,7 +96,7 @@ export function CreateProviderDialog() {
 			</Button>
 
 			<Dialog open={isDialogOpen} onOpenChange={setIsDialogOpen}>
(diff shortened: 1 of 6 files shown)


edorgeville:
Shouldn't the Hardcover API requests be made on the backend rather than the frontend ?  
As per [Hardcover docs](https://docs.hardcover.app/api/getting-started/#important-notes-about-the-hardcover-api): 
> This should only be used from a code backend — never from a browser.  

I noticed this while attempting to setup my Hardcover API key, which resulted in graphql requests to hardcover directly from my browser (!), which failed their pre-flight CORS checks (as they should).

aaronleopold:
Yeah it's something I changed in [another branch](https://github.com/stumpapp/stump/tree/comic-vine-metadata-provider)[](https://github.com/stumpapp/stump/tree/comic-vine-metadata-provider) but haven't had the time to wrap it up yet

PR #1115, opened 2026-04-25
👷 Add OCI image labels for changelog/metadata discovery

Kernald:
This allows tools like Renovate to pick up changelogs automatically, for example.


aaronleopold (approved):
TIL how this works, thanks for adding this! It was very conveniently timed, too, since I was _just_ looking into adding renovate automation into my own homelab setup the other day

PR #1116, opened 2026-04-25

aaronleopold:


retmas-gh:
Hey.
Two thing I noticed while translating this portion:
- I think `mobileApp -> sorting -> sortField -> CREATED_AT` has wrong english translation (should be `Created at` instead of `Date Added` which is a duplication of `ADDED_AT`)
- Probably wrong pr to mention this but here goes: I don't know your source of information about this, but since you've added webtoons and manhwa (korean comics), an addition of manhua (chinese comics) would be welcome as well

aaronleopold:
> I think mobileApp -> sorting -> sortField -> CREATED_AT has wrong english translation (should be Created at instead of Date Added which is a duplication of ADDED_AT)

This was actually intentional for me, I felt that `Date Added` sounded better than `Date Created`. While it maps literally to `created_at` for Stump, more semantically it's the date which it was "added" to Stump. I kept both because they do refer to different fields (`ADDED_AT` for downloads, `CREATED_AT` for Stump-managed entities). If folks disagree I'm happy to make the swap, though.

> Probably wrong pr to mention this but here goes: I don't know your source of information about this, but since you've added webtoons and manhwa (korean comics), an addition of manhua (chinese comics) would be welcome as well

I'd be happy to chat more about it in https://github.com/stumpapp/stump/issues/1062, I think it fits there since library types were added as part of the WIP/ongoing metadata fetching effort. What exists today came to be conversationally, when asking what types people wanted to see (and also comparing to what a few other servers have as far as categories go).

This is also to say that I don't have a single, concrete source, nor am I overly knowledgeable in this area, which is why I have comments [like this](https://github.com/stumpapp/stump/blob/nightly/apps/expo/components/filter/EntityFilterMenu.tsx#L36-L44) in the code, but particularly:

> i honestly don't know the conceptual boundaries between formats but we have

So if we feel there should be more, then we can add them. Just keep in mind that the primary goal of the categorization is for metadata fetching and, eventually, more dynamic UI presentation (e.g., `Books` would show an Author -> (canonical) Series -> Books flow instead of straight to Series -> issues, etc).

PR #1117, opened 2026-04-26
🐛 (expo): Fix image ratio in reader

Arklaum:
Fixes the incorrect image ratio being used in the image reader, where last time I accidentally used the flashlist index rather than the page index

### DIFF
--- a/apps/expo/components/book/reader/image/ImageBasedReader.tsx
+++ b/apps/expo/components/book/reader/image/ImageBasedReader.tsx
@@ -15,11 +15,10 @@ import { Success } from 'react-native-turbo-image'
 
 import { TurboImage } from '~/components/image'
 import { useDisplay, usePrevious } from '~/lib/hooks'
-import { cn } from '~/lib/utils'
 import { useReaderStore } from '~/stores'
 import { useBookPreferences } from '~/stores/reader'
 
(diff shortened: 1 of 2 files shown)


apps/expo/components/book/reader/image/ImageBasedReader.tsx
```diff
@@ -321,38 +319,38 @@ const PageSet = React.memo(
 					}}
 				>
 					<View
-						className={cn('relative flex-row items-center justify-center', {
-							'gap-0 mx-auto': indexes.length > 1,
-						})}
```
aaronleopold, line 326: I think this was added a longggg time ago, and potentially even when we used a different image library. Did you confirm a double spread is still grouped together with this removed?

aaronleopold (approved):
I had one question but otherwise this looks great, thank you!!

apps/expo/components/book/reader/image/ImageBasedReader.tsx
Arklaum: Yeah it works fine without so I removed it

In fact the actual issue that prompted this bug fix happens only in double spread mode (since that's when flashlist index and pageSet index stop lining up) so I'm double sure

PR #1120, opened 2026-04-28
💄 (expo): Add blur to progress bars and smart list stats

Arklaum:
Adds blur to progress bars and smart list stats, but I couldn't get android working in the image reader footer though

I also noticed my recent changes made some of the types unhappy, so that's fixed, and with that came updating the `generatePageSets` function to be a bit more readable.


aaronleopold (approved):
I finally had a moment to pull it down and this looks great! Thank you for adding it, and thank you for the misc fixes too!

PR #1137, opened 2026-05-03
Fetch user info from OIDC provider

### LINKED ISSUE #1135: [BUG] OIDC client does not use user info endpoint, requires dangerous Authelia configuration
### Description of the Bug

Stump assumes that the ID token includes the email address of the user, and fails if it does not (https://github.com/stumpapp/stump/blob/v0.1.2/apps/server/src/config/oidc.rs#L123). It does not request the email from the user info OIDC endpoint, which would provide that information.

Authelia does not include the email address in the ID token. In the original OIDC PR, [wolfsen9 provides a workaround Authelia config to include the email address](https://github.com/stumpapp/stump/issues/246#issuecomment-3597935738) (which has made its way to the docs), but [Authelia remarks that such a configuration is dangerous](https://www.authelia.com/configuration/identity-providers/openid-connect/provider/#id_token-1):

> This option [`id_token`] is a escape hatch which should not normally be used. It allows confidential personally identifiable information to be hydrated into the ID Token which is not normally encrypted. In addition this behavior is only necessary for clients which do not actually support OpenID Connect 1.0 and indicates a significant bug with the client.
>
> This also is a common indicator that the client uses claims other than iss and sub to link users with the provider, which is a fairly significant security issue.
>
> For these reasons this option is highly discouraged and it’s recommended the client in question fixes this significant bug instead. This option is provided only on a best effort basis

### Expected Behavior

Stump should work with Authelia without any special/dangerous configuration

### Steps To Reproduce

1. Configure a normal OIDC client for Stump, ex:
```yaml
      - client_id: stump
        client_name: Stump
        client_secret: "<REDACTED>"
        scopes:
          - openid
          - email
          - profile
        redirect_uris:
          - https://stump.nas.internal/api/v2/auth/oidc/callback
```
2. Attempt to connect to it in Stump, receive the following error:

```
stump-1  |   2026-05-02T19:21:30.465694Z ERROR stump_server::routers::api::v2::oidc: Failed to exchange code for claims: OIDCMissingEmail
stump-1  |     at apps/server/src/routers/api/v2/oidc.rs:190
```

- [ ] Web App
- [ ] Mobile App
- [x] Server

#### Server Details:
  - Release Channel: latest
  - Server OS: Arch Linux
  - Using Docker: Yes
  - Server Version: 0.1.2

#### Client Details:
  - Device: Arch Linux
  - OS: Arch Linux
  - Browser: Firefox

### Additional context

Stump should call [`Client::user_info`](https://docs.rs/openidconnect/latest/openidconnect/struct.Client.html#method.user_info) to request the user information using the ID token, and receive the email from that instead.

ColonelThirtyTwo:
Fixes #1135. OIDC providers aren't required to (and, in Authelia's case, explicitly discourage) include user info in the ID token, and require fetching from the OIDC provider's user info endpoint. This patch fectches that endpoint in the exchange code flow to get the required info.

### DIFF
--- a/apps/server/src/config/oidc.rs
+++ b/apps/server/src/config/oidc.rs
@@ -4,10 +4,11 @@ use openidconnect::{
 		CoreErrorResponseType, CoreGenderClaim, CoreJsonWebKey,
 		CoreJweContentEncryptionAlgorithm, CoreProviderMetadata, CoreRevocableToken,
 		CoreRevocationErrorResponse, CoreTokenIntrospectionResponse, CoreTokenResponse,
+		CoreUserInfoClaims,
 	},
 	AuthorizationCode, Client, ClientId, ClientSecret, CsrfToken, EmptyAdditionalClaims,
-	EndpointMaybeSet, EndpointNotSet, EndpointSet, IssuerUrl, Nonce, RedirectUrl, Scope,
-	StandardErrorResponse, TokenResponse,
+	EndpointMaybeSet, EndpointNotSet, EndpointSet, IssuerUrl, Nonce, OAuth2TokenResponse,
+	RedirectUrl, Scope, StandardErrorResponse, TokenResponse,
 };
 use serde::{Deserialize, Serialize};
 use stump_core::config::OidcConfig;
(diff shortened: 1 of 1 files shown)


ColonelThirtyTwo:
I tested this with my Authelia server with the configuration linked to in the issue and it appears to work. The separate policy should no longer be needed with this change..

aaronleopold (approved):
LGTM, thanks for correcting this!

PR #1138, opened 2026-05-03
🐛(expo): Fix undefined image size

Arklaum:
Fix where opening a book and swiping forwards roughly 25-30 pages, the image sizes started being undefined, causing no image to show up

And fix the wrong ratio being used in RTL double paged mode when the left and right images have different ratios


aaronleopold (approved):
Thank you!

PR #1141, opened 2026-05-06
🐛 (expo): Fix header button colour for pre-ios 26 devices

Arklaum:
Fixes header button colour for pre-ios 26 devices so that they're easily visible in light mode. It should probably be made into a proper ios looking button rather than just the icon (like it used to be ages ago / currently does on android), but I don't feel like doing that right now.

Also rename `IS_IOS_24_PLUS` to `IS_IOS_26_PLUS`, and remove unused `ENABLE_LARGE_HEADER`


aaronleopold (approved):
Thank you!

> Also rename `IS_IOS_24_PLUS` to `IS_IOS_26_PLUS`, and remove unused `ENABLE_LARGE_HEADER`

It's so funny to me that I ever made this typo but especially that my eyes just totally missed detecting it in the months since.

PR #1146, opened 2026-05-10
♻️ Exclude active series from on deck

Arklaum:
Excludes active series, meaning a series where one of its books is currently being read, from on deck "Your Next Read"


aaronleopold (approved):
Looks good to me, thank you!

PR #1148, opened 2026-05-11

aaronleopold:


aaronleopold:
that is a lot of conflicts, i'll try to fix it during the week so these translations can flow through to nightly

PR #1152, opened 2026-05-13
🐛 Fix reading time updates for paged image reader

Arklaum:
Fixes reading time not updating the database when using the browser's paged image reader, and the reset timer button not resetting the timer properly on browser and mobile.

One thing I haven't fixed: only when changing page does the database update, so clicking `Reset Timer` doesn't push the newly reset timer until you change page. Not too much point in fixing this, and we'll have a new reading timer for the reading journal which will change things anyways


apps/expo/stores/reader.ts
aaronleopold, line 1: This is great, and moving the focus logic (which went duplicated multiple times) here is very smart. Thank you!

aaronleopold (approved):
lgtm, thank you for the fixes and convenient refactors!!

PR #1162, opened 2026-05-19
🐛 (expo): Fix Android themed icon

Kernald:
The Android monochrome icon asset was an opaque (RGB, no alpha channel) image. Android derives the icon's shape from the alpha, so a fully-opaque image was rendered as a fully opaque, plain disc/squircle/whatever theme the OS was using.

This new asset was generated from the logo.svg, keeping the linework and background arcs, in the same size/padding as before, matching the Google guidelines.

Screenshots on Android 16:
stump-themed-icon-dark
stump-themed-icon-light


aaronleopold (approved):
Thank you!!

Issue #1168, opened 2026-05-25
Add additional metadata providers: iTunes, Wikidata, Wikipedia, GCD, Metron


balazs-szucs:
> [!NOTE]
> I truncated the responses to only useful information; run the curl commands to see the full output.

More providers exist, but for now I think these (can) be considered without much problems. Well besides already mentioned providers in the docs: https://www.stumpapp.dev/docs/guides/integrations/metadata-fetching/providers

### iTunes / Apple Books

No API key required. High hit rate for books. Covers available via `artworkUrl100`. Supports multiple lookup strategies and non-English storefronts. Limit: 20 req/sec

Search by title + author

```bash
curl -s "https://itunes.apple.com/search?term=blood+of+elves+andrzej+sapkowski&entity=ebook&limit=1&country=us" \
  | python3 -m json.tool
```

```json
{
  "resultCount": 1,
  "results": [{
    "trackId": 357669161,
    "trackName": "Blood of Elves",
    "artistName": "Andrzej Sapkowski & Danusia Stok",
    "genres": ["Epic Fantasy", "Books", "Sci-Fi & Fantasy", "Fantasy"],
    "releaseDate": "2009-05-01T07:00:00Z",
    "artworkUrl100": "https://is1-ssl.mzstatic.com/.../100x100bb.jpg",
    "description": "...",
    "price": 8.99,
    "averageUserRating": 4.5,
    "userRatingCount": 1984,
    "kind": "ebook"
  }]
}
```

Lookup by ISBN-13

```bash
curl -s "https://itunes.apple.com/lookup?isbn=9780316073714&entity=ebook"
```

**Lookup by Apple Book ID**

```bash
curl -s "https://itunes.apple.com/lookup?id=357669161"
```

Full author catalog by Artist ID

```bash
curl -s "https://itunes.apple.com/lookup?id=281832452&entity=ebook&limit=200"
```

**Non-US storefronts**

```bash
curl -s "https://itunes.apple.com/search?term=ostatnie+zyczenie+sapkowski&entity=ebook&limit=5&country=pl"
curl -s "https://itunes.apple.com/search?term=der+letzte+wunsch+sapkowski&entity=ebook&limit=5&country=de"
curl -s "https://itunes.apple.com/search?term=blood+of+elves+sapkowski&entity=ebook&limit=5&country=gb"
```

---

### Wikidata

Structured bibliographic metadata via entity API. No key required.

Search by title

```bash
[...]

PR #1169, opened 2026-05-25
:sparkles: Comic vine metadata provider

aaronleopold:
Relates to https://github.com/stumpapp/stump/issues/1168

This is a heavy WIP, but I am opening a PR so I don't forget to come back to it after the few immediate items in my queue 🙈


aaronleopold:
I was really hoping to have more time to come back to this, but I haven't. So, I will merge this in and let the `nightly` folks help test. That said, I will still do a final sanity pass to ensure nothing critical fails

PR #1172, opened 2026-05-26
♻️ Add infinite scrolling for files in UploadModal to improve UX when uploading large amount of files

balazs-szucs:
This PR adds infinite scroll to the upload dialog so that users may use that to upload large amount of books in one sitting.

Previously, if there were too many books loaded there was layout overflow, that:
- was not scroll-able
- made it impossible to go forward with the upload
- loaded every book into DOM (though, not necessary problematic, but not exactly a great thing either)

### Before

*The modal expands off-screen, scrollbar is missing, the list is not scrollable at all, buttons are unreachable*
image

#### Autumn theme:

image

#### White theme:

image

#### Dark theme:
image

First PR so feel free to let me know if I forgot something. Thanks!

### DIFF
--- a/packages/browser/src/components/explorer/upload/UploadModal.tsx
+++ b/packages/browser/src/components/explorer/upload/UploadModal.tsx
@@ -48,6 +48,21 @@ export default function UploadModal() {
 	const { currentPath, refetch, uploadConfig, libraryID } = useFileExplorerContext()
 
 	const [uploadProgress, setUploadProgress] = useState(0)
+	const [visibleCount, setVisibleCount] = useState(50)
+
+	useEffect(() => {
+		setVisibleCount(50)
+	}, [files.length, uploadType])
+
+	const handleScroll = useCallback(
+		(e: React.UIEvent<HTMLDivElement>) => {
+			const target = e.currentTarget
+			if (target.scrollHeight - target.scrollTop <= target.clientHeight + 150) {
+				setVisibleCount((prev) => Math.min(prev + 50, files.length))
+			}
+		},
+		[files.length],
+	)
 
 	const config = useMemo(
 		() => ({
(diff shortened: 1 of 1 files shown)


balazs-szucs:
As for autumn theme, i think that black/dark list div does not look very good, may need to do follow up on this. But it is also just visual problem, so no big deal there.

balazs-szucs:
And fixed grammar mistakes sorry about that. 50 book, is also just a magic number, i have no particular attachment to that, it may be 100 or any reasonable number imho.

packages/browser/src/components/explorer/upload/UploadModal.tsx
```diff
@@ -48,6 +48,21 @@ export default function UploadModal() {
...
+
+	const handleScroll = useCallback(
+		(e: React.UIEvent<HTMLDivElement>) => {
+			const target = e.currentTarget
+			if (target.scrollHeight - target.scrollTop <= target.clientHeight + 150) {
+				setVisibleCount((prev) => Math.min(prev + 50, files.length))
+			}
+		},
+		[files.length],
+	)
```
aaronleopold, line 65: > [previously this modal] loaded every book into DOM (though, not necessary problematic, but not exactly a great thing either)

This is definitely an improvement, but it will still load everything into the DOM once you reach the end. If we wanted to improve this further, I'd suggest virtualizing the list so it can properly cycle things out from the DOM for us as you scroll. The web app uses both [virtuoso](https://virtuoso.dev/) and [tanstack/virtual](https://tanstack.com/virtual/latest/docs/framework/react/react-virtual) so there should be a couple of examples to pull from (e.g., `DirectoryPickerModal`).

aaronleopold (approved):
I had one suggestion if we wanted to take this a step further, but I think this is good as-is!

> As for autumn theme, i think that black/dark list div does not look very good, may need to do follow up on this. But it is also just visual problem, so no big deal there. 

I agree, I have another branch where I am working on theming for web so will make a note to give it a look there.

> And fixed grammar mistakes sorry about that. 50 book, is also just a magic number, i have no particular attachment to that

No worries about grammar! I don't have a preference, I think 50 is totally fine. I honestly didn't even consider that folks would be uploading that many at a time when I added support for this feature 😅

packages/browser/src/components/explorer/upload/UploadModal.tsx
balazs-szucs: Ah, that is true. Well, provided the user scroll down the bottom :'D

Feels like good improvement for future, i can make follow up on this.

aaronleopold: Yeah I think a follow-up is totally fine. I'll merge this as-is shortly, then. Thank you!!

PR #1174, opened 2026-05-27
⚡️ Serve precompressed frontend assets with brotli and gzip

balazs-szucs:
> **Note:** 
> If Axum has built-in on-the-fly compression I'm not aware of,  this PR's value is reduced/becomes quite useless. If it doesn't, this should yield a meaningful perf boost.

This PR improves frontend load performance by serving precompressed static assets.

Changes:

- Add vite-plugin-compression2 to generate gzip and brotli files in the web build. See here: https://www.npmjs.com/package/vite-plugin-compression2
- Use precompressed file lookup in the Axum static asset routes.
- Keep fallback to normal files when the client does not support compression.
- Add cache headers for hashed assets

TLDR:

An 80% or so reduction in transfer size (5.16 MB -> 806 kB) for the main JS bundle alone. Actual gains will vary depending on the environment e.g., which compressed assets are served or whether they are served at all (it does not force it, if the headers indicate the client cannot handle it, normal assets will be served) but it is a good indicator what we are expecting here.

Image 1 (uncompressed): index-D9RW9Hni.js transferred at 5.16 MB, no content-encoding header, Cache-Control: no-cache

Image 2 (compressed): Same file transferred at only 806.12 kB, content-encoding: br (Brotli), Cache-Control: public, max-age=31536000, immutable, no-transform

image

image


apps/web/vite.config.ts
```diff
@@ -27,6 +29,19 @@ export default defineConfig({
...
+			threshold: 1024,
+			algorithms: [
+				defineAlgorithm('gzip', { level: 9 }),
+				defineAlgorithm('brotliCompress', {
+					params: {
+						[zlibConstants.BROTLI_PARAM_QUALITY]: 11,
+					},
+				}),
+			],
+		}),
```
balazs-szucs, line 44: Copied from: https://www.npmjs.com/package/vite-plugin-compression2

balazs-szucs:
Also: GIFs are terrible for web assets. If possible then a good follow-up on this, the GIFs should be converted to WebP or similar. 

But this not a huge deal.

aaronleopold:
Excited to give this one a look! I probably won't have time today, though.

> **Note:**
> If Axum has built-in on-the-fly compression I'm not aware of, this PR's value is reduced/becomes quite useless. If it doesn't, this should yield a meaningful perf boost.

I don't believe it is built-in, but `tower_http` does have a [`CompressionLayer`](https://docs.rs/tower-http/latest/tower_http/compression/struct.CompressionLayer.html) behind a feature flag (e.g., `compression-br`) which we may want to explore, as well.

balazs-szucs:
No worries on the timeline.

Also forgot to say:

There are inherent uncertainties here (when it comes to changes made in this PR), local dev environments commonly, depending on their setup, may not serve compressed assets even when a production build would. (At least i have worked on codebases where this was the case)

I'm also not fully certain how this affects other clients, if at all. These caveats mostly stem from the fact that I've only been looking at this codebase for ~3 days, so I'm still not across every quirk.

Anyways, hoping this will work out well :)

For tower suggestion, i will need to double check if that makes sense, I am afraid purely looking from the docs, i cannot say.

aaronleopold:
I will try to come back to this today! I took some time over the last couple days here and there to review the tower stuff I found, and can share more in the review. TLDR; it isn't really applicable for us here

balazs-szucs:
> I will try to come back to this today! I took some time over the last couple days here and there to review the tower stuff I found, and can share more in the review. TLDR; it isn't really applicable for us here

I also did some digging, and came to a similar opinion, with the caveat that the most important word is "here". As far as I can tell, tower compression is not really for FE static assets BUT it is for example for GraphQL responses. so i was able to make like few LOC changes, and got good pretty good results :)


image
image


I feel like we should enable this. Unless there is some big thing I am missing, this is _very_ easy win :)

balazs-szucs:
The diff looks like this:

```diff
@@ -49,7 +49,8 @@ tower-http = { version = "0.5.2", features = [
   "fs",
   "cors",
   "set-header",
-  "trace"
+  "trace",
+  "compression-full"
 ] }
 thiserror.workspace = true
 time = "0.3.36"
diff --git a/apps/server/src/http_server.rs b/apps/server/src/http_server.rs
index 50ecce85..7fd02a73 100644
--- a/apps/server/src/http_server.rs
+++ b/apps/server/src/http_server.rs
@@ -9,7 +9,7 @@ use stump_core::{
 };
 use tokio::net::TcpListener;
 use tokio::sync::Notify;
-use tower_http::trace::TraceLayer;
+use tower_http::{trace::TraceLayer, compression::CompressionLayer};
 
 use crate::{
        config::{cors, session::get_session_layer},
@@ -72,6 +72,7 @@ pub async fn run_http_server(config: StumpConfig) -> ServerResult<()> {
                .with_state(app_state.clone())
                .layer(get_session_layer(app_state.clone()))
                .layer(cors_layer)
+               .layer(CompressionLayer::new())
                .layer(TraceLayer::new_for_http());
 
        let shutdown_notify = Arc::new(Notify::new());
```

For this amount of effort it looks like a no-brainer. The only thing i do not get... why is the compression ratio vary so much :thinking: like sometimes it's well 90% other times seem to be like less 10%... Not sure what's going on there, but hey; as I said; free perf gains.

aaronleopold (commented):
I'm not explicitly approving yet only because I know these changes will conflict with https://github.com/stumpapp/stump/pull/1185 and I'll have to give it another look, but these changes lgtm!

> As far as I can tell, tower compression is not really for FE static assets

Yeah exactly, I came to the same conclusion as well. You've already noodled with it and have the diff ready, so feel free to plop it in here too.

As for the inconsistent compression ratio, I'm not too sure myself and wouldn't want to guess

balazs-szucs:
> Yeah exactly, I came to the same conclusion as well. You've already noodled with it and have the diff ready, so feel free to plop it in here too.

Yeah, changes are quite lean. But here is the relevant commit for that: [343cc46](https://github.com/stumpapp/stump/pull/1174/commits/343cc469654efb1acf7769cbf99d2ecd73c39f11)

As far as I can tell, there is really nothing to this besides that. The defaults all... very sensible and any further "tuning" would be "reapplying" the defaults or just waste of time imho. 

> As for the inconsistent compression ratio, I'm not too sure myself and wouldn't want to guess

I think... it is due to the fact that some responses contain binary media files, and those are skipped. (Which makes sense why ZIP etc... are not compressed again, for instance.), either way, not every mystery needs to be solved here :)

aaronleopold (approved):
Thanks again!

PR #1177, opened 2026-05-27
👷 Add cooldown period and min release age for dependencies updates

balazs-szucs:
This PR adds a 3-day cooldown before Dependabot raises PRs for newly published dependency versions.

Supply chain attacks through compromised packages have been on the rise lately, TanStack being just one recent example. While there's no ultimate solution against these, a short cooldown gives security researchers and the community a window to catch and report a malicious release before we unknowingly pull it in. Three days is a small wait, but often enough for the worst offenders to get flagged.

E.g., the Tanstack comprosed packages were detected after 20 mins (according to post-mortem: 19:26:14	were the packages published, 19:46 step-security already reported the incident), so 3 days should be good enough. See: https://tanstack.com/blog/npm-supply-chain-compromise-postmortem

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,6 +10,8 @@ updates:
       - 'dependencies'
     versioning-strategy: increase
     target-branch: 'nightly'
+    cooldown:
+      default-days: 3
 
   - package-ecosystem: 'cargo'
     directory: '/'
@@ -19,6 +21,8 @@ updates:
     labels:
       - 'dependencies'
     target-branch: 'nightly'
+    cooldown:
+      default-days: 3
 
   - package-ecosystem: 'github-actions'
     directory: '/'
@@ -28,3 +32,5 @@ updates:
     labels:
       - 'dependencies'
     target-branch: 'nightly'
+    cooldown:
+      default-days: 3
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,3 @@
 strict-peer-dependencies=false
 node-linker=hoisted
+min-release-age=3


balazs-szucs:
I am quite new to this gitmoji thing, i am not quite sure this is the right one. Hope it's fine.

aaronleopold:
> I am quite new to this gitmoji thing, i am not quite sure this is the right one. Hope it's fine.

It's all good, there are times where a change doesn't quite fit cleanly within the system. 🔖 is for releases, so not that one. I'd say this falls under :construction_worker:, since it updates automation (even though it isn't CI which that one is meant for). A catch-all `chore` would be nice, but they [don't have that](https://gitmoji.dev/). Edit to add maybe :wrench: could be a decent enough catch-all

balazs-szucs:
> > I am quite new to this gitmoji thing, i am not quite sure this is the right one. Hope it's fine.
> 
> It's all good, there are times where a change doesn't quite fit cleanly within the system. 🔖 is for releases, so not that one. I'd say this falls under 👷, since it updates automation (even though it isn't CI which that one is meant for). A catch-all `chore` would be nice, but they [don't have that](https://gitmoji.dev/). Edit to add maybe 🔧 could be a decent enough catch-all

Perhaps an easy solution to this in general is to have PR template where on the bottom (or top) there is little guide of most common gitmoji stuff. That would decrease time to write PR messages because I can "fill in" instead writing the whole thing :)

However, that is quite low priority improvement. It would also help establish the baseline "due dilligence" and so on, in regards when making changes. E.g., checklist of commands, common TODOs etc...

aaronleopold:
Yeah I'm not at all opposed to adding some more guidance and structure 🙂

aaronleopold (approved):
Thank you!

PR #1178, opened 2026-05-27
🔧 Pin actions in workflow files

balazs-szucs:
While working on: #1177, I noticed that majority of CI runner were not pinned.

Generally not pinning is bad practices because tags like @v4 can be silently force-pushed to point to a completely different commit, meaning a compromised upstream maintainer can swap in malicious code without us noticing. A full SHA is immutable by design; what we reviewed is exactly what runs. This is the standard recommendation for any CI pipeline handling secrets or deployments.

A human-readable version comment is kept alongside each SHA so it stays clear what version we're actually on, and Dependabot will handle keeping them up to date automatically.


balazs-szucs:
And while double checking this: in https://github.com/balazs-szucs/stump/blob/8b0d61059fc646f883f193ced6da3de18c5770d1/.github/workflows/release_docker.yml#L17 and few others.

The secrets should be on runner scope not on workflow scope. Right now, all the runners have access to the .env variables. (I think so at least, i am also no CI wizard sadly.)

aaronleopold:
> The secrets should be on runner scope not on workflow scope. Right now, all the runners have access to the .env variables

Yeah that sounds reasonable, I think if you want while you're at it you can remove that `env` block and just provide them to the composite action directly. E.g.

```yml
with:
  username: ${{ secrets.DOCKERHUB_USERNAME }}
  password: ${{ secrets.DOCKERHUB_TOKEN }}
```

.github/actions/setup-rust/action.yml
```diff
@@ -16,7 +16,7 @@ runs:
   using: composite
   steps:
     - name: Install Rust
-      uses: actions-rs/toolchain@v1
+      uses: actions-rs/toolchain@63eb9591781c46a70274cb3ebdf190fce92702e8 # v1
```
aaronleopold: May be worth migrating to `dtolnay/rust-toolchain` down the road, I see `actions-rs/toolchain` is archived (a few years, back, too 🙈)

.github/actions/upload-artifact/action.yml
```diff
@@ -22,7 +22,7 @@ runs:
         echo "normalized_path=$UPLOAD_PATH" >> $GITHUB_OUTPUT
 
     - name: Upload
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
```
aaronleopold: We can probably bump to v6:

```suggestion
      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v6
```

aaronleopold (commented):
I validated each commit rev and they all seem to be valid, not that I don't trust you but figured it was best to check 😅 

A few of the pins can be upgraded, though. I started to note them in the review but figure while I'm at it I'll just swap em out during my lunch hour later. Thanks for getting the ball rolling, here!

.github/actions/setup-rust/action.yml
balazs-szucs: 😢 

Should be a good first issue/follow-up

.github/actions/upload-artifact/action.yml
balazs-szucs: I think so too :)

Commit: [3c655da](https://github.com/stumpapp/stump/pull/1178/commits/3c655da2d78bb3e12a3606ae720eff708bbb9ac3)

balazs-szucs:
> I validated each commit rev and they all seem to be valid, not that I don't trust you but figured it was best to check 😅


The pins reference upstream commits, so I don't think there's any realistic way to tamper with those hashes.

That said, as mentioned, this was really just a side-quest. The main concern is still #1177, I really don't want to accidentally pull in any fresh npm packages. It's more of a liability for developers than for end users, since you're unlikely to trigger a release immediately after a Dependabot merge. However, if you do local dev work after Dependabot's PR gets merged, that could potentially put your local environment in a bad state. E.g., coming back to the Tanstack thing, it was designed to steal keys, API keys etc... (If I remember correctly)

.github/actions/setup-rust/action.yml
balazs-szucs: Done: #1184

aaronleopold (approved):
Thanks, again!

Issue #1176, opened 2026-05-27
Optimize frontend assets (formats, compression, and delivery)


balazs-szucs:
This issue tracks frontend asset optimization to reduce transfer size, CPU usage, and overall load performance.

Some assets are currently served in suboptimal formats (e.g., GIF for animations, unoptimized PNG/JPEG, uncompressed fonts), which increases payload size and negatively impacts rendering speed. In some cases (notably GIF), this also leads to significantly higher CPU usage on client devices.

Standardizing formats and applying proper compression can yield immediate UX improvements, especially on slower networks and lower-end devices (e.g., e-ink readers).

The aim in general terms:

- Replace GIFs with video formats (MP4/WebM) or modern image formats where appropriate.
- Convert JPEG/PNG assets to WebP or AVIF where supported.
- Apply lossless or controlled lossy compression.
- Ensure proper caching behavior for static assets.

Expected Impact

- Reduced asset size and faster load times.
- Improved Core Web Vitals (e.g., LCP).
- Lower bandwidth usage, especially on constrained networks.
- Lower CPU usage, especially on lower end client devices (e.g., e-ink devices)

Already delivered on:
- Compression of primarily text-based frontend assets (JS, HTML, etc.) in #1174.

TODOs:

- Audit all served assets that are not optimized for web delivery (e.g., PNG, GIF, uncompressed fonts).
- Identify candidates for format conversion (GIF -> video, PNG/JPEG -> WebP/AVIF).
- Review and improve caching policies for static assets.
- Evulate Tower provided on the flight compression see this comment: https://github.com/stumpapp/stump/pull/1174#issuecomment-4555777202

Example of current issues:
<img width="3456" height="1746" alt="Example of unoptimized assets" src="https://github.com/user-attachments/assets/90c73ce6-8a64-486d-b7d7-2832a0b527c7" />

aaronleopold:
> Replace GIFs with video formats (MP4/WebM) or modern image formats where appropriate.

I wonder if we should reuse the loader from the mobile app. The existing GIF for web is something I threw together in photoshop a good bit ago, whereas the [assets in the mobile app](https://github.com/stumpapp/stump/tree/main/apps/expo/assets/splash) (used [here](https://github.com/stumpapp/stump/blob/main/apps/expo/app/_layout.tsx#L171-L180)) were commissioned from donations and a bit higher quality.

balazs-szucs:
[...]

PR #1185, opened 2026-05-28
:zap: Add service worker registration and improve location handling

balazs-szucs:
Part of: #1176

These address some minor lighthouse/IDE flagged performance problems.

Essentially: Bootstraps the web app with a service worker registration strategy, performance hints for font loading, and defensive global access patterns.

Changes:

- Preconnect + DNS-prefetch hints for `cdn.jsdelivr.net` (CDN is used to load emojis, apparently.)
- Preload hint for Inter Latin WOFF2 to reduce FOUT
- Apply `font-inter` class on `<body>` for consistent font rendering

Service Worker:

- Register `/sw.js` in production only, deferred via `requestIdleCallback`
  (with `setTimeout(_, 0)` fallback for Safari < 16.4)
- Preflight `HEAD /sw.js` before registration to avoid noisy browser errors during partial deployments; 
- Tolerate `405` from CDN hosts that reject HEAD on static assets
- Guard against non-browser environments with `isBrowser` check

Environment Safety:
- Introduce `getLocation()` helper wrapping `globalThis.location`
- Replace bare `window.*` references with `globalThis.*` throughout
- Fallback to `'localhost'` in `getDebugUrl()` when location is unavailable

Generally not the most coherent set of changes, but I think it should be fine.


apps/web/src/App.tsx
aaronleopold, line 1: I think the changes here are largely necessary, mostly in that it aims to solve a problem we shouldn't have. The web app is a browser-based SPA, and so I can't think of a reason why `window` would be undefined. 

That said, I'm okay with swapping it out for `globalThis`, since in a browser it should just be the same as `window`, but realistically neither should ever be nullish in the context which Stump is shipped and so these guards don't feel valuable to me. Always happy to discuss it if you feel strongly in the other direction, though 🙂

apps/web/src/index.html
```diff
@@ -10,7 +10,7 @@
 				).state
 				let appTheme = (zustandState.userPreferences?.appTheme || '').toLowerCase()
 				if (appTheme === 'system') {
-					const prefersDark = window.matchMedia('(prefers-color-scheme: dark)')
+					const prefersDark = globalThis.matchMedia('(prefers-color-scheme: dark)')
```
aaronleopold, line 17: Same as above, fine with the change but it realistically should not matter

apps/web/src/index.tsx
aaronleopold, line 1: I think these changes are potentially redundant, although admittedly I will need to double check the `vite-plugin-pwa` docs. My initial reaction is that this added logic does exactly what I understand `registerType: 'autoUpdate'` aims to do wrt automatically injecting the worker registration into the build

apps/web/src/index.html
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-<html>
+<html lang="en">
```
aaronleopold, line 2: This is fine, but might be good to swap it out if the selected language changes. It would probably be a one-liner in the context provider

apps/web/vite.config.ts
```diff
@@ -35,7 +35,7 @@ export default defineConfig({
 			workbox: {
 				maximumFileSizeToCacheInBytes: 5 * 1024 * 1024, // 5MB
 			},
-			outDir: '../dist/assets/',
+			outDir: '../dist',
```
aaronleopold, line 41: What was the reasoning for this one? In part I'm curious, if it is just to get the service worker registration at the root that's fine, but this introduces a bug if a corresponding change is not made to the server so it can serve it up.

You'd need something like what I have for the favicon in spa.rs, to the point where I'd say just create a helper to serve from dist (I noodled with it a bit):

```rust
// with a dedicated:
// .route("/sw.js", get(serve_sw))
// in the mount fn

// https://github.com/tokio-rs/axum/discussions/608#discussioncomment-7772294
async fn favicon(
	State(ctx): State<AppState>,
	headers: HeaderMap,
) -> APIResult<impl IntoResponse> {
	serve_dist_file(ctx, headers, "favicon.ico").await
}

async fn serve_sw(
	State(ctx): State<AppState>,
	headers: HeaderMap,
) -> APIResult<impl IntoResponse> {
	serve_dist_file(ctx, headers, "sw.js").await
}

async fn serve_dist_file(
	ctx: Arc<Ctx>,
	headers: HeaderMap,
	path: &str,
) -> APIResult<impl IntoResponse> {
	let mut req = Request::new(Body::empty());
	*req.headers_mut() = headers;
	match ServeFile::new(Path::new(&ctx.config.client_dir).join(path))
		.try_call(req)
		.await
	{
		Ok(res) => Ok(res),
		Err(e) => {
			tracing::error!(error = ?e, path, "Error serving file");
			Err(APIError::InternalServerError(e.to_string()))
		},
	}
}
```

aaronleopold (commented):
Thanks for taking the time to work on this! I had a few bits I'd like to poke at, let me know if you have any questions.

> CDN is used to load emojis, apparently

[Yep](https://github.com/stumpapp/stump/blob/main/packages/components/src/emoji/EmojiPicker.tsx#L38-L56), it is. However, I have loose plans to go the route I went with the mobile app wrt just [dumping a local json file](https://github.com/stumpapp/stump/blob/main/apps/expo/components/emoji/data.json) for it.

apps/web/src/index.tsx
balazs-szucs: This has sent me down a rabbit i was not aware of, so thanks.

I have not extensively worked neither with Vite/React. So i did not know that, I'll push shortly commit that disables that so, that is NOT double registered. The main (biggest) reason i have done this to avoid first-load performance contention it delays work until load and then idle, instead of competing with critical render during initial load. 

Generally, i was messing around with Lighthouse, and was trying all sort of thing to improve load metrics, and this had a measurable good impact :)

I can really recommend Lighthouse as it did flag lot of stuff that is in this PR, and even after this, there still quite a few metric in the red.

See:
image


I mostly got the implementation stuff from: https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/register

apps/web/src/index.html
balazs-szucs: Can do, i see there is a Crowdin and such setup. Thanks!

apps/web/vite.config.ts
balazs-szucs: The reason service worker scope. The change itself was made a bit late in the dev process, hence it may not have most wisest decision, but i think you are quite right. I'll push a commit for this.

This is how it ended up: [60da02e](https://github.com/stumpapp/stump/pull/1185/commits/60da02ea49af4155facd6e4066072656c7a5637b)

apps/web/src/index.tsx
aaronleopold: Yeah Lighthouse is great! I think the only difficult one to solve realistically is the image delivery, since it greatly depends on the thumbnails of a given user's libraries

apps/web/src/index.html
aaronleopold: I see you changed it to `""` for the default. I think `en` was actually better, it _is_ the default language until another is selected. At the same time, though, for a self-hosted service I don't know how much it meaningfully matters

apps/server/src/routers/spa.rs
```diff
@@ -38,15 +43,57 @@ async fn favicon(
 	State(ctx): State<AppState>,
 	headers: HeaderMap,
 ) -> APIResult<impl IntoResponse> {
+	let mut response = serve_dist_file(ctx, headers, "favicon.ico").await?;
+	response.headers_mut().insert(
+		header::CACHE_CONTROL,
+		HeaderValue::from_static("public, max-age=86400"),
+	);
```
aaronleopold, line 50: Nice

apps/web/src/App.tsx
aaronleopold: Coming back to this in my second review, I'd say let's revert this change per the above

apps/web/src/index.html
```diff
@@ -88,8 +92,17 @@
...
+		<link
+			rel="preload"
+			href="/assets/fonts/inter/files/inter-latin-standard-normal.woff2"
+			as="font"
+			type="font/woff2"
+			crossorigin
+		/>
 	</head>
-	<body>
+	<body class="font-inter">
```
aaronleopold, line 105: Don't think we need to address here, but realizing we could do something similar as language for the font here instead of defaulting to inter. That way if I've selected a different font it won't flash

apps/web/src/index.tsx
```diff
@@ -3,6 +3,75 @@ import { createRoot } from 'react-dom/client'
...
+		() => {
+			if ('requestIdleCallback' in globalThis) {
+				globalThis.requestIdleCallback(register)
+			} else {
+				globalThis.setTimeout(register, 0)
+			}
+		},
+		{ once: true },
+	)
+}
```
aaronleopold, line 28: Okay so I did my own learning exercise here because the specifics of this aren't my area. I do see the benefits you mentioned with this code, but what I'm trying to play around with is how to offload most of the complexity to the existing installed library so we don't have to maintain our own implementations. 

From local testing, it seems like we can achieve essentially the same with:

```ts
function registerServiceWorkerWhenIdle() {
  // note i dropped the redundant browser checks, we'll always be in browser
  if (!import.meta.env.PROD || !('serviceWorker' in navigator)) return

  // registerSW from 'virtual:pwa-register'
  // took me a min to find https://github.com/vite-pwa/vite-plugin-pwa/issues/277 and realize i needed to add the line `/// <reference types="vite-plugin-pwa/client" />` in the env.d.ts
  // also dropped manual err handling, can provide callback but just swallowing err is prob fine?
  // also also, no need to specify sw.js since the library will template it for us (sw.js should be the defualt)
  const doRegister = () => registerSW({ immediate: true })

  if (document.readyState === 'complete') {
    'requestIdleCallback' in globalThis
      ? globalThis.requestIdleCallback(doRegister)
      : globalThis.setTimeout(doRegister, 0)
    return
  }

  globalThis.addEventListener('load', () => {
    'requestIdleCallback' in globalThis
      ? globalThis.requestIdleCallback(doRegister)
      : globalThis.setTimeout(doRegister, 0)
  }, { once: true })
}
```

I think what you have works, so I won't die on this hill, but would ask if you could help confirm my findings here

apps/web/src/index.html
balazs-szucs: Changed it back to "en" :)

In: [3d7a203](https://github.com/stumpapp/stump/pull/1185/commits/3d7a203565eb82c78a4044a14050387bedc22a8d)

apps/web/src/index.tsx
balazs-szucs: > Yeah Lighthouse is great! I think the only difficult one to solve realistically is the image delivery, since it greatly depends on the thumbnails of a given user's libraries

Yes! The first item will be already I think green after the precompressed FE assets goes through. Render-blocking request, are kind of headache (in my experience) as much of it is (or at least was for me last time I tried) was a guessing game.


> since it greatly depends on the thumbnails of a given user's libraries

I agree with this, but not with the reasoning. I feel like user's thumbnails are deterministically generated... by the project, so there shouldn't be that big variance in the files themselves. Still, getting them to load at the right time, while not causing big problem is difficult. E.g., the first visible thumbnail in a grid is effectively the LCP candidate and must not be lazy-loaded, however others have too be. So, there is some rules to that and cannot be "just do X for everything" and so on.

apps/web/src/App.tsx
balazs-szucs: Apologies for not replying to this. Yes. Can do.

apps/web/src/index.tsx
aaronleopold: I guess I wasn't nuanced enough in how I meant. Yes you can generate thumbnails with Stump, but you don't _have_ to. And even if you do, you might generate nearly full-res thumbs, or maybe something silly with absolutely illegible blurry pixely tiny ones that are tiny in size and load immediately. Or maybe you change formats in some libraries, resize strategy, quality, etc. The experience will vary between both extremes, many knobs you can use to adjust the output. The main point I meant to convey was more that it is a problem _less_ within our control, but not that there isn't anything to be done about it at all.

I agree though everything you said :)

apps/web/src/index.tsx
balazs-szucs: I think so too. But why `immediate: true` ? Doesn't that invalidate the code below that?  Just purely reading it as English text, immidate should mean that the SW gets registered straight away, and the code would never progress, but the entire if/addEventListener block below it is specifically designed to delay registration until after load + idle, there seem to be some logical problems there? :thinking:

balazs-szucs: Looks like it works without the argument of immediate so i can omit that :+1:

aaronleopold: > But why `immediate: true` ? Doesn't that invalidate the code below that? 

I could be [misreading the source](https://github.com/vite-pwa/workbox-window-es/blob/main/src/Workbox.ts#L111-L113), but I don't think it matters either way. Semantically, it is not named the best imo and is a bit misleading

aaronleopold (approved):
lgtm, thank you!!

PR #1190, opened 2026-05-30
🔧 Upgrade pdfium-render to 0.9.1 and update docker build process to use ARG TARGETARCH

balazs-szucs:
Updated PDFium in both the Rust backend and Docker build setup.

Changes

- Upgraded the `pdfium-render` crate to version `0.9.1` in `core/Cargo.toml` for improved PDF rendering capabilities and compatibility.
- Updated the Docker build to download and verify newer PDFium binaries (`7763`) for both `amd64` and `arm64` architectures, using the `TARGETARCH` build argument for better cross-platform support. SHA1 checksums were also updated accordingly.
- Updated PDF page indexing to use the correct `PdfPageIndex` type, improving type safety and compatibility with the new `pdfium-render` version. 

`PdfPageIndex` and `bitmap.as_image()?` were needed for upgrade. 

See: https://github.com/ajrcarey/pdfium-render#version-history


balazs-szucs:
I am not quite certain if I am missing something or not, but as I was doing this I was getting segmentation fault, I imagine because there is no locking mechanism. (And PDFium is not thread safe.)

From what I can tell from this: https://github.com/ajrcarey/pdfium-render/blob/master/examples/axum_once_cell.rs

We should be using: `OnceCell<Mutex<Pdfium>>`  to avoid such errors. But not entirely sure.

core/Cargo.toml
```diff
@@ -44,7 +44,7 @@ tokio.workspace = true
 toml.workspace = true
 trash = "3.1.2"
 pdf = "0.9.1"
-pdfium-render = "0.8.24"
+pdfium-render = { git = "https://github.com/ajrcarey/pdfium-render.git" }
```
balazs-szucs: This will, I imagine need a bit conversation, i do not have the experience with the crate to say whether this a good idea or no. However, according their version history there is 9.2 released, but that does not seem to be on crate.io, see: https://docs.rs/crate/pdfium-render/latest

I feel like pdfium-render seems like very stable package where the author pushes a new release on every big change, but forgot to also push to crate.io (since I do not see automation for this, I imagine he is doing it manually) so it should be fine? 🤔 


Alternatively, I can just pin to the commit. What do you think?

balazs-szucs: Or we can postpone this, and I can just make issue ticket at the repo, see what the maintainer says? :thinking:

aaronleopold: Is there something we really need from their patch version which warrants the change? If not, I'd say `pdfium-render = "0.9.1"` would be preferred and can just be bumped once it is released.

balazs-szucs: > Is there something we really need from their patch version which warrants the change? If not, I'd say `pdfium-render = "0.9.1"` would be preferred and can just be bumped once it is released.

I guess not :)

Can dowgrade, 1 sec.

Cargo.lock
aaronleopold, line 1: This is a _really_ large diff for one dependency change. I may want to pull this branch down locally and test the upgrade path myself to see what's going on.

aaronleopold (commented):
The changes look good to me, but I want to double check the lock changes shortly before giving my stamp.

> I am not quite certain if I am missing something or not, but as I was doing this I was getting segmentation fault, I imagine because there is no locking mechanism. (And PDFium is not thread safe.)
> 
> From what I can tell from this: https://github.com/ajrcarey/pdfium-render/blob/master/examples/axum_once_cell.rs
> 
> We should be using: `OnceCell<Mutex<Pdfium>>` to avoid such errors. But not entirely sure.

Nice, yeah convenient they have a demo using axum already. We can definitely put it behind a once cell, I'll make a note to do that but feel free to create an issue to track it if you'd like

Cargo.lock
balazs-szucs: That may be, apologies. I did not modify it manually just ran usual commands. Not sure why it is. :/

balazs-szucs:
> Nice, yeah convenient they have a demo using axum already. We can definitely put it behind a once cell, I'll make a note to do that but feel free to create an issue to track it if you'd like

Yes, i can take a closer at this :)

Though, if it has been working so far, that it pretty good signal this may not be that widespread issue and I was just unlucky i guess or something with those segfault stuff. Still, i guess if the examples recommend it, I would assume that is "best-practice", so yeah.

Cargo.lock
aaronleopold: No worries! I just patched it. It's a much nicer diff size now 🙂

aaronleopold (approved):
Thank you!

PR #1167, opened 2026-05-25
feat: enable PSQL

sbeckstrand:
YOLO, no local testing. Just seeing if I can easily connect to a psql db instead. zero expectation of this working

PR #1191, opened 2026-05-30
:sparkles: PostgreSQL support

aaronleopold:
This PR piggybacks off of the work done by another (I became aware from https://github.com/stumpapp/stump/pull/1167, thank you @sbeckstrand for your work) to add PostgreSQL support. There are a few seemingly unrelated changes in the mix, e.g. cargo chef in the Docker build, that on the surface seem fine but I have yet to meaningfully review. I may wind up pruning to keep this a more focused change, but we'll see.

I'm opening this PR a bit early in the hopes to get better visibility and help seeing it to the finish line. There are quite a few assurances I'd want via testing suites before really seriously considering a merge, since my immediate priority is ensuring existing or future SQLite-based installs do not break.

If you have time to help validate and/or improve, please feel free to base PRs into this branch. 

You can provide a fully-resolved connection string via `STUMP_DATABASE_URL` or as separate vars:

- `STUMP_DB_PASSWORD`
- `STUMP_DB_HOST`
- `STUMP_DB_PORT`
- `STUMP_DB_NAME`
- `STUMP_DB_USER`


docs/content/docs/developer/contributing.mdx
```diff
@@ -49,6 +49,33 @@ Or `cargo` for the server:
...
+DATABASE_URL=postgresql://stump:stump@localhost:5432/stump \
+cargo run --package stump_server --bin stump_server
+```
+
+When you're done:
+
+```bash
+docker rm -f stump-postgres-dev
+```
+
```
balazs-szucs, line 78: I think it would make much more sense to have dev-docker with everything :)

E.g., make one docker compose with everything so we can just go:

```bash
docker compose -f dev.docker-compose.yml up --build

docker compose -f dev.docker-compose.yml down -v
```

aaronleopold: Yeah you can do. I don't personally use docker for local development, it's more overhead than I care for when I have a very well-defined dev environment already. The value would be for folks who don't, I guess.

Not against it, to be clear, but in a world where postgres support was merged I'd still just use a one-liner to start the container.

sbeckstrand:
I can come back to this and make some contributions next week. I was mostly just exploring if I could easily plug in a psql connection setup but found that it required some changes to existing queries. Agree that it would make sense to heavily vet and add testing as I was really in an exploration phase of seeing if I could make it work. Then I stepped away the last week and a half for vacation :sweat_smile: 

Changes to the dockerfile can be reverted entirely (related to the github workflow changes already reverted). I was making attempts to improve the image build time, but no real improvement was made there.

aaronleopold:
This will go to `nightly` once https://github.com/stumpapp/stump/pull/1315 is merged and released

aaronleopold:
I haven't encountered anymore postgres issues, doing a sanity pass on sqlite and then will merge this in

PR #1193, opened 2026-05-30
⚡ Lazy-load heavy components on FE, on-demand locales, and per-font stylesheets

balazs-szucs:
Cuts initial JS payload by ~50% (4050 kB to 2039 kB gzipped).

The app was eagerly loading a lot of weight that most users never need on boot, e.g., 32 locale JSON files, the full `@emoji-mart` dataset + renderer, and a markdown parser.

Changes are as follows:

- `MarkdownPreview` and `@emoji-mart/react` are code-split via `React.lazy()`; emoji data fetch gated on `isOpen`
- 32 locale JSONs replaced with `import.meta.glob`; only `en-US` is eager, rest load via `loadLocaleResources`
- `resolveLocale` normalizes raw browser strings (`fr`, `zh_CN`) to a valid `AllowedLocale`, replacing a silent bad cast (At least I think so? I was getting weird errors)
- EPUB reader injects only the selected font's stylesheet instead of a monolithic `fonts.css`

As for `ResizeObserver` added guard against post-teardown resize calls, it was giving also some headaches, not quite certain if I caused it somehow but 🤷‍♂️


balazs-szucs:
I made this a draft, as I think, something stinks here. I'll take second/third look at this. May be nothing. Just a gut feeling.

balazs-szucs:
So, what tripped me up the maths vs my older precompress PR, so did some digging:

| State | Old PR | This PR | Difference |
| :--- | :--- | :--- | :--- |
| Uncompressed (Main bundle) | 5.16 MB (~5,283 kB) | 2.31 MB (2,309.48 kB) | ~56% |
| Gzipped (Main bundle) | 806 kB | 713 kB | ~11.5% |

So, yeah this is more like 12% percent improvement not 50%. Still, good? I think? I guess compression does not scale as the non-compressed file goes, so lot of resource that were cut off are exactly the stuff that were well compressed, hence why does does not bigger gains. So, in other words: lazy loading them shrinks the raw file dramatically, but their contribution to the compressed size was already minimal. GZIP "hid" the cost of those resources before, so deffering them doesn't move the needle as much post-compression. Probably locales files, i think, that contained huge amount formatting/whitespaces etc, so those are very well compressed. (best guess)

Apologies for the confusion.

balazs-szucs:
> ../dist/assets/hina-mincho-japanese-400-normal-CLxaJXTw.woff2                   1,424.40 kB


That seems like a very hefty thing to be pulling in. There should be a more lightweight option, i think. 


(I am defo not suggesting dropping it, rather finding more suitable way to get the fonts.)

packages/i18n/src/config.ts
```diff
@@ -1,186 +1,32 @@
...
-	'zh-TW': {
-		'zh-TW': zhTW,
-	},
 }
 
-export type Translation = (typeof resources)['en-US']['en-US']
+const localeLoaders = import.meta.glob<{ default: Translation }>([
+	'./locales/*.json',
+	'!./locales/en-US.json',
+])
```
aaronleopold: This will likely break expo. `import.meta.glob` is a vite API, right? Not all apps in the repo are using vite, so if we want to lazily load things we will need another mechanism for doing so

packages/i18n/src/config.ts
```diff
@@ -203,4 +49,57 @@ i18n.use(initReactI18next).init({
 	resources,
 })
 
+export const resolveLocale = (inputLocale?: string): AllowedLocale => {
```
aaronleopold, line 113: I'm a little conflicted about this function, at least for some of it. I think it might be overly defensive, really. I'm of the mentality that breaking fast will be easier to catch, e.g. if someone's navigator returns a locale which we don't support I'd rather bake in support for it than not realize it isn't working as expected.

packages/i18n/src/config.ts
```diff
@@ -203,4 +49,57 @@ i18n.use(initReactI18next).init({
...
+		return fallback
+	}
+
+	if (languageCode === 'en') {
+		return 'en-US'
+	}
+
+	const languageMatch = LOCALES.find((locale) =>
+		locale.toLowerCase().startsWith(`${languageCode}-`),
+	)
```
aaronleopold: There are locales which could be matched here, but we'd only ever return the first. A bit of a guessing game (e.g. if my navigator passed just `pt` it would flow thorugh here to `pt-BR` when I might actually want `pt-PT`).

packages/i18n/src/config.ts
```diff
@@ -203,4 +49,57 @@ i18n.use(initReactI18next).init({
...
+
+	if ((LOCALES as readonly string[]).includes(inputLocale)) {
+		return inputLocale as AllowedLocale
+	}
+
+	const normalizedInput = inputLocale.replace('_', '-').toLowerCase()
+	const exactMatch = LOCALES.find((locale) => locale.toLowerCase() === normalizedInput)
+	if (exactMatch) {
+		return exactMatch
+	}
```
aaronleopold: This feels redundant, did you observe a locale which came through in this format?

packages/components/src/emoji/EmojiPicker.tsx
```diff
@@ -50,10 +52,10 @@ export default function EmojiPicker({
 			}
 		}
 
-		if (!data) {
+		if (isOpen && !data) {
```
aaronleopold, line 55: I don't see much value here, if you are on a route which has an emoji picker then IMO it should fetch the emojis _before_ you open it. I think the provides a better UX, but maybe I am missing something.

Also, though, long term we won't be fetching emojis in this way so I don't feel too strongly.

packages/browser/src/AppRouter.tsx
```diff
@@ -5,21 +5,31 @@ import { Route, Routes } from 'react-router-dom'
...
+)
+const SeriesRouter = lazy(() =>
+	import('./scenes/series').then((m) => ({ default: m.SeriesRouter })),
+)
+const SettingsRouter = lazy(() =>
+	import('./scenes/settings').then((m) => ({ default: m.SettingsRouter })),
+)
+const SmartListRouter = lazy(() =>
+	import('./scenes/smartList').then((m) => ({ default: m.SmartListRouter })),
+)
```
aaronleopold: I disagree with lazy loading routers, for two reasons really:

- These are fairly slim, even those which kick off queries. I would be worried about waterfall loading if we went this route
- The routers lazily load all of their children, already. So lazily loading a slim-enough wrapper feels like not enough gain for the UX hit IMO

packages/browser/src/scenes/smartList/settings/SmartListSettingsSideBar.tsx
```diff
@@ -7,7 +7,7 @@ import { useLocation, useNavigate } from 'react-router'
 import { usePreferences } from '@/hooks/usePreferences'
 import { formatRouteKey } from '@/hooks/useRouteGroups'
 import paths from '@/paths'
-import { SideBarLinkButton } from '@/scenes/settings'
+import SideBarLinkButton from '@/scenes/settings/SettingsSideBarLink'
```
aaronleopold, line 10: I noticed a lot of imports changing. Any particular reason?

packages/browser/src/components/ReadMore.tsx
aaronleopold, line 1: I'd prefer not to lazy load this component. If I am imagining this correctly, the first time I encounter a markdown text it would render as plain text, then once the suspend is lifted and markdown loaded it would flash back to markdown. I dislike that personally, even if it is a problem concerning primarily the initial experience.

You could maybe strip the markdown before render in the fallback, however it would still flash a change which I am generally not a fan of.

packages/i18n/src/LocaleProvider.tsx
```diff
@@ -1,34 +1,47 @@
...
+				await Promise.all([i18n.changeLanguage(resolvedLocale), initDateFnsLocale(resolvedLocale)])
+				document.documentElement.lang = resolvedLocale
+			} catch (error) {
+				console.error('Failed to load locale resources', error)
+			}
+		}
+		prepare()
+		return () => {
+			active = false
+		}
```
aaronleopold: I assume `active` is an attempt to guard loading? I don't believe it will work in the right scenarios and likely just needs a proper cancellation method. E.g. if I quickly select another language before the first loads.

packages/i18n/src/LocaleProvider.tsx
```diff
@@ -1,34 +1,47 @@
...
+import { useEffect, useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 
-import { AllowedLocale, i18n } from './config'
+import { AllowedLocale, i18n, loadLocaleResources, resolveLocale } from './config'
 import { getDefaultLocale, LocaleContext } from './context'
 import { initDateFnsLocale } from './dateFnsLocale'
 
-type Props = {
+type Props = Readonly<{
```
aaronleopold, line 8: Why readonly? Props are effectively immutable in React, this annotation feels unnecessary. Did a lint pop off somewhere to trigger this?

packages/i18n/src/vite-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="vite/client" />
```
aaronleopold: We should remove this, as previously mentioned re: not all consumers are vite apps

packages/browser/src/components/readers/epub/EpubJsReader.tsx
```diff
@@ -586,10 +646,34 @@ export default function EpubJsReader({ id, isIncognito }: EpubJsReaderProps) {
...
+				if (width <= 0 || height <= 0) {
+					continue
+				}
+
+				const renditionWithManager = rendition as Rendition & {
+					manager?: { stage?: unknown }
+				}
+				if (!renditionWithManager.manager?.stage) {
+					continue
+				}
```
aaronleopold, line 675: Curious about this one, was this something observed/corrected during runtime? I won't nitpick epubjs changes too much because I am eager to just eventually replace it with readium, but curiosity won 😄

packages/browser/src/components/markdown/MarkdownEditor.tsx
aaronleopold, line 1: Same remarks here as with the markdown preview

apps/web/vite.config.ts
```diff
@@ -16,6 +16,33 @@ export default defineConfig({
...
+					if (id.includes('node_modules')) {
+						if (id.includes('react') || id.includes('scheduler')) {
+							return 'vendor-react'
+						}
+						if (id.includes('@tanstack')) {
+							return 'vendor-tanstack'
+						}
+						if (id.includes('lucide-react')) {
+							return 'vendor-lucide'
+						}
```
aaronleopold: We don't reach this because we would have matched `id.includes('react')` above, right?

apps/web/vite.config.ts
```diff
@@ -16,6 +16,33 @@ export default defineConfig({
 		assetsDir: './assets',
 		manifest: true,
 		outDir: '../dist',
+		emptyOutDir: true,
+		rollupOptions: {
+			output: {
+				manualChunks(id) {
+					if (id.includes('node_modules')) {
+						if (id.includes('react') || id.includes('scheduler')) {
```
aaronleopold: Many packages will include `react` in the name, right?

apps/web/vite.config.ts
aaronleopold, line 1: I don't disagree with manual chunking in general, but I think maybe things might be a bit higher fidelity if we compare more complete paths instead of just substrings. E.g.

```ts
rollupOptions: {
  output: {
    manualChunks(id) {
      const path = id.replaceAll('\\', '/')
      if (!path.includes('/node_modules/')) return

      if (path.includes('/node_modules/lucide-react/')) return 'vendor-lucide'
      if (path.includes('/node_modules/@tanstack/')) return 'vendor-tanstack'
      // ... etc ...

      if (
        path.includes('/node_modules/react/') ||
        path.includes('/node_modules/react-dom/') ||
        path.includes('/node_modules/scheduler/')
      ) {
        return 'vendor-react'
      }
    },
  },
}
```

aaronleopold (commented):
Thanks for working through this! I had some comments throughout.

> So, yeah this is more like 12% percent improvement not 50%. Still, good? I think?

Yeah I think 12% is a good improvement, but based on what I saw I am worried that the improvement might be more artificial than practical (e.g., waterfall navigation from lazy loading routers).

apps/web/vite.config.ts
balazs-szucs: In practice, they are not that big :)

So, i think it should be just fine.

aaronleopold: I think my point was really in the same vein as [this](https://github.com/stumpapp/stump/pull/1193#discussion_r3330741307) above, that it could falsely match packages (like `lucide-react`)

packages/i18n/src/LocaleProvider.tsx
balazs-szucs: SonarCube complained, yes.

From what I know this is not redundant, it is a shorthand for:

```ts
type Props = {
  readonly children: React.ReactNode;
  readonly locale?: AllowedLocale;
};
```

packages/i18n/src/vite-env.d.ts
balazs-szucs: Can do, yes.

packages/i18n/src/config.ts
balazs-szucs: I'll double check, I am afraid I did not think to test with expo. So whether this is breaking for it, i would need check.

Apologies.

packages/components/src/emoji/EmojiPicker.tsx
balazs-szucs: > Also, though, long term we won't be fetching emojis in this way so I don't feel too strongly.

That would a huge update, (in terms of UX) but I am afraid I did take a look at this already, and emoji land seems to be a... Bit more complicated than I had realised, hence I think updates to this is fine. (It is probably bulky change when/if somebody gets to this.). I am not attached to this particular change.

packages/i18n/src/config.ts
balazs-szucs: That may be, i think this is fine? I think the fallback to en-US is fine, most of (if any) problems are logged well enough, at the end of the day, I do not think this should cause any harm.

On the client, I think we would want to avoid throwing/failing on principle.

For the `pt`, i think you are right, i'll push commit when I am at PC again.

packages/i18n/src/LocaleProvider.tsx
aaronleopold: Oh yeah I know what the utility does, my point was really that in React props should be treated as immutable and this change is not technically required nor actually changes runtime behavior. Does not hurt to have it, though, but it feels more of a quirk of Sonar

packages/browser/src/scenes/smartList/settings/SmartListSettingsSideBar.tsx
balazs-szucs: Tree shaking/and avoiding circular deps.

balazs-szucs:
> Yeah I think 12% is a good improvement, but based on what I saw I am worried that the improvement might be more artificial than practical (e.g., waterfall navigation from lazy loading routers).

I redid the benchmark now its 626.23 kB so -22.3%, I think some of more commits after I made my initial comment improved it significantly. 

In practice, I feel like this should _be_ an improvement, BUT a "hard to measure" improvement though. Lot of the chunking improvement will manifest themselves with caching/etc improvement, that is not necessarily easy to get a good measurement on.

Anyways, last 8 commit, I think addressed everything.

packages/i18n/src/LocaleProvider.tsx
aaronleopold: I think this bit still needs to be addressed, and also will change a bit with updates from `nightly` which fixed things for expo

aaronleopold (commented):
I think only one of my comments went missed, but otherwise I think this looks good 🙂

packages/browser/src/components/readers/epub/EpubJsReader.tsx
balazs-szucs: Yes, I was getting an error on unmounting, and since the EPUB reader component is now frequently mounted and unmounted, I think this was a pre-existing problem (meaning it would error on unmounting sometimes), but this branch would have made those errors considerably worse/more frequent. Generally, I think this is more cosmetic in the sense that I do not believe the errors I've seen would have caused any real headaches. But still, better not to pollute the console logs. Also, AFAIK this is kind of a common problem with EPUB.js, at least whenever I worked on a project that used it, there were always problems with unmounting/teardown.

According to a quick search, these also tend to occur more often with EPUBs that are NOT "correctly formatted" (whatever that means, thanks DuckDuckGo assistant). Either way, here's hoping for a speedy move to Readium 🙏 (I have not started that yet, sorry 😭)

packages/i18n/src/LocaleProvider.tsx
balazs-szucs: > I think this bit still needs to be addressed, and also will change a bit with updates from `nightly` which fixed things for expo

Sorry for not getting to this earlier! :(

I think my (not so) elegant solution should work, but I'm not 100% sure this is "the way". Relevant commit: [84ec991](https://github.com/stumpapp/stump/pull/1193/commits/84ec99191cad59bbff8e1faba73a6ab7683b6666). It also had a bunch of conflicts, but anyway 😄

(I'm open to suggestions on how to do this better or in a less verbose way. It feels like it's not even JS with the amount of error handling 😅)

aaronleopold:
Hey @balazs-szucs, coming back to this and seeing it has gone a bit stale. Is this an effort you would like to revive? Or should I go ahead and close this one out?

balazs-szucs:
> Hey @balazs-szucs, coming back to this and seeing it has gone a bit stale. Is this an effort you would like to revive? Or should I go ahead and close this one out?

I can fix, no worries. Apologies for not getting to this. Burnout and all. Obviously, no pressure eitherway.

balazs-szucs:
Okey, i think some changes upstream make stuff here bit obselete (e.g., reader changes). Still, i think majority of the stuff here is mostly unaffected, so no big worries here. Most of the conflicts look easy enough to resolve, so yeah all good.

aaronleopold:
> Apologies for not getting to this

Never any problems in this regard on my side, life is always first you're good

> i think some changes upstream make stuff here bit obselete

That's okay too, if it's not worth reviving it for now that's totally okay by me

balazs-szucs:
> That's okay too, if it's not worth reviving it for now that's totally okay by me

I can quickly re-measure, i think. (Probably tmrw, as it's midnight)

But for me this looks good now 👍 

I quickly retested/clicked around for a while it does looks good. Locale changes work, EPUB reader works, so no worries there i think.

Also 150 LOC leaner so that's also nice.

aaronleopold:
Thank you! I'll resolve the conflicts and aim to get this merged in sometime during the week

PR #1204, opened 2026-06-01
Work towards removing `is_server_owner`

Kernald:
Another batch of work towards removing `is_server_owner`. I'm not using some of the features here (especially book clubs), so would be keen on getting some feedback on whether the permission sets make sense here!

### DIFF
--- a/crates/graphql/src/mutation/user.rs
+++ b/crates/graphql/src/mutation/user.rs
@@ -368,11 +368,6 @@ impl UserMutation {
 			.await?
 			.ok_or("User not found")?;
 
-		// TODO(permissions): remove this? if we rm is_server_owner then i guess admins can delete each other?
-		if existing_user.is_server_owner {
-			return Err("You cannot delete the server owner".into());
-		}
-
 		let deleted_user = if hard_delete {
 			user::Entity::delete_by_id(id.to_string())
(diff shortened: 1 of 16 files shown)


crates/graphql/src/mutation/user.rs
```diff
@@ -600,14 +595,10 @@ async fn update_user(
 
 	let txn = conn.begin().await?;
 
-	// TODO(permissions): ensure to revisit this
-	let is_updating_server_owner = by_user.is_server_owner && by_user.id == for_user_id;
-	if !is_updating_server_owner {
-		update_user_age_restriction(&for_user_id, &input.age_restriction, &txn).await?;
+	update_user_age_restriction(&for_user_id, &input.age_restriction, &txn).await?;
```
aaronleopold, line 598: I suppose it's possible for a user to have age restrictions and manage server permissions? I don't personally use age restriction features, and esp not for those I grant admin access lol, so perhaps a non-issue for people who do use it

aaronleopold (approved):
This looks great to me! 

> I'm not using some of the features here (especially book clubs), so would be keen on getting some feedback on whether the permission sets make sense here!

I think what you have is perfectly reasonable 🙂

PR #1209, opened 2026-06-02
:recycle: Refactor PDF processing code to align with PDFium thread-safety and rendering best practices

balazs-szucs:
- Resolves concurrent rendering crashes by serializing PDFium FFI calls with a global mutex lock.
- Moves blocking page analysis tasks to a dedicated thread pool to prevent problems due to newly added Locks.
- Improves low-resolution rendering by using configured DPI settings and preserving aspect ratios.
- Refactors PDF stuff to use PDFium-render, and removes now (unused) pdf.rs library.

On the last point, generally, PDFIum should be much more reliable/performant than from what I can moreso a hobbyist Rust PDF library. So, using PDF.rs then to fallback PDFium, that seemed like bit of strange choice, and when I flipped them, i thought, there is very unlikely that PDFium would fail and PDF.rs would catch it, so at that point better to just remove. PDF.rs was also pulling in seemingly quite large of other deps. So, I think that's a win.

Date parsing still hacky. But it is mostly in working condition.

I am still not happy with date stuff, but otherwise this should be improvements. It's able parse in the bit more metadata (e.g., Summary/Description that was previously ignored), i have not been able to trigger segfault, but yeah bit slower overall (not a huge surprise.)


core/src/filesystem/media/format/pdf.rs
```diff
@@ -35,34 +40,18 @@ impl FileProcessor for PdfProcessor {
...
-				Err(e) => {
-					tracing::debug!(error = ?e, path, "Failed to digest PDF file");
-					None
-				},
-			}
-		} else {
-			None
-		}
+		let sample = PdfProcessor::get_sample_size(path).ok()?;
+		hash::generate(path, sample).ok()
```
aaronleopold, line 49: This does change behavior in that now the sample size error would bubble up. I think that's fine, but mostly noting it

core/src/filesystem/media/format/pdf.rs
```diff
@@ -87,26 +76,28 @@ impl FileProcessor for PdfProcessor {
 	}
 
 	fn process_metadata(path: &str) -> Result<Option<ProcessedMediaMetadata>, FileError> {
-		let file = FileOptions::cached()
-			.parse_options(ParseOptions::tolerant())
-			.open(path)?;
-
-		Ok(file.trailer.info_dict.map(ProcessedMediaMetadata::from))
+		let _lock = PDFIUM_LOCK.lock().unwrap_or_else(|e| e.into_inner());
```
aaronleopold: There are multiple instances of this in the file. The `.unwrap_or_else(|e| e.into_inner())` basically swallows that error and I assume fails to lock and therefore prevents the mechanism from doing its job.

Would it be worth doing something like:

```rust
fn aquire_pdfium_lock() -> Result<MutexGuard<'static, ()>, FileError> {
	PDFIUM_LOCK.lock().map_err(|e| {
		FileError::PdfProcessingError(format!("Failed to acquire PDFium lock: {}", e))
	})
}
```

And then bubbling up the error if a lock cannot be made?

core/src/filesystem/media/format/pdf.rs
```diff
@@ -26,6 +28,9 @@ use crate::{
 	},
 };
 
+static PDFIUM: OnceLock<Result<Pdfium, FileError>> = OnceLock::new();
+static PDFIUM_LOCK: Mutex<()> = Mutex::new(());
```
aaronleopold: I'm curious if we even really need to do this? It seems like even in the [canonical example](https://github.com/ajrcarey/pdfium-render/blob/master/examples/thread_safe.rs) for thread-safe usage it isn't required.

core/Cargo.toml
```diff
@@ -43,8 +43,7 @@ stump-config-gen = { path = "../crates/macros/stump-config-gen"}
 tokio.workspace = true
 toml.workspace = true
 trash = "3.1.2"
-pdf = "0.9.1"
-pdfium-render = "0.9.1"
+pdfium-render = { version = "0.9.1", features = ["thread_safe"] }
```
aaronleopold: We don't actually need to specify, [this is a default feature](https://github.com/ajrcarey/pdfium-render/blob/master/Cargo.toml#L71) and we have not opted out of default features

core/src/filesystem/media/format/pdf.rs
```diff
@@ -746,41 +744,34 @@ impl FileConverter for PdfProcessor {
...
-			if let Err(err) = std::fs::write(image_path, image_buf) {
-				tracing::error!(error = ?err, "Failed to write image to file");
-			}
+			let image_path = unpacked_path.join(format!(
+				"{}_{:03}.{}",
+				file_stem,
+				idx + 1,
+				output_extension
+			));
+			std::fs::write(image_path, image_buf)?;
```
aaronleopold: There is another behavioral change here, and the original comment was also removed:

> // NOTE: This isn't bubbling up because I don't think at this point it should
> // kill the whole conversion process.

Do you disagree? If I have a 300-page PDF and go to convert it, and it failed on page 100, I'm not sure I'd want the entire thing to fail. At the same time, I will almost certainly never use this myself 😅 so am happy deferring to those who would.

core/src/filesystem/media/format/pdf.rs
```diff
@@ -155,40 +147,137 @@ impl FileProcessor for PdfProcessor {
...
+						.unwrap_or_default(),
+				];
+				for p in local_paths {
+					if p.exists() {
+						tracing::info!(path = ?p, "Found local PDFium library");
+						path = Some(p.to_string_lossy().to_string());
+						break;
+					}
+				}
+			}
```
aaronleopold: Is this duplicating what `bind_to_system_library` does? I have not yet checked, however it feels like it would fall within that domain

aaronleopold (commented):
Thanks for starting this! I had a few comments to sort out but nothing major

core/src/filesystem/media/format/pdf.rs
balazs-szucs: The way PDFs work, corruption and invalid objects tend to occur at the document-wide (global) level rather than at the page level. It would be quite unusual for a PDF to have a completely valid structure and then contain a bunch of broken page-level objects severe enough to make PDFium fail.

I think (on a second thought) it's better to avoid failing hard in any case, so I'll push a commit for this.

Thanks!

MB: I got tripped up this message is a response to: https://github.com/stumpapp/stump/pull/1209#discussion_r3359585151 (I was reading stuff from discord, and when saw this message I though it was the other I read on Discord :rofl: )

apologies.

core/src/filesystem/media/format/pdf.rs
balazs-szucs: I am afraid so. We are spinning up multiple seperate PDFium singletons, hence our usage vs example it somewhat different. 

I could not make it work by completely following the example. (THough that may be skill issue :man_shrugging: )

core/Cargo.toml
balazs-szucs: Sounds good :)

core/src/filesystem/media/format/pdf.rs
balazs-szucs: It was needed for me, because I am working with (by my count) 3 different versions of PDFIum. That `bind_to_system_library` looks for the system level PDFium, the helper looks PDFium at that specific directory, hence it does interfere with people stuff who are working with more version :)

core/src/filesystem/media/format/pdf.rs
balazs-szucs: > There are multiple instances of this in the file. The `.unwrap_or_else(|e| e.into_inner())` basically swallows that error and I assume fails to lock and therefore prevents the mechanism from doing its job.
> 
> Would it be worth doing something like:
> 
> ```rust
> fn aquire_pdfium_lock() -> Result<MutexGuard<'static, ()>, FileError> {
> 	PDFIUM_LOCK.lock().map_err(|e| {
> 		FileError::PdfProcessingError(format!("Failed to acquire PDFium lock: {}", e))
> 	})
> }
> ```
> 
> And then bubbling up the error if a lock cannot be made?

Yeah, i needed to double check, but this seems good/reasonable. Thank you!!

core/src/filesystem/media/format/pdf.rs
aaronleopold: > We are spinning up multiple seperate PDFium singletons, hence our usage vs example it somewhat different

That is what they are doing in the example, though, right? Granted, they are not using singletons (you've introduced that pattern in this PR) however in the example they:

- Kick off a [`render` call in parallel threads](https://github.com/ajrcarey/pdfium-render/blob/master/examples/thread_safe.rs#L55-L56)
- Create a [new pdfium instance](https://github.com/ajrcarey/pdfium-render/blob/master/examples/thread_safe.rs#L67) inside that `render` fn

This is effectively Stump's model. [This explanation](https://github.com/ajrcarey/pdfium-render/blob/master/examples/thread_safe.rs#L10-L13) also reads to me that internally they are doing this for us. I mostly just want to make sure before accepting this that we've done the due diligence to make sure that I am not taking on maintenance burden unnecessarily

core/src/filesystem/media/format/pdf.rs
aaronleopold: Sort of related to my above remark about taking on more maintenance burden, this would not be considered standard usage then, right? If juggling multiple instances of pdfium, why not just swap the path provided to Stump as needed? Especially since this is for dev only. Maybe I am misunderstanding?

core/src/filesystem/media/format/pdf.rs
aaronleopold: Pulling in your reply from the other block since I resolved it and don't want it to get lost:

> The way PDFs work, corruption and invalid objects tend to occur at the document-wide (global) level rather than at the page level. It would be quite unusual for a PDF to have a completely valid structure and then contain a bunch of broken page-level objects severe enough to make PDFium fail.
>
> I think (on a second thought) it's better to avoid failing hard in any case, so I'll push a commit for this.
>
> Thanks!

core/src/filesystem/media/format/pdf.rs
balazs-szucs: > > We are spinning up multiple seperate PDFium singletons, hence our usage vs example it somewhat different
> 
> That is what they are doing in the example, though, right? Granted, they are not using singletons (you've introduced that pattern in this PR) however in the example they:
> 
>     * Kick off a [`render` call in parallel threads](https://github.com/ajrcarey/pdfium-render/blob/master/examples/thread_safe.rs#L55-L56)
> 
>     * Create a [new pdfium instance](https://github.com/ajrcarey/pdfium-render/blob/master/examples/thread_safe.rs#L67) inside that `render` fn
> 
> 
> This is effectively Stump's model. [This explanation](https://github.com/ajrcarey/pdfium-render/blob/master/examples/thread_safe.rs#L10-L13) also reads to me that internally they are doing this for us. I mostly just want to make sure before accepting this that we've done the due diligence to make sure that I am not taking on maintenance burden unnecessarily

Yeah understood :saluting_face: 

I'll try to get back to this.

core/src/filesystem/media/format/pdf.rs
balazs-szucs: Oh... that already exists: https://github.com/stumpapp/stump/blob/70c7d00f738ff6f228a282316916f3b1900ac61a/core/src/config/stump_config.rs#L175

MB. I'll revert this, apologies, i have not realised that there is already better solution :(

Thanks!

aaronleopold:
I haven't had much Stump time this week unfortunately, but I've pushed a few changes based on our discussions and will aim to give a final review by the weekend to get this into nightly. Thanks again for getting the ball rolling here, and sorry for how delayed getting it in has been!

core/src/filesystem/media/format/pdf.rs
aaronleopold: I removed it for now, if it becomes a problem we know where to look and how to fix 🙂

aaronleopold (approved):

PR #1211, opened 2026-06-03
:zap: Improve scanner efficiency and fix exploding PIDs

aaronleopold:
This PR addresses some performance issues:

- The default `200` concurrency limit was too high for scanner workloads. Not overly noticeable in my typical local testing environment, but easily overwhelming for larger ones
   - Instead I've swapped to a multiplier system with a somewhat conservative default derived from the number of logical cpus available and scaled by `STUMP_PARALLELISM_MULTIPLIER` (which defaults to `2`)
- Remove rayon from the scanner, no longer mixing runtimes and spinning up extra workers for each
- Refactor tag queries during the scanner loop to remove duplicate trips via a cache at the start. Previously every book would query the database for whether its tags existed before linkage, which at scale ballooned
- Speed up subsequent scans by reducing work via explicit tracking of visited directory modified times in a new table
- Refactor the `book_book` section of the scanner lifecycle to be more of a fill/drain pattern as to not needlessly spawn and poll a bunch of futures
- Reintroduce transactions for media/metadata/tag database operations. I previously removed it because of bind limits, but I've since added some utilities to avoid those errors and so felt we could re-add it to cut on the trips
- Extend preloaded state to reduce necessary queries throughout lifecycle of scanning
- Tune the SQLite connection string to improve busy timeouts during heavy scans

Based on my own testing, with the help of some folks with very large libraries, these changes will be most noticeable for libraries >50k wrt both time and resource utilization.


aaronleopold:
I haven't had any testing feedback here, it exists in `unstable` if you can try it, so will likely just merge it in sometime this week and hope the `nightly` folks shake out any issues

Edit: I've tested myself some more, and a couple others reached out with good findings, so will keep on working through it.

Issue #1151, opened 2026-05-13
[FEATURE] Better access control for libraries


radium213:
### Is your feature request related to a problem?

Presently all libraries have in terms of access control is a field to exclude specific users. This means all libraries are open by default to every new user created on the system. This, combined with the `STUMP_OIDC_ALLOW_REGISTRATION` flag creates a scenario where a new user can register themselves through OIDC and have immediate read access to all libraries, thus defeating the whole point of access control.

### Describe the solution you'd like

Change the exclude-list to an allow-list. Change the default library behavior to either:

- If the allow-list is empty, allow all users. If the allow-list contains any users, allow only those users.
- Deny all by default. If the allow-list contains a special value (ex.: `*`) then allow all users. Otherwise, allow only listed users.

I would prefer the second option, as it has similar UX to group based ACLs if there is ever a plan to support that.

### Where would this feature be implemented?

- [x] Web App
- [ ] Mobile App
- [x] Server
- [x] Documentation

PR #1214, opened 2026-06-04
Migrate off is_server_owner in CLI

Kernald:
This should cover the CLI aspect entirely (except for test fixtures depending on the `is_server_owner` field existing, I'll clean them when removing the field).

This is what's remaining, off the top of my head:
 - Library excluded users
 - GraphQL schema update, removing `isServerOwner`
 - Actual DB migration, granting `ManageServer` to anyone who currently has `is_server_owner = true`
 - Bootstrap logic, granting `ManageServer` rather than `is_server_owner` to the first user
 - Clean up all test fixtures that still set/read `is_server_owner`
 - Pass on the frontend to update potential wording inconsistencies
 - Docs

### DIFF
--- a/crates/models/src/shared/permission_set.rs
+++ b/crates/models/src/shared/permission_set.rs
@@ -22,18 +22,9 @@ impl PermissionSet {
 	/// Returns true iff the target permission would be effectively granted to a user
 	/// holding this set — directly granted, or reachable transitively via associations.
 	pub fn contains(&self, target: UserPermission) -> bool {
-		let mut seen: Vec<UserPermission> = Vec::new();
-		let mut queue: Vec<UserPermission> = self.0.clone();
-		while let Some(permission) = queue.pop() {
-			if permission == target {
-				return true;
-			}
-			if !seen.contains(&permission) {
-				seen.push(permission);
-				queue.extend(permission.associated());
-			}
-		}
-		false
+		PermissionSet::new(self.0.clone())
+			.resolve_into_vec()
+			.contains(&target)
 	}
 
 	/// Returns a new PermissionSet with the target permission added to the explicit
--- a/crates/models/src/shared/permission_set.rs
+++ b/crates/models/src/shared/permission_set.rs
@@ -36,6 +36,16 @@ impl PermissionSet {
 		false
 	}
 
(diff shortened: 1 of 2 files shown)


crates/models/src/shared/permission_set.rs
```diff
@@ -19,6 +19,43 @@ impl PermissionSet {
...
+		let mut queue: Vec<UserPermission> = self.0.clone();
+		while let Some(permission) = queue.pop() {
+			if permission == target {
+				return true;
+			}
+			if !seen.contains(&permission) {
+				seen.push(permission);
+				queue.extend(permission.associated());
+			}
+		}
```
aaronleopold: Any reason not to just invoke `resolve_into_vec` and run a `contains` on that? Is it just an optimization to cut an iterative loop?

crates/models/src/shared/permission_set.rs
```diff
@@ -19,6 +19,43 @@ impl PermissionSet {
 		PermissionSet(permissions)
 	}
 
+	/// Returns true iff the target permission would be effectively granted to a user
```
aaronleopold, line 22: Thought `iff` was a typo at first 😅 been a long while since I've seen it

aaronleopold (approved):
lgtm! Thank you!

> This is what's remaining, off the top of my head:
> - Library excluded users

This will eventual invert into an [allow-list](https://github.com/stumpapp/stump/issues/1151). Not sure if we want to tackle it as part of this effort, but I'm not strictly opposed to it if easier to do all at once

Otherwise, the rest of the items align with what I have in my head!

crates/models/src/shared/permission_set.rs
Kernald: None besides me being a bit too tired when I wrote this :-D Updated to use `resolve_into_vec`

crates/models/src/shared/permission_set.rs
Kernald: Happy to change it if that's not obvious enough!

aaronleopold: Nah it's totally fine! Just figured I'd share, wasn't implying a preference or anything against

PR #1216, opened 2026-06-06
:bug: Fix finished session dedupe logic

aaronleopold:
This PR changes the behavior of finished session deduplication:

- The control mechanism is now exclusively through the grace window user preference, rather than a config var specific to only deduplication
- If you are within the grace window after finishing, e.g. you got back 10 pages after reading the end, it will re-open the session

I still have a little testing to do but will merge after

### DIFF
--- a/apps/server/src/routers/koreader/sync.rs
+++ b/apps/server/src/routers/koreader/sync.rs
@@ -304,14 +304,8 @@ async fn put_progress(
 		},
 	}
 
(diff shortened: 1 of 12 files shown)


aaronleopold:
I did some light testing over the past few days and think things are looking good, but don't have much more time so will merge and let the `nightly` folks carry the testing torch 🙂

PR #1218, opened 2026-06-09
⬆️ Bump reqwest from 0.12.23 to 0.13.4

### LINKED ISSUE #1136: [FEATURE] Use native-tls instead of rustls, to support self-signed certs

Stump uses rustls, which bundles its own CA roots and makes it impossible to use an OIDC or email servers with self-signed or nonstandard CA certificates.

Switching from the `rustls-tls` to the `native-tls` feature for reqwests, and the `tokio1-rustls-tls` to the `tokio1-native-tls` feature for lettre, seemed to be enough for Stump to connect to my OIDC server.

reqwests 0.12 has the `rustls-tls-native-roots` feature but it seems to be gone in reqwests 0.13.

- [ ] Web App
- [ ] Mobile App
- [x] Server
- [ ] Documentation

### Additional context

_No response_

romanZupancic:
Other dependency changes:
- (Cargo.toml) `reqwest-middleware` 0.4.0 -> 0.5.2
- (Cargo.toml) `reqwest-retry` 0.7.0 -> 0.9.1
- (apps/server/Cargo.toml) `openidconnect` disable default features
- (apps/server/Cargo.toml) `oauth2-reqwest` add dependency

## Why?
Before this commit, any https sessions initiated by stump would use an internal trust store. So, certificates of services signed by a private root will get rejected as being signed by an unknown authority. 

Reqwest crate version 0.13.4 uses the system trust store by default. This allows stump to communicate and trust certificates signed by roots that are trusted by the host it is running on. 

Very specifically, I would like stump to communicate with an OIDC provider that is behind a certificate signed by a private authority. The changes in the PR facilitate that!

Also, #1136 asks for the exact same thing!

## How?
Unfortunately a few things depend on reqwest having a consistent version. Namely, stump server's `openidconnect` crate dependency refused to work without some shenanigans. The shenanigans in question: `openidconnect` has a dependency on the `oauth` crate, which has its own dependency on `reqwest` version 0.12.x. Changing stumps' version of `reqwest` causes is consumption of `openidconnect` to break (because, I think, stump creates a 0.13.x request http client and passes that to `openidconnect` which tries to use it like a 0.12.x http client).

This Issue in the `oauth` crate's repo goes over the current solution: https://github.com/ramosbugs/oauth2-rs/issues/333

Basically, we stop `openidconnect` (and `oauth`) from pulling in its own version of `reqwest` by disabling all its features. Then, we shim the `reqwest` object with a wrapper provided by the `oauth2-reqwest` crate. The maintainer of `oauth2` seems to be committed to maintaining this new `oauth2-reqwest` crate.

## Problems?
- `oauth2-reqwest` is technically in an alpha version. I assume it'll be promoted to an actual release soon.

Let me know if I explained things poorly!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,7 +53,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
 
 [[package]]
@@ -569,6 +569,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.13.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c953fe1ba023e6b7730c0d4b031d06f267f23a46167dcbd40316644b10a17ba"
+dependencies = [
+ "aws-lc-sys",
+ "zeroize",
+]
+
+[[package]]
+name = "aws-lc-sys"
+version = "0.30.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dbfd150b5dbdb988bcc8fb1fe787eb6b7ee6180ca24da683b61ea5405f3d43ff"
+dependencies = [
+ "bindgen",
+ "cc",
+ "cmake",
+ "dunce",
+ "fs_extra",
+]
+
+[[package]]
 name = "axum"
 version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


aaronleopold (commented):
Hey! 👋 

On the surface these changes look okay to me, but I'll want to spend a little more time eyeballing the `Cargo.lock` changes a bit more closely. I will aim to give this another look sometime over the next few days (a few other PRs to round out / look at in the queue)

> This Issue in the oauth crate's repo goes over the current solution: https://github.com/ramosbugs/oauth2-rs/issues/333

I think I'll also want to give this a proper read, as well, to make sure I understand. Thank you for linking and breaking it down!

> Let me know if I explained things poorly!

Nope! Your explanations were great, much appreciated

romanZupancic:
Sorry! I thought that `Cargo.lock` looked suspicious but didn't double check it.

I just regenerated it by grabbed the one from nightly, running a `cargo tree` on it so it picks up the reqwest changes, and now it looks a bit more reasonable.

Also, thanks for your time! (And take your time, I'm dogfooding this on my own right now anyways!)

aaronleopold (approved):
lgtm! I've subscribed to the linked issue (thank you, again) so if the situation changes I can take it as it comes

PR #1219, opened 2026-06-10
add PKCE support and remove duplicate openid scope

gzsombor:
I wanted to use Stump with Rauthy, which is a bit stricter than other SSO solutions, so it requires PKCE support and complain if a scope is duplicated.

- Remove openid from REQUIRED_SCOPES; the openidconnect crate already adds it automatically on authorize_url(), causing duplicate scopes
- Add PKCE (S256) support for authorization code flow, required by providers like Rauthy that enforce RFC 7636
- Pass PKCE code verifier through the OAuth state parameter instead of the session store, which rejects records without a user_id

### DIFF
--- a/apps/server/src/routers/api/v2/oidc.rs
+++ b/apps/server/src/routers/api/v2/oidc.rs
@@ -81,7 +81,7 @@ pub struct AuthorizeQuery {
 
 /// The OIDC state parameter passed through the authorization flow.
 /// Contains both the original query params and the PKCE code verifier.
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, Default)]
 struct OidcState {
 	#[serde(flatten)]
 	query: AuthorizeQuery,
@@ -119,6 +119,7 @@ async fn authorize(
 
 	let (pkce_challenge, pkce_verifier) = PkceCodeChallenge::new_random_sha256();
 
+	let generate_token = query.generate_token;
 	let state = OidcState {
 		query,
 		pkce_verifier: pkce_verifier.secret().to_string(),
@@ -129,6 +130,7 @@ async fn authorize(
 		APIError::InternalServerError("Failed to encode state".to_string())
 	})?;
 
+	let pkce_challenge_code = pkce_challenge.as_str().to_owned();
 	let redirect_to = get_oidc_authorize_url(
 		&client,
 		&oidc_config.get_scopes(),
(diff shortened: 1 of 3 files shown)


apps/server/src/routers/api/v2/oidc.rs
```diff
@@ -106,16 +117,27 @@ async fn authorize(
...
+	let redirect_to = get_oidc_authorize_url(
+		&client,
+		&oidc_config.get_scopes(),
+		&state_value,
+		Some(pkce_challenge),
+	);
 
 	tracing::debug!(
-		generate_token = %query.generate_token,
+		pkce = true,
```
aaronleopold: Any particular reason for this removal? Also, a hard-coded `pkce` in the span doesn't feel as useful to me

apps/server/src/routers/api/v2/oidc.rs
```diff
@@ -144,10 +166,13 @@ impl axum::response::IntoResponse for OidcCallbackResponse {
...
-fn parse_state(state: Option<&str>) -> AuthorizeQuery {
+fn parse_state(state: Option<&str>) -> OidcState {
 	state
-		.and_then(|s| serde_json::from_str::<AuthorizeQuery>(s).ok())
-		.unwrap_or_default()
+		.and_then(|s| serde_json::from_str::<OidcState>(s).ok())
+		.unwrap_or_else(|| OidcState {
+			query: AuthorizeQuery::default(),
+			pkce_verifier: String::new(),
+		})
```
aaronleopold: We could derive `Default` for `OidcState` and avoid the manual construction, both fields here are effectively defaults already

aaronleopold (commented):
I had a couple of very minor nitpicks, but I think once addressed this is ready for merge. I can try to push them up over the next few days, if you prefer to leave it here.

Either way, thank you!

apps/server/src/routers/api/v2/oidc.rs
gzsombor: No, I just wanted to have a small change to see that something is really different, but yes, the actual verification code would be better to debug.

apps/server/src/routers/api/v2/oidc.rs
gzsombor: Good idea, I've overlooked that :)

aaronleopold (approved):
Thank you!

aaronleopold:
Note for posterity: I did not catch that this was based into main. I cherry picked to `nightly` and removed it from main to let it flow through the normal release process.

PR #1220, opened 2026-06-10
Add configuration option to accept self-signed/expired certificate

gzsombor:
I'm trying to setup stump in an internal LAN, with a SSO using a self-signed certificate, unfortunately the OIDC client checks the validity, with this new flag this can be disabled.


aaronleopold:
Would https://github.com/stumpapp/stump/pull/1218 also solve your use case here? It seems to overlap, at least on the surface

gzsombor:
Thats a good question, but I don't think that would fix my problem, as I don't think that a process running in a docker container could access any system certificate. Either disabling the certificate-checking by the client, or an explicit way to add a self-created certificate could help. I will look into the second option as well, when I will have time.

romanZupancic:
@gzsombor 

On Debian based distros, `/etc/ssl/certs/ca-certificates.crt` contains many (all?) global roots.

I've just appended my private CA root cert to (a copy of) this file and mounted it into the stump container at that same path. I'm pretty sure you can just put your self-signed cert into that file, mount it into your container, and it should be respected by most libraries that reach into the 'systems' trust store (which the new version of `reqwest` library does by default! And was just merged into stump nightly). 

Note by copying this file the system will no longer mange it and you risk that the global certs go out of date.

Technically the way to modify this file is via another command which regenerates it, but appending is the hacky equivalent. It might be possible to modify or extend stumps Dockerfile to inject the cert more officially. See here for more info: https://superuser.com/questions/437330/how-do-you-add-a-certificate-authority-ca-to-ubuntu

aaronleopold (commented):
@gzsombor Does the outlined approach work for you, or are you still looking to get these changes in?

I think, besides the conflicts to resolve, these changes look good

gzsombor:
I would rather not modify the system maintained certificate store, that will surely cause some unexpected errors on the next system upgrade - or when next time I try to test my setup scripts. 
What do you think, adding an environmental variable/config option to specify a private cert and register it explicitly with the http client? That would be a better, more secure option.

gzsombor:
I've opened a new PR (https://github.com/stumpapp/stump/pull/1231) about using a private certificate, so we can close this, as this is IMHO a less secure solution

PR #1222, opened 2026-06-11
🐳 Add `EXPOSE` statement to Dockerfile

### LINKED ISSUE #1221: [BUG] Dockerfile does not expose the Stump port.
### Description of the Bug

In my current home server setup, I have all my containers sharing the same external networking with no ports forwarded, and a container with nginx acting as ingress. This container, taking advantage of the [exposed ports](https://docs.docker.com/reference/dockerfile/#expose) of the other containers, will redirect the requests to the right container. However, Stump does exposes its application port.

### Expected Behavior

Stump could expose its default port, making it accessible to other containers in the same network without having to do port forwarding to the host. The solution is adding in the [line 129 of the Dockerfile](https://github.com/stumpapp/stump/blob/main/docker/Dockerfile#L129) the following statement:

```Dockerfile
EXPOSE 10801
```

### Steps To Reproduce

_No response_

- [ ] Web App
- [ ] Mobile App
- [x] Server

#### Server Details:
  - Release Channel: latest
  - Server OS: OpenMediaVault 8
  - Using Docker: Yes
  - Server Version: latest

### Additional context

_No response_

threkk:
Exposes the default port, configurable using a build argument. It also uses the same argument to configure the environment port for consistency. Closes #1221

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -96,6 +96,8 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry,id=${TARGETARCH} \
 
 FROM gcr.io/distroless/cc-debian12:debug
 
+ARG PORT=10801
+
 LABEL org.opencontainers.image.title="Stump" \
       org.opencontainers.image.description="A free and open source comics, manga and digital book server with OPDS support" \
       org.opencontainers.image.source="https://github.com/stumpapp/stump" \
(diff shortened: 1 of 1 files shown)


aaronleopold (approved):
Thank you!

aaronleopold:
Note for posterity: I did not catch that this was based into main. I cherry picked to `nightly` and removed it from main to let it flow through the normal release process.

threkk:
Thanks for the merge! That was my mistake, I didn't know against which branch I should have opened the PR :(

aaronleopold:
No worries at all, no harm done! I caught it, just not until afterwards

PR #1231, opened 2026-06-16
feat(oidc): add ca_cert_file config option for custom CA certificates

gzsombor:
## Description

Currently, if a self hoster uses a single-signon service for authentication behind a firewall, it is really hard to have a server certificate that is trusted by clients (Let's encrypt coudn't create a certificate if it couldn't reach it). So sometimes it would be great, if Stump could talk to an SSO server, when the certificate is self-signed and explicitly added to Stump.
 The other change is to make the OIDC client shared between the request, so this way we can avoid the repeated initialization in each route.

## Ready?

- [X] I read the [contributing guidelines](https://github.com/stumpapp/stump?tab=contributing-ov-file)
- [X] I searched for existing issues or pull requests that may be related to my contribution
- [X] This PR is based into `nightly` and not `main`
- [X] I added tests and/or documentation for my changes if applicable

- The [expo application](./apps/expo/LICENSE) is licensed under [GPL-3.0](https://www.gnu.org/licenses/gpl-3.0.html)


apps/server/src/config/oidc.rs
```diff
@@ -37,54 +37,104 @@ pub type StumpOidcClient = Client<
...
 
-	let client = CoreClient::from_provider_metadata(
-		provider_metadata,
-		ClientId::new(config.client_id.clone()),
-		Some(ClientSecret::new(config.client_secret.clone())),
-	)
-	.set_redirect_uri(redirect_url);
+/// Create OIDC client from configuration (convenience wrapper for one-shot use).
+/// Prefer [OidcClientState] when handling multiple requests.
+#[allow(dead_code)]
```
aaronleopold: Did we need this annotation? If it isn't in use, we should remove it instead.

apps/server/src/config/oidc.rs
```diff
@@ -37,54 +37,104 @@ pub type StumpOidcClient = Client<
...
-	let redirect_uri = format!("{}/api/v2/auth/oidc/callback", frontend_url);
-	let redirect_url = RedirectUrl::new(redirect_uri).map_err(|e| {
-		APIError::InternalServerError(format!("Invalid redirect URI: {}", e))
-	})?;
+	/// Create a per-request [StumpOidcClient] from the cached state.
+	/// This is cheap — no I/O, just constructs the client with the correct redirect URL.
+	pub fn create_client(&self, frontend_url: &str) -> StumpOidcClient {
+		let redirect_uri = format!("{}/api/v2/auth/oidc/callback", frontend_url);
+		let redirect_url = RedirectUrl::new(redirect_uri)
+			.expect("Invalid redirect URI constructed from frontend URL");
```
aaronleopold: We shouldn't introduce a panic here imo

apps/server/src/main.rs
```diff
@@ -41,6 +41,10 @@ async fn main() -> Result<(), EntryError> {
 		// level is used for logging.
 		init_tracing(&resolved_config);
 
+		if let Some(ref oidc) = resolved_config.oidc {
+			tracing::info!(?oidc, "OIDC configuration");
+		}
```
aaronleopold, line 46: We should not log the config, it contains secrets

core/src/config/stump_config.rs
```diff
@@ -55,6 +55,11 @@ pub mod env_keys {
 	pub const OIDC_ALLOW_REGISTRATION_KEY: &str = "STUMP_OIDC_ALLOW_REGISTRATION";
 	pub const OIDC_DISABLE_LOCAL_AUTH_KEY: &str = "STUMP_OIDC_DISABLE_LOCAL_AUTH";
 	pub const OIDC_EXTRA_AUDIENCES_KEY: &str = "STUMP_OIDC_EXTRA_AUDIENCES";
+	pub const OIDC_DANGER_ACCEPT_INVALID_CERTS_KEY: &str =
+		"STUMP_OIDC_DANGER_ACCEPT_INVALID_CERTS";
+	pub const OIDC_CA_CERT_FILE_KEY: &str = "STUMP_OIDC_CA_CERT_FILE";
+	pub const BOOK_COMPLETION_DEDUP_TIMEOUT_SECS_KEY: &str =
+		"STUMP_BOOK_COMPLETION_DEDUP_TIMEOUT_SECS";
```
aaronleopold: I think this is meant to be:


```suggestion
	pub const OIDC_CA_CERT_FILE_KEY: &str = "STUMP_OIDC_CA_CERT_FILE";
```

apps/server/src/http_server.rs
```diff
@@ -60,6 +60,25 @@ pub async fn run_http_server(config: StumpConfig) -> ServerResult<()> {
 		.await
 		.map_err(|e| ServerError::ServerStartError(e.to_string()))?;
 
+	// Initialize the OIDC client once at startup. If OIDC is configured but
+	// the provider is unreachable, crash hard — no degraded mode.
```
aaronleopold: I think this block is self-describing imo

apps/server/src/http_server.rs
```diff
@@ -60,6 +60,25 @@ pub async fn run_http_server(config: StumpConfig) -> ServerResult<()> {
...
+		if let Some(oidc_config) = config
+			.oidc
+			.as_ref()
+			.filter(|c| c.is_configured() && c.enabled)
+		{
+			let state = OidcClientState::new(oidc_config).await.map_err(|e| {
+				tracing::error!(?e, ?oidc_config, "OIDC client initialization failed");
+				ServerError::ServerStartError(format!("OIDC client init failed: {e:?}"))
+			})?;
+			tracing::info!("OIDC client initialized successfully");
```
aaronleopold:
```suggestion
			tracing::debug!("OIDC client initialized successfully");
```

apps/server/src/routers/api/v2/oidc.rs
```diff
@@ -179,14 +181,15 @@ fn parse_state(state: Option<&str>) -> OidcState {
...
 	HostExtractor(service): HostExtractor,
 	session: Session,
 	Query(query): Query<CallbackQuery>,
 ) -> Result<OidcCallbackResponse, APIError> {
 	let config = &*ctx.config;
 
-	let oidc_state = parse_state(query.state.as_deref());
-	let generate_token = oidc_state.query.generate_token;
+	let oidc_state_parsed = parse_state(query.state.as_deref());
+	let generate_token = oidc_state_parsed.query.generate_token;
```
aaronleopold: Naming is hard lol but I don't love that this collides so closely with `OidcState`. It makes understanding a bit harder when just glancing at it.

I'd suggest renaming `OidcClientState` and then the various var declarations of it accordingly. Maybe `OidcProvider`?

apps/server/src/config/oidc.rs
```diff
@@ -37,54 +37,104 @@ pub type StumpOidcClient = Client<
...
-	let http_client = oauth2_reqwest::ReqwestClient::from(
-		reqwest::ClientBuilder::new()
-			.redirect(reqwest::redirect::Policy::none())
-			.build()
-			.map_err(|error| {
+		let mut client_builder =
+			reqwest::ClientBuilder::new().redirect(reqwest::redirect::Policy::none());
+
+		if let Some(ca_cert_path) = &config.ca_cert_file {
+			let cert_bytes = std::fs::read(ca_cert_path).map_err(|error| {
```
aaronleopold: We should use `tokio::fs` here instead so the read isn't blocking the executor thread

core/src/config/oidc_config.rs
```diff
@@ -36,6 +36,9 @@ pub struct OidcConfig {
 	/// Additional trusted audiences for ID token verification
 	#[serde(default)]
 	pub extra_audiences: Vec<String>,
+	/// Path to a CA certificate file (PEM-encoded) to trust when connecting to the OIDC issuer
+	#[serde(default)]
```
aaronleopold, line 41: I think maybe we can add a `#[graphql(skip)]` here, too. Oh and while you're here, if you can add it to `client_secret` as well

apps/server/src/http_server.rs
```diff
@@ -60,6 +60,25 @@ pub async fn run_http_server(config: StumpConfig) -> ServerResult<()> {
 		.await
 		.map_err(|e| ServerError::ServerStartError(e.to_string()))?;
 
+	// Initialize the OIDC client once at startup. If OIDC is configured but
+	// the provider is unreachable, crash hard — no degraded mode.
+	let oidc_state: Option<Arc<OidcClientState>> = {
+		if let Some(oidc_config) = config
+			.oidc
+			.as_ref()
+			.filter(|c| c.is_configured() && c.enabled)
```
aaronleopold: `is_configured()` will check `enabled` already

apps/server/src/config/oidc.rs
```diff
@@ -37,54 +37,104 @@ pub type StumpOidcClient = Client<
...
+				tracing::error!(?error, path = %ca_cert_path, "Failed to read CA certificate file for OIDC");
+				APIError::InternalServerError(format!(
+					"Failed to read CA certificate file '{}': {}",
+					ca_cert_path, error
+				))
```
aaronleopold, line 70: I think we should avoid exposing `ca_cert_path` to clients, which this would do

apps/server/src/http_server.rs
```diff
@@ -60,6 +60,25 @@ pub async fn run_http_server(config: StumpConfig) -> ServerResult<()> {
...
+		{
+			let state = OidcClientState::new(oidc_config).await.map_err(|e| {
+				tracing::error!(?e, ?oidc_config, "OIDC client initialization failed");
```
aaronleopold: We should avoid logging the config here, too, imo

aaronleopold (commented):
Thanks for adding this! I agree it is a bit safer than the previous approach.

I had a few points of feedback, let me know if any of it needs a bit more discussion

apps/server/src/config/oidc.rs
gzsombor: Good question, no, this function is not needed anymore

core/src/config/stump_config.rs
gzsombor: Yes, you are right, I messed up during rebasing from main to nightly and removing the invalid cert handling code.

apps/server/src/config/oidc.rs
gzsombor: Good point, it could block, but it's only called during startup.

apps/server/src/http_server.rs
gzsombor: It is great, when you want to debug what went wrong, but I understand your concerns.

gzsombor:
Hopefully, I fixed all of your comments :smile:

apps/server/src/http_server.rs
aaronleopold: Yeah that's totally fair. If you want to emit a log you could override the right impls to redact the secret, but even for self-hosted software like this I'd still personally prefer not leaking secrets into the console so would say it is simpler to just remove it (which you did 🙂)

aaronleopold:
> Hopefully, I fixed all of your comments 😄

Thank you! I'll try to come back to this later this evening or tomorrow

apps/server/src/routers/api/v2/oidc.rs
aaronleopold: I'll be by a computer later this evening, but at a glance I think this is the only comment that was not addressed

apps/server/src/main.rs
aaronleopold: Looks like this survived, I thought I saw a commit that removed it but still appears in the diff

apps/server/src/routers/api/v2/oidc.rs
aaronleopold: I think `OidcProvider` is probably fine

aaronleopold (approved):
Thank you! There are a couple items I'll want to clean up but everything else is good to me.

I'll try to sort those out tomorrow, but if you have time feel free to push them up

PR #1235, opened 2026-06-19
♻️ Config generator maintenance

JMicheli:
## Description

Two main changes here:
- General code cleanup in the stump-config-gen macro.
- Implement `Default` on the partial config struct generated by the macro.

The config generator macro is a couple of years old and it seems like it has been working well. I gave it a review and made a few changes to improve the ease of reading the code and eliminate what would be some clippy warnings if clippy worked on procedural macro crates. I also fixed some grammatical issues in the macro's documentation comments.

I think a shortcoming of the way config is approached is that the tests have to be updated in a lot of places whenever a new variable is added. Some of that is probably good since it forces verifying that, e.g., a fresh start will have the value that you expect. But I would also guess it feels tedious. To try and reduce that a _little_ I changed the part of the macro that generates the `PartialStumpConfig` struct so that it derives `Default`. That allowed one of the big structs in the `stump_config::test` module to end with `..Default::default()` which will eliminate the need to add new fields there in the future. 

I wonder if there are other QOL improvements that could be made to the testing approach - open to suggestions.

- The [expo application](./apps/expo/LICENSE) is licensed under [GPL-3.0](https://www.gnu.org/licenses/gpl-3.0.html)

### DIFF
--- a/core/src/config/stump_config.rs
+++ b/core/src/config/stump_config.rs
@@ -471,22 +471,7 @@ mod tests {
 			enable_playground: Some(false),
 			enable_koreader_sync: Some(false),
 			enable_kobo_sync: Some(false),
-			password_hash_cost: None,
-			session_ttl: None,
-			access_token_ttl: None,
-			refresh_token_ttl: None,
-			expired_session_cleanup_interval: None,
-			max_image_upload_size: None,
-			enable_upload: None,
-			max_file_upload_size: None,
-			pdf_render_dpi: None,
-			pdf_max_dimension: None,
-			pdf_render_format: None,
-			pdf_cache_pages: None,
-			pdf_prerender_range: None,
-			pdf_high_quality: None,
-			oidc: None,
-			trust_proxy_headers: None,
+			..Default::default()
 		};
 		partial_config.apply_to_config(&mut config);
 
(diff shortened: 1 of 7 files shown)


core/src/config/stump_config.rs
```diff
@@ -471,22 +471,7 @@ mod tests {
...
-			max_file_upload_size: None,
-			pdf_render_dpi: None,
-			pdf_max_dimension: None,
-			pdf_render_format: None,
-			pdf_cache_pages: None,
-			pdf_prerender_range: None,
-			pdf_high_quality: None,
-			oidc: None,
-			trust_proxy_headers: None,
+			..Default::default()
```
aaronleopold, line 474: 😍

aaronleopold (approved):
lgtm! I _really_ appreciate the `Default` impl and how much it cuts out for tests.

> I wonder if there are other QOL improvements that could be made to the testing approach

The only thing I can think of, and it isn't specific to the testing approach but more generally the ergonomics of the macro, is dealing with nested configs. Like for [oidc](https://github.com/stumpapp/stump/blob/main/core/src/lib.rs#L98-L101). I basically just sidestep the macro by loading it manually today.

Otherwise, nothing else I can think of but I'll keep it in mind.

PR #1255, opened 2026-06-20
fix: prevent PdfiumLibraryBindingsAlreadyInitialized race condition

ethancls:
## Problem

When multiple PDFs are processed concurrently (e.g. during library scanning), every call to `PdfProcessor::renderer()` attempts `Pdfium::bind_to_library()`, which can only succeed **once per process**. The second and subsequent concurrent calls fail with:

```
PdfiumLibraryBindingsAlreadyInitialized → 500 Internal Server Error
```

This breaks PDF thumbnail generation, page rendering, and the built-in reader for all PDF files when more than one PDF is scanned or opened at the same time.

## Fix

Wrap the `Pdfium` instance in a global `Mutex<Option<Pdfium>>`:

- **First call**: binds the PDFium native library and stores the `Pdfium` instance
- **Subsequent calls**: reuse the already-initialized instance
- **`PdfiumRef` wrapper**: holds the `MutexGuard` and implements `Deref<Target=Pdfium>`, so all existing callers work without changes
- **Serialization**: the global lock also serializes PDFium access, preventing the thread-safety issues discussed in #1190

## Changes

`core/src/filesystem/media/format/pdf.rs` — 23 lines added, 16 removed. No other files touched.

## Related

- PR #1190 discussed using `OnceLock<Mutex<Pdfium>>` but the locking mechanism was left as a follow-up item

## Tested

Built and deployed on macOS (ARM64 Docker) with a library containing 100+ PDFs — all thumbnails now generate correctly, no more 500 errors.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/src/filesystem/media/format/pdf.rs
+++ b/core/src/filesystem/media/format/pdf.rs
@@ -1,7 +1,9 @@
 use std::{
 	collections::HashMap,
 	io::Cursor,
+	ops::Deref,
 	path::{Path, PathBuf},
+	sync::{Mutex, MutexGuard},
 };
 
 use models::shared::image_processor_options::SupportedImageFormat;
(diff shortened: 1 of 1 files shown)


aaronleopold:
There is already an active effort to improve this situation in https://github.com/stumpapp/stump/pull/1209

PR #1277, opened 2026-07-02
feat(metadata): manual per-book metadata search and apply

### LINKED ISSUE #789: [FEATURE] Metadata Search System
Feature Request
A built in way to search some form of service (for example comicvine) for metadata for a manga or comic completely missing or partially missing Metadata.

An option in the metadata management to search for it.

arunoruto:
## Description

**Disclaimer**: this PR has been made using Claude Fable 5, but has been tested extensively on a NixOS server and a test (copy) library.

Metadata fetching was built around the fully automatic library-wide scan job: it searches every enabled provider with the book's stored name and may auto-apply matches above a confidence threshold. There was no way to intervene for a single book, so a badly named file ("08 - Ascension.epub") could never be matched manually against the correct title/author.

This adds a "Search metadata" action to the book management scene (gated behind MetadataFetchRecordManage) with title/author/isbn/year fields and a provider picker, feeding results into the existing MatchReviewDialog accept/reject flow, and fixes several defects that surfaced along the way.

API:
- Add MediaMetadataSearchInput as an optional 'search' argument on fetchMediaMetadata. When present, caller fields take precedence over stored metadata, the search is restricted to one provider if given, and auto-apply is always skipped so nothing is written without review. The mutation now returns the MetadataFetchRecord (matching accept/reject) instead of a bare candidate list. The bulk job in fetch_job.rs is untouched.

Applying matches:
- ExternalMediaMetadata.series_name/.number were fetched from providers and threaded through GraphQL into the frontend fragment, but dropped on both ends: MEDIA_FIELD_DEFS had no candidateKey for Series/Number so the review table filtered the rows out, and apply_media_fields/build_media_metadata_insert never wrote the columns. Both are now handled like every other field. This populates the ComicInfo-style media_metadata.series text column, not the relational media.series_id.

Partial results:
- Hardcover requires one search plus a separate per-hit detail fetch; any single detail fetch failing (e.g. rate limited) silently dropped that candidate with only a log line. search_media/ search_series now return SearchOutcome{candidates, requested}, the shortfall is persisted as a new had_partial_results column on metadata_fetch_records (additive migration, DEFAULT FALSE), and the UI shows a toast suggesting a retry.

Review strategy:
- The review dialog defaults to FillGaps, which is right for the unattended bulk scan but wrong for a manual search: a book with any title at all would never be offered a correction. The store's open() now accepts a strategy and only the manual entry point passes PreferExternal; per-field exclusions and overrides still apply, bulk behavior is unchanged.

Frontend enablement:
- Hoist the PendingMatchRecord fragment out of PendingMatchesTable into a shared fragments.ts so the new dialog can reuse it.
- Extend useMatchActions' cache invalidation to also refresh the book's own query keys after accepting/rejecting a match.
- Add Number/Volume as sort options for book listings; the MediaMetadataModelOrdering GraphQL path existed but no UI emitted it, so a populated series position can now actually order books.
- Expose METADATA_FETCH_RECORD_READ/MANAGE and METADATA_PROVIDER_READ/MANAGE as a "Metadata management" group in the user permissions table. The backend has enforced these since the metadata foundations landed, but no checkbox ever rendered, so non-owner admins could not be granted access.

All schema changes are additive and the migration runs on startup.

---

Closes #789 (at least the system is there, but not the provider... yet?)

If this PR does not close an issue, please remove the above line. If it does, please make sure to link the issue properly so it can be automatically closed when this PR is merged.

## Screenshots

image

image

image

## Ready?


aaronleopold:
> Disclaimer: this PR has been made using Claude Fable 5, but has been tested extensively on a NixOS server and a test (copy) library

I appreciate the disclosure, thank you!

***

I'll aim to take a look at this towards the end of the week, but might wind up being over the weekend

> Closes #789 (at least the system is there, but not the provider... yet?)

The provider mentioned in the issue is in progress via https://github.com/stumpapp/stump/pull/1169, just haven't had a moment to see it through. It's in the queue, though!

arunoruto:
I know that its a lot of code and sadly my rust skills are subpar, but I am currently learning it so I can contribute to projects like these with my own code!

core/src/filesystem/metadata/fetch.rs
```diff
@@ -97,8 +98,11 @@ pub async fn fetch_series_metadata(
 				};
 
 				match provider.search_series(&query).await {
-					Ok(candidates) => {
-						all_candidates.extend(candidates);
+					Ok(outcome) => {
+						if outcome.failed() > 0 {
+							had_partial_results = true;
+						}
```
aaronleopold: It might be answered through the process of me reviewing, but curious why this was added?

core/src/filesystem/metadata/fetch.rs
```diff
@@ -204,6 +212,14 @@ pub async fn fetch_media_metadata(
 	let provider_configs =
 		filter_providers_for_library_type(provider_configs, &library_type);
 
+	let provider_configs = match provider_filter {
+		Some(provider) => provider_configs
+			.into_iter()
+			.filter(|c| c.provider_type == provider)
+			.collect(),
+		None => provider_configs,
+	};
```
aaronleopold: Double filtering feels awkward at a glance, perhaps a `filter_providers` fn that takes a `library_type` or a set `provider_filter` as the filter?

It also makes the error messages underneath slightly inaccurate in some scenarios, e.g. `No enabled metadata providers configured for this library type` should probably be `No enabled metadata providers for configured filters` (with two pieces in log)

core/src/filesystem/metadata/fetch.rs
```diff
@@ -268,37 +289,40 @@ pub async fn fetch_media_metadata(
...
-		if let Err(e) = apply::apply_media_match(
-			conn,
-			media_id,
-			&candidate,
-			config.strategy,
-			config.exclude_fields,
-			vec![],
-		)
-		.await
+	if !skip_auto_apply {
```
aaronleopold, line 320: This gate shouldn't be necessary, since `find_auto_apply_candidate` considers auto-apply opt-in. It would save some compute I guess, so not against it being here

crates/models/src/entity/metadata_fetch_record.rs
```diff
@@ -25,6 +25,9 @@ pub struct Model {
 	#[sea_orm(column_type = "Json", nullable)]
 	#[graphql(skip)]
 	pub accepted_match_candidate: Option<JsonValue>, // auto or manual
+	/// True if a provider search found more raw hits than could be turned into
+	/// candidates (e.g. a per-hit detail fetch failed after the initial search)
+	pub had_partial_results: bool,
```
aaronleopold: Perhaps instead we can just track `raw_hits` as a number, if this is needed on the frontend somewhere I haven't encountered yet?

packages/browser/src/scenes/book/settings/BookManagementScene.tsx
```diff
@@ -133,6 +137,31 @@ export default function BookManagementScene() {
 					</div>
 				)}
 
+				{checkPermission(UserPermission.MetadataFetchRecordManage) && (
+					<div className="gap-y-2 flex flex-col">
+						<div>
+							<Heading size="sm">Search metadata</Heading>
+							<Text size="sm" variant="muted">
+								Search an external provider for this book and choose a match to apply
+							</Text>
```
aaronleopold: It's something I am trying to be better about myself, but while we are here I'd prefer to localize these strings immediately

packages/browser/src/scenes/book/settings/BookManagementScene.tsx
```diff
@@ -133,6 +137,31 @@ export default function BookManagementScene() {
...
+						</div>
+
+						<div>
+							<Button size="default" onClick={() => setIsSearchingMetadata(true)}>
+								Search Metadata
+							</Button>
```
aaronleopold, line 154: Same here

packages/browser/src/scenes/book/settings/BookMetadataSearch.tsx
```diff
@@ -0,0 +1,197 @@
...
+	})
+
+	const handleSubmit = () => {
+		mutate({
+			id: mediaId,
+			search: {
+				title: title.trim() || undefined,
+				author: author.trim() || undefined,
+				isbn: isbn.trim() || undefined,
+				year: year.trim() ? Number(year) : undefined,
```
aaronleopold: I'd `parseInt(year, 10)` on the year, instead, and prevent passing `NaN` values

aaronleopold (commented):
Sorry for the delay! This looks really good, I had a few nitpicks but otherwise mostly just a few points for my own clarification

packages/browser/src/scenes/book/settings/BookMetadataSearch.tsx
arunoruto: Will switch to `parseInt(year, 10)` and skip the field entirely if it comes back NaN instead of trusting `Number()`

packages/browser/src/scenes/book/settings/BookManagementScene.tsx
arunoruto: Agreed, will move those into the locale file like the rest of the new UI. (Noticed the existing "Analysis" section above it is also hardcoded... idn't touch that since it's pre-existing, but happy to clean it up too in this pass if you'd rather bundle it)

crates/models/src/entity/metadata_fetch_record.rs
arunoruto: Agree, `SearchOutcome` already carries requested so there's no reason to collapse it into a bool at the DB layer. I'll switch the column to `raw_hits: i32` (or similar) and derive the "were results incomplete" check from `raw_hits > candidates.len()` wherever it's needed.

packages/browser/src/scenes/book/settings/BookManagementScene.tsx
aaronleopold: Yeah there are a lot of surfaces in the web app that are still hardcoded, you're more than welcome to clean up existing items as you notice (always greatly appreciated) but I only care about asking for the new stuff added to be localized. That way I don't increase the amount of work to go back to for translations, get it when added and what not

core/src/filesystem/metadata/fetch.rs
arunoruto: Yeah, that's fair, it does read a bit clunky as two sequential filters. I'll fold it into one `filter_providers(configs, library_type, provider_filter)` helper and split the error message so it's accurate for both cases...

core/src/filesystem/metadata/fetch.rs
arunoruto: Hardcover's search is two-step: the initial search returns raw hits, but each one needs a separate per-hit detail fetch, and any of those can fail on its own (rate limiting mostly). Before this, a failed detail fetch just got logged and the candidate silently disappeared — so a manual search could come back looking like "no match" when really one just got dropped mid-fetch. Since the whole point of this PR is letting someone deliberately go looking for the right match, I didn't want a transient failure to look identical to "there's really nothing here." That's what the flag is surfacing.

core/src/filesystem/metadata/fetch.rs
arunoruto: I think this one might actually need to stay for correctness rather than just perf... `find_auto_apply_candidate` being opt-in is per-provider-config (whether that provider has auto-apply turned on), not per-request. So if someone does have auto-apply enabled on a provider, a manual search without this gate would still risk silently applying a match before the user gets to see it — which is the exact behavior this PR is trying to get away from. So I'd like to keep the explicit `skip_auto_apply: true` on the manual path as a hard guarantee, independent of whatever the provider config says. Let me know if I'm missing something on your end though!

aaronleopold: Ah okay, if true then that makes sense! When I have a moment to take another look at your changes (thank you, by the way!) I'll take extra care to re-read that bit

aaronleopold (approved):
lgtm, thank you! once CI passes I'll go ahead and merge this in

arunoruto:
Glad it could be merged! Looking forward to the next release! 😁

PR #1281, opened 2026-07-06
:bug: Fix multiple job scheduler issues

### LINKED ISSUE #1274: [BUG] Errors messages when creating a scheduled job should be displayed to the user
### Description of the Bug

When creating a scheduled job, the server may reject the form data and provide a proper error message, but this error message is never shown the the end-user. I spent several minutes trying to figure out why this one wasn't working, as all I got was:

Failed to create scheduled job

I tried looking at server logs (`docker compose logs -f`), and found nothing. I tried looking at the [docs](https://www.stumpapp.dev/docs), and found nothing. Finally, after opening the browser dev tools, I found this at the Network tab:

Invalid cron expression

### Expected Behavior

The data from `response.errors.map(e => e.message).join('\n')` should have been displayed to the user. Otherwise, the user won't know what went wrong.

### Steps To Reproduce

1. Go to `/settings/jobs`
2. Click on `New scheduled job`
3. Type a name, such as "Weekly library scan"
4. Select `Weekly (Sunday)`, which auto-fills the cron expression
5. Click on `Create`
6. BUG!

- [x] Web App
- [ ] Mobile App
- [x] Server

#### Server Details:
  - Release Channel: stable
  - Server OS: Debian
  - Using Docker: Yes
  - Server Version: 0.1.3 commit 9546397727b604b8f790354319764b2379fb10fc

#### Client Details:
  - Device: PC
  - OS: N/A
  - Browser: Firefox (the browser doesn't matter for this bug)

### Additional context

There are three bonus bugs:

1. The built-in `Weekly (Sunday)` cron expression isn't accepted by the server.
2. 5-element cron expressions are usually valid, but they are (likely) not accepted by the server.
3. There is no client-side validation for invalid cron expressions (i.e. a simple regex pattern would probably be enough).

I'm gonna submit new bugs for those.

### LINKED ISSUE #1275: [BUG] Problems with the cron expression at the frontend
### Description of the Bug

The current cron expression support is limited and also buggy in certain edge cases. I'm putting multiple small issues together in this bug report, as they are related and will very likely be fixed together.

#### Broken by default

The built-in preset `Weekly (Sunday)` fills the cron expression with `0 0 0 * * 0`, which will then give an error at the backend: `Days of Week must be greater than or equal to 1`.

This sounds like a backend error, as [cron expressions][w] are written with days of the week from 0 to 6.

#### 5-field vs 6-field expressions

Most [cron expressions][w] only have 5 fields. Even the most common cron explainers on the web also use 5 fields: <https://crontab.guru/> <https://it-tools.tech/crontab-generator>

And yet, if using a 5-field version of [04:02 on Sunday](https://crontab.guru/#2_4_*_*_0), the backend tries to parse the weekday as a month:

```
Invalid cron expression:
2 4 * * 0
        ^
Months must be greater than or equal to 1. ('0' specified.)
```

This is unexpected and surprising, as people expect 5-field cron expressions.

#### Lack of feedback for errors

I've already reported #1274.

Additionally, the `<input name="schedule">` should have a [pattern="…" attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/pattern) to give users early feedback about issues. Of course, that only works if we add something like `:invalid { border-color: red }` to the CSS.

This regex can work:

     ([-,0-9]+ +|\* +)?([-,0-9]+|\*) +([-,0-9]+|\*) +([-,0-9]+|\*) +([-,0-9A-Z]+|\*) +([-,0-9A-Z]+|\*)

It supports both 5 and 6 fields. It also supports named months and weekdays. It accepts things that look like a cron expression, but doesn't try to semantically parse it. And, for the sake of easy-to-understand, it doesn't implement the full syntax (which would make a gigantic regex pattern). Thus, `123465 1-----3 ,,,,,,,,, JAN-ZZZ SAT-SAT` would pass the check, but hopefully be rejected by the backend.

#### Support for named months and weekdays

[According to the Wikipedia][w], cron expressions are expected to support named months (e.g. `JAN`) and weekdays (e.g. `SUN`). These are indeed supported by the online tools mentioned above.

Frankly, I haven't tested it myself, so I don't know if they work properly.

And if it works, perhaps the built-in default expressions (from the `<select>` box) should use names instead of (cryptic) numbers. It would make them easier to understand for users.

[w]: https://en.wikipedia.org/wiki/Cron#Cron_expression

### Expected Behavior

* [ ] 5-field cron expressions should be supported
* [ ] weekday should start at 0
* [ ] weekdays and months could use names instead of numbers (untested, maybe already supported)
* [ ] cron expressions should be validated client-side using the [pattern="…" attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/pattern).
* [ ] invalid cron expressions should have a proper `:invalid { border-color: red }` CSS.

### Steps To Reproduce

1. Go to `/settings/jobs`
2. Click on `New scheduled job`
3. Type a name, such as "Weekly library scan"
4. Select `Weekly (Sunday)`, which auto-fills the cron expression
5. Click on `Create`
6. BUG!

Alternatively, change the cron expression by one with just 5 pieces instead of 6.

- [x] Web App
- [ ] Mobile App
- [x] Server

#### Server Details:
  - Release Channel: stable
  - Server OS: Debian
  - Using Docker: Yes
  - Server Version: 0.1.3 commit 9546397727b604b8f790354319764b2379fb10fc

#### Client Details:
  - Device: PC
  - OS: N/A
  - Browser: Firefox (the browser doesn't matter for this bug)

### Additional context

_No response_

aaronleopold:
Fixes #1274
Fixes #1275

## Description

- Swap to `croner` crate for 5-field support
- Put server error message into toast body upon error
- Add basic client-side validation for regex input with localized message

A few notes/responses to ac outlined in the original issues:

> cron expressions should be validated client-side using the [pattern="…" attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/pattern)

I did not go this route, I leaned into the existing validation patterns used in Stump.

> invalid cron expressions should have a proper :invalid { border-color: red } CSS

This kinda thing will work out of the box now that the cron has validation, outside of the `min(1)` that existed in latest.

***

I will leave as draft until I can more properly test the changes e2e


aaronleopold:
I am going to merge this in shortly, I have not had time to put it through its paces but will let the nightly folks help confirm

PR #1286, opened 2026-07-14
:bug: Fix data loss during RAR-to-ZIP conversion

### LINKED ISSUE #1284: Data loss: convert_rar_to_zip destroys source CBRs and produces corrupt CBZs
## Description of the Bug

With `convert_rar_to_zip` enabled on a library, scanning CBR files produces **corrupt, near-empty CBZ archives** and then **sends the original CBR to the trash**. The source is deleted without any validation of the produced archive.

On my library this silently replaced ~95 volumes (11 GB of CBRs, 100–230 MB each) with stub `.cbz` files of 150–250 KB. The originals were recoverable only because `trash::delete` happened to work — with `hard_delete_conversions` enabled they would have been permanently lost.

### Impact

- Originals are deleted (trashed) after every conversion, **regardless of whether the conversion succeeded**.
- The produced CBZ contains a single ZIP entry with an **empty filename**, holding only the *last* image of the RAR.
- Every converted book becomes unreadable. Stump then logs `Failed to find valid image in zip file` / `Empty zip file` for them on subsequent scans.

### Root cause

Four bugs chain together.

#### 1. Every RAR entry is extracted to the same destination path

`core/src/filesystem/media/format/rar.rs`

```rust
let unpacked_path = cache_dir.join(file_stem);

while let Ok(Some(header)) = archive.read_header() {
    archive = if header.entry().is_file() {
        header.extract_to(&unpacked_path)?   // <-- same path for EVERY entry
    } else {
        header.skip()?
    };
}
```

`extract_to()` treats its argument as the **destination file**, not a directory. All entries are written to the same path in turn, each overwriting the previous one. When the loop ends, `unpacked_path` is a *single file* (the last image), not a directory of pages.

#### 2. `zip_dir` then produces an entry with an empty name

`core/src/filesystem/archive.rs`

```rust
for entry in WalkDir::new(unpacked_path) {
    let path = entry.path();
    let name = path.strip_prefix(Path::new(prefix)).unwrap();  // prefix == unpacked_path
    if path.is_file() {
        zip_writer.start_file_from_path(name, options)?;
```

Since `unpacked_path` is a file, `WalkDir` yields exactly one entry: the path itself. `strip_prefix` of a path by itself returns an **empty** path, so the ZIP entry gets a zero-length filename.

Confirmed on a produced file:

```
$ zipinfo -v "One Piece T001 (...).cbr.cbz"
  compression method:      none (stored)
  length of filename:      0 characters

$ unzip -l "One Piece T001 (...).cbr.cbz"
  Length      Date    Time    Name
---------  ---------- -----   ----
   179032  1980-01-01 00:00
---------                     -------
   179032                     1 file
```

For three volumes the extraction failed entirely and the resulting `.cbz` was **22 bytes** (an empty ZIP central directory).

#### 3. Output filename keeps the original extension

`core/src/filesystem/media/format/rar.rs`

```rust
let zip_path = create_zip_archive(&unpacked_path, &file_name, &extension, parent)?;
//                                                 ^^^^^^^^^^ file_name, not file_stem
```

and in `archive.rs`:

```rust
let zip_path = destination.join(format!("{name}.{ext}"));
```

`file_name` still contains the extension, so the output is `Foo.cbr.cbz` instead of `Foo.cbz`. This also breaks the book title in the UI, since `name` is derived from `file_stem` (`Foo.cbr`).

#### 4. The source is trashed without validating the output

```rust
// TODO: won't work in docker
if delete_source {
    if let Err(err) = trash::delete(path) {
        warn!(error = ?err, path, "Failed to delete converted RAR file");
    }
}
```

The deletion is unconditional on the success of the conversion. (Side note: the `won't work in docker` comment is inaccurate — it does work, and moved my originals to `/opt/media/books/.Trash-1000`.)

## Expected Behavior

- Conversion should produce a valid CBZ containing every page of the source RAR, with correct per-entry filenames and image extensions.
- The output file should be named `<name>.cbz`, not `<name>.cbr.cbz`.
- The original CBR should only be deleted **after** the produced archive is validated to be non-empty and readable. A failed conversion must leave the source untouched.

### Suggested fixes

- Extract each entry to its own path (e.g. `unpacked_path.join(entry_name)`), or use `extract_with_base`.
- Pass `file_stem` rather than `file_name` to `create_zip_archive`.
- **Validate the produced archive** (reopen it, check entry count and that entries have non-empty names / image extensions) **before** deleting the source.
- Abort the conversion and leave the source untouched if the produced archive contains zero images.
- Consider whether the feature should delete sources at all by default.

## Steps To Reproduce

1. Put a valid multi-page `.cbr` in a library.
2. Enable `convert_rar_to_zip` in the library config.
3. Scan the library.
4. Observe: a `<name>.cbr.cbz` of a few hundred KB appears; the original `.cbr` is gone (in `.Trash-<uid>`); subsequent scans log `Failed to find valid image in zip file`.

## Where is this issue happening?

- [x] Server

#### Server Details:
  - Release Channel: latest (also reproduced on nightly — same code)
  - Server OS: Docker (`aaronleopold/stump`)
  - Using Docker: Yes
  - Server Version: 0.1.5
  - Database: SQLite

## Additional context

Stump reads CBR natively via the `unrar` crate, so this conversion isn't required for functionality. Given the blast radius, it may be worth disabling the option until the extraction is fixed.

aaronleopold:
Fixes #1284

## Description

A clear and concise description of the PR.

- Fix double extension in destination path
- Fix same destination path for each entry
- Basic validation before trashing file
   - I did not implement all the suggested validations, I tracked reason in comment

### DIFF
--- a/core/src/filesystem/error.rs
+++ b/core/src/filesystem/error.rs
@@ -40,6 +40,8 @@ pub enum FileError {
 	RarReadError,
 	#[error("Error reading RAR byte content")]
 	RarByteReadError(#[from] std::str::Utf8Error),
+	#[error("RAR archive is empty")]
+	RarEmpty,
 	#[error("Unsupported file type: {0}")]
 	UnsupportedFileType(String),
 	#[error("{0}")]
@@ -50,6 +52,8 @@ pub enum FileError {
 	DirectoryReadError,
 	#[error("Incorrect image processor for requested format")]
 	IncorrectProcessorError,
+	#[error("File not found on disk")]
+	NotFound,
 	#[error("An unknown error occurred: {0}")]
 	UnknownError(String),
 }
--- a/core/src/filesystem/media/format/rar.rs
+++ b/core/src/filesystem/media/format/rar.rs
@@ -31,6 +31,30 @@ use crate::{
 /// A file processor for RAR files.
 pub struct RarProcessor;
 
+fn validate_converted_zip(zip_path: &Path) -> Result<(), FileError> {
+	if !zip_path.exists() {
+		return Err(FileError::NotFound);
+	}
+
+	let metadata = std::fs::metadata(zip_path)?;
+	if metadata.len() == 0 {
+		return Err(FileError::ArchiveEmptyError);
+	}
+
+	let file = std::fs::File::open(zip_path)?;
+	let archive = zip::ZipArchive::new(file)?;
+
+	if archive.is_empty() {
+		return Err(FileError::ArchiveEmptyError);
+	}
+
+	// note: there are more checks we can likely add here, however i'd like to think
+	// about it a bit more before adding them to prevent accidentally failing valid
+	// conversions. see https://github.com/stumpapp/stump/issues/1284
+
+	Ok(())
+}
+
 impl RarProcessor {
 	fn init() {
 		// See https://github.com/muja/unrar.rs/issues/44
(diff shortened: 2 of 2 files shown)


aaronleopold:
I updated the tests to assert the fixes but haven't had time to do much real testing. I want to get this into nightly, though, so will hope that release channel can help validate

PR #1288, opened 2026-07-16
:sparkles: Support epub streaming and Readium web

### LINKED ISSUE #62: Epub streaming
Add support for streaming pieces of an epub to a client. This is already partially implemented in the Rust core.

Sump already supports *very* basic epub fetching and rendering (#34). It consumes the entire epub file, which is NOT ideal. I would like to be able to stream parts of the epub file as the user reads it, as opposed to downloading the entire thing to read it on the client. This will require a lot of research on epub file structure, as the core will likely have a lot of work required to support that kind of flow. Some work has already been started towards this.

Once there is a proof of concept for some sort of in house ebpub engine, I'll probably separate it into a standalone library that can be installed for both Rust and JavaScript/TypeScript projects (using likely https://github.com/getditto/safer_ffi).

> References:
> - https://github.com/getditto/safer_ffi -> for using Rust library in node
> - https://readium.firebaseapp.com/?epub=epub_content%2Fpage-blanche&goto=epubcfi(/6/2!/4/2/2)
> 
> above link appears to: [grab the container.xml](https://readium.firebaseapp.com/epub_content/page-blanche/META-INF/container.xml), [grab the package.opf](https://readium.firebaseapp.com/epub_content/page-blanche/EPUB/package.opf), [grab encryption.xml](https://readium.firebaseapp.com/epub_content/page-blanche/META-INF/encryption.xml), [grab nav.xhtml](https://readium.firebaseapp.com/epub_content/page-blanche/EPUB/Navigation/nav.xhtml), grabs a page, like [page 6](https://readium.firebaseapp.com/epub_content/page-blanche/EPUB/Content/PageBlanche_Page_006.xhtml), grabs [epub styles](https://readium.firebaseapp.com/epub_content/page-blanche/EPUB/Style/style.css)
> 
> - https://www.iana.org/assignments/media-types/application/epub+zip

sandmor:
Closes #62, #660
Develops or closes the EPUB annotations were implemented as part of the readium based web client. Including support for highlights and notes, this wasn't the focus so I'm not sure if is enough to close the issue.

If this PR does not close an issue, please remove the above line. If it does, please make sure to link the issue properly so it can be automatically closed when this PR is merged.

## Description

Workarounds for readium web were required, alternatives were explored but ultimately the library was just not ready for some features at the time of implementing these changes.

The most notable of these workarounds is patchDurableIframeSrc: Readium's FrameManager navigates via contentWindow.location.replace(blobUrl) and never sets iframe.src. Any browser-driven reload (observed on window resize) reloads from the default src (about:blank). The patch changes this to iframe.src = blobUrl, with a full frame-recovery pipeline for post-resize blank frames. This touches private/unexported Readium internals, it is documented with a version pin warning.

Only the web client is affected by this PR and Expo is untouched. Expo was used as inspiration for some functionality.

The search feature is minimal, a linear spine scan per request. The goal was to keep it simple for an intiial implementation.

AI was used in this PR, part of the code was written by it and human reviewed, and I used it extensively to familiarize myself with the codebase and relevant libraries used in the application.

## Ready?


aaronleopold:
I'm looking forward to having the time to give this a proper review! At a glance, it looks like there was a lot of miscellaneous formatting to files largely unrelated to the task at hand. Any way we could trim those out of this PR? It should make things a lot more manageable for me to review

sandmor:
Done, yes the formatting issues all mapped to a single commit that I managed to clean. Now the formatting noise should be fixed!

I know the PR is very large, I tried to separate the changes logically but still the final version ended up being massive so don't worry take your time to review, and if there is any issue or question don't doubt to reach out here or in discord.

apps/desktop/vite.config.ts
```diff
@@ -24,6 +24,17 @@ export default defineConfig({
...
+	resolve: {
+		alias: {
+			react: resolve(__dirname, '../../node_modules/react'),
+			'react-dom': resolve(__dirname, '../../node_modules/react-dom'),
+		},
+		dedupe: ['react', 'react-dom'],
+	},
+	optimizeDeps: {
+		include: ['react', 'react-dom', 'react/jsx-runtime', 'react/jsx-dev-runtime'],
+	},
```
aaronleopold: Do we actually need this? I noticed react was pinned in the resolutions block in the root `package.json` so would think that would cover us?

core/src/filesystem/media/epub_cfi.rs
aaronleopold, line 1: Honestly this is pretty cool, but if we are dropping epub.js and have no need for epubcfi strings anymore i'm curious why we opted have this translation layer at all? 

I think it would be acceptable to just remove support entirely, the blast radius shouldn't be too large since annotations are really only supported for the mobile app anyways, which already uses locators, so folks would lose progress on active ebooks + bookmarks (which stinks, but would be much more maintainable for me)

core/src/filesystem/media/epub_cfi.rs
```diff
@@ -0,0 +1,192 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum EpubCfiResolveError {
+	InvalidFormat,
+	TooLong,
+	MissingContentComponent,
+	SpineNotFound,
+	OpenFailed(String),
+}
+
+impl std::fmt::Display for EpubCfiResolveError {
```
aaronleopold: For errors on the Rust side of things we can use `thiserror` instead

apps/server/src/routers/api/v2/epub.rs
```diff
@@ -1,52 +1,99 @@
...
 		.layer(middleware::from_fn_with_state(app_state, auth_middleware))
 }
 
-/// Get a resource from an epub file. META-INF is a reserved `root` query parameter, which will
-/// grab a resource by resource ID (e.g. `META-INF/container.xml`, where `container.xml` is the
-/// resource ID). Otherwise, the `resource` query parameter represents the path to the requested
-/// resource. (e.g. `/EPUB/chapter1.xhtml`, where `EPUB` is the root and `chapter1.xhtml` is
-/// the resource path)
+// TODO(readium): Can probably kill the get_epub_meta and legacy chapter routes once
+// the web client fully migrates to rwpm streaming.
```
aaronleopold: I'm pretty sure this is my comment, since I know you based originally off of that branch I started, but should sort it out here imo

apps/server/src/routers/api/v2/epub.rs
```diff
@@ -1,52 +1,99 @@
...
+	host_details: &HostDetails,
+	id: &str,
+) -> APIResult<String> {
+	let origin = match server_config::Entity::find().one(conn).await? {
+		Some(config) => config
+			.public_url
+			.filter(|url| !url.is_empty())
+			.unwrap_or_else(|| host_details.url()),
+		None => host_details.url(),
+	};
```
aaronleopold: The pattern used throughout the rest of the app constructs URLs based on how Stump is being accessed, e.g., if accessed by direct IP it would construct with IPs, etc. I think this has the potential to be incorrect, if for whatever reason I am unable to resolve the public URL (e.g., behind a VPN).

The `public_url` is intended to be used for externally-facing features not yet finalized

apps/server/src/routers/api/v2/epub.rs
```diff
@@ -88,3 +129,192 @@ async fn get_epub_meta(
...
+	let ebook = find_ebook_for_user(ctx.conn.as_ref(), &user, &id).await?;
+
+	let base_url = epub_service_base_url(ctx.conn.as_ref(), &host_details, &id).await?;
+	let generator = ReadiumManifestGenerator::new(&ebook.path, base_url);
+	let manifest = generator.generate_manifest()?;
+
+	Ok((
+		[(header::CONTENT_TYPE, "application/webpub+json")],
+		Json(manifest),
+	))
```
aaronleopold: We can shorthand this if we impl `IntoResponse` for `RWPManifest` so that it uses the content type we want in the response. If you want a reference, I did this for `OPDSBasicAuth`

apps/server/src/routers/api/v2/epub.rs
```diff
@@ -88,3 +129,192 @@ async fn get_epub_meta(
...
+	let generator = ReadiumManifestGenerator::new(&ebook.path, base_url);
+	let positions = generator.generate_positions()?;
+
+	Ok((
+		[(
+			header::CONTENT_TYPE,
+			"application/vnd.readium.position-list+json",
+		)],
+		Json(positions),
+	))
```
aaronleopold: Same as above but for `RWPMPositions`

apps/server/src/routers/api/v2/epub.rs
```diff
@@ -88,3 +129,192 @@ async fn get_epub_meta(
...
+}
+
+#[derive(Debug, Deserialize)]
+struct EpubResolveCfiQueryParams {
+	cfi: String,
+}
+
+/// Best-effort legacy `epubcfi` → Readium locator resolution (read-only).
+async fn get_epub_resolve_cfi(
```
aaronleopold: I'll stop commenting on it but I think just removing the cfi stuff everywhere will be a lot cleaner

aaronleopold:
I've started reviewing this today and will try to finalize feedback by the weekend

apps/server/src/routers/api/v2/epub.rs
```diff
@@ -88,3 +129,192 @@ async fn get_epub_meta(
...
+	let path = ebook.path.clone();
+
+	// Dropping the request future cancels the token so the blocking scan can stop.
+	struct CancelOnDrop(CancellationToken);
+	impl Drop for CancelOnDrop {
+		fn drop(&mut self) {
+			self.0.cancel();
+		}
+	}
+	let _guard = CancelOnDrop(cancel);
```
aaronleopold: I don't think we need to create effectively our own version of [DropGuard](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.DropGuard.html). E.g., we should be able to do:


```suggestion
	let _guard = cancel.drop_guard();
```

crates/models/src/services/reading_progress.rs
```diff
@@ -10,6 +10,17 @@ use crate::{
...
+/// Which coordinate system an incoming progression update is authoritative for.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum ProgressLocationSource {
+	/// Percentage / elapsed / completion only — do not change stored locations.
+	#[default]
+	None,
+	Readium,
+	EpubCfi,
+	Page,
+}
```
aaronleopold: I don't think we need this, and especially if we do away with cfi tracking. My understanding of how it is used (jumping around a bit to get a better picture) is to narrow what gets updated during progression mutations to preserve previously-set fields, but this is a highly unlikely scenario since the format of the book will not change mid-reading (ignoring cfi vs readium since again my preference would be to remove it). Let me know if I am misunderstanding, though

packages/browser/src/components/book/BookCard.tsx
```diff
@@ -85,12 +89,9 @@ const BookCard = memo(function BookCard({
 		if (!data.readProgress && !data.readHistory) {
 			return null
 		} else if (data.readProgress) {
-			const { epubcfi, percentageCompleted, page } = data.readProgress
-			if (epubcfi && percentageCompleted) {
-				return Math.round(percentageCompleted * 100)
-			} else if (page) {
-				const percent = Math.round((page / data.pages) * 100)
-				return Math.min(Math.max(percent, 0), 100)
+			const percent = readProgressPercent(data.readProgress, data.pages, data.extension)
```
aaronleopold, line 91: Nit: `readProgressPercent` is seemingly used for multiple formats but comes from `@/utils/ebookProgress`

docs/content/docs/apps/web/readers.mdx
aaronleopold, line 1: Thanks for touching the docs, too!

packages/browser/src/components/readers/epub/annotations/AnnotationDialog.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+	const handleSave = useCallback(() => {
+		onSave(note.trim())
+	}, [note, onSave])
+
+	return (
+		<Dialog open={open} onOpenChange={isPending ? undefined : onOpenChange}>
+			<Dialog.Content size="sm">
+				<Dialog.Header>
+					<Dialog.Title>{mode === 'create' ? 'Add note' : 'Edit annotation'}</Dialog.Title>
```
aaronleopold: While we are here, I'd prefer to localize these strings to reduce future work of finding and localizing after merge. This applies generally for all UI changes, but will leave this anchored here

packages/browser/src/components/readers/epub/annotations/AnnotationDialog.tsx
```diff
@@ -0,0 +1,91 @@
...
+}: Props) {
+	const [note, setNote] = useState(initialNote ?? '')
+
+	useEffect(() => {
+		if (open) {
+			setNote(initialNote ?? '')
+		}
+	}, [open, initialNote])
+
+	const handleSave = useCallback(() => {
```
aaronleopold: We can probably let react-compiler optimize for us here, and other simple, manual memoizations too

packages/browser/src/components/readers/epub/annotations/locator.ts
```diff
@@ -0,0 +1,188 @@
...
+ * before persistence. Prefer selection text / progression when present.
+ */
+export function enrichSelectionLocator(args: {
+	selectionLocator: Locator | ReaderLocator | null | undefined
+	selectedText: string
+	positions: Locator[]
+	readingOrderHrefs: string[]
+	chapterTitleFromToc?: string
+}): ReaderLocator | null {
+	const { selectionLocator, selectedText, positions, readingOrderHrefs, chapterTitleFromToc } = args
```
aaronleopold: I usually like to avoid inlining types like this, and since you are deconstructing anyways I'd say something like this might be cleaner:


```suggestion
type EnrichSelectionLocatorArgs = {
	selectionLocator?: Locator | ReaderLocator | null
	selectedText: string
	positions: Locator[]
	readingOrderHrefs: string[]
	chapterTitleFromToc?: string
}

export function enrichSelectionLocator({
	selectionLocator,
	selectedText,
	positions,
	readingOrderHrefs,
	chapterTitleFromToc,
}: EnrichSelectionLocatorArgs): ReaderLocator | null {
```

packages/browser/src/components/readers/epub/annotations/SelectionToolbar.tsx
```diff
@@ -0,0 +1,54 @@
+import { cx } from '@stump/components'
```
aaronleopold: We should use `cn` since I see `className` being provided as overrides, otherwise it will not properly normalize classes

packages/browser/src/components/readers/epub/annotations/locator.ts
aaronleopold: The ternaries here are also a bit much to read, I understand it largely comes from slightly incongruent types `Locator | ReaderLocator` (class vs obj lit) but I think creating some helper type like `LocatorLike` and doing a cast might be really helpful here

core/src/opds/v2_0/progression.rs
```diff
@@ -44,40 +44,72 @@ impl OPDSProgression {
...
+						total_progression: percentage_completed.or_else(|| {
+							Some(current_page as f64 / data.book.pages as f64)
+						}),
+						..Default::default()
+					};
+					let _type = OPDSLinkType::ImageJpeg;
+					(Some(title), Some(href), Some(_type), Some(locations))
+				},
+				_ => (None, None, None, None),
+			}
```
aaronleopold: I found this really difficult to read, we've reached a complexity here that should probably just be split up. Offhand, maybe:

- A `new` fn on `OPDSProgressionLocator`
   - invokes e.g. `epub_progression_locator(end_locator, percentage_completed)` for epub ext
   - otherwise invokes `paged_progression_locator(...)`

packages/browser/src/components/readers/epub/readium/useReadiumNavigator.ts
```diff
@@ -0,0 +1,402 @@
...
+import {
+	attachFrameReloadGuard,
+	patchDurableIframeSrc,
+	recoverBlankFrames,
+} from './patchDurableIframeSrc'
+
+type LoadState =
+	| { status: 'idle' | 'loading' }
+	| { status: 'ready' }
+	| { status: 'error'; message: string }
```
aaronleopold: I usually avoid these shapes when I could just do:

```ts
type LoadState = {
   status: 'idle' | 'loading | 'ready' | 'error'
   error?: string
}
```

And not have to juggle the types. Or, at the very least, I don't see a reason `ready` needs to be isolated:

```ts
type LoadState =
	| { status: 'idle' | 'loading' | 'ready' }
	| { status: 'error'; message: string }

File: packages/browser/src/components/readers/epub/readium/useReadiumNavigator.ts [hunk 17]
```diff
@@ -0,0 +1,402 @@
...
+	preferences,
+	onPositionChanged,
+	onToggleControls,
+	onTextSelected,
+	onTextCleared,
+	onDecorationActivated,
+}: UseReadiumNavigatorArgs): {
+	loadState: LoadState
+	api: ReadiumNavigatorApi | null
+} {
```
aaronleopold: Nit: I like to avoid inline typing

packages/browser/src/components/readers/epub/readium/useReadiumNavigator.ts
```diff
@@ -0,0 +1,402 @@
...
+
+	useEffect(() => {
+		const nav = navigatorRef.current
+		if (!nav || loadState.status !== 'ready') return
+		void nav.submitPreferences(new EpubPreferences(preferences)).catch((err) => {
+			console.error('[useReadiumNavigator] submitPreferences failed', err)
+		})
+	}, [preferences, loadState.status])
+
+	const api: ReadiumNavigatorApi | null =
```
aaronleopold: Any reason not to make this `ReadiumNavigatorApi` and just noop the functions if the loadState is not ready?

packages/browser/src/components/readers/epub/readium/useReadiumNavigator.ts
```diff
@@ -0,0 +1,402 @@
...
+			if (isCancelled()) {
+				resolve()
+				return
+			}
+			if (el.clientWidth >= MIN_VIEWPORT_PX && el.clientHeight >= MIN_VIEWPORT_PX) {
+				resolve()
+			} else {
+				reject(new Error('EPUB viewport never received a non-zero size'))
+			}
+		}, 5000)
```
aaronleopold: Was this an arbitrary decision? If not, I'd like a comment I think, since it reads arbitrary at a glance

packages/browser/src/components/readers/epub/readium/locator.ts
```diff
@@ -0,0 +1,223 @@
...
+ * Resolve an initial open locator against the server positions list.
+ *
+ * Priority: matching stored Readium locator (by package path) → nearest
+ * totalProgression from saved percentage → undefined (navigator uses positions[0]).
+ */
+export function resolveInitialLocator(args: {
+	positions: Locator[]
+	storedLocator?: ReadiumLocator | null
+	percentageCompleted?: number | null
+}): Locator | undefined {
```
aaronleopold: More inline types, I'll stop calling them out but generally I'd like to avoid them if possible

packages/browser/src/components/readers/epub/readium/locator.ts
```diff
@@ -0,0 +1,223 @@
...
+		return aPos === bPos
+	}
+
+	const aProg = safeNumber(a.locations?.progression)
+	const bProg = safeNumber(b.locations?.progression)
+	if (aProg != null && bProg != null) {
+		return Math.abs(aProg - bProg) < 0.02
+	}
+
+	return true
```
aaronleopold: Should we be returning true as the default? If `position` and `progression` are both null, we'd flow here and return `true` which feels a bit off

packages/browser/src/components/readers/epub/annotations/useEpubAnnotations.ts
aaronleopold, line 1: We should follow tanstack guidelines for optimistic updates, instead of going the more manual route as each does in this file: https://tanstack.com/query/latest/docs/framework/react/guides/optimistic-updates#via-the-cache

packages/browser/src/components/readers/epub/annotations/useEpubAnnotations.ts
```diff
@@ -0,0 +1,302 @@
...
+		[mediaId, isIncognito, createMutation],
+	)
+
+	const updateAnnotation = useCallback(
+		async (id: string, annotationText: string | null) => {
+			if (isIncognito) return
+
+			let previous: EpubAnnotation | undefined
+			setAnnotations((prev) => {
+				previous = prev.find((annotation) => annotation.id === id)
```
aaronleopold: I think this is discouraged in React, having side-effects inside updater. But also, moving to recommendations from tanstack should remove having to do this manually

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,941 @@
...
+
+fn package_path_for_zip(path: &str) -> String {
+	path.trim_start_matches('/')
+		.split('#')
+		.next()
+		.unwrap_or(path)
+		.to_string()
+}
+
+fn find_zip_index<R: Read + std::io::Seek>(
```
aaronleopold: This reads really inefficient, since it does a full loop for each spine item and the a secon for the percent-decoded fallback. I think what might help is having a couple of hashmaps ([raw name]: index, [decoded name]: index) to trim that down

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,941 @@
...
+		.as_ref()
+		.map(|c| c.spine_index as usize)
+		.unwrap_or(0);
+	let mut start_text_offset = options
+		.cursor
+		.as_ref()
+		.map(|c| c.text_offset as usize)
+		.unwrap_or(0);
+
+	if start_spine > spine_meta.len() {
```
aaronleopold: Should this be `>=`?

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,941 @@
...
+	}
+
+	let needle = query.to_lowercase();
+	let mut results = Vec::new();
+	let mut truncated = false;
+	let mut bytes_scanned = 0usize;
+	let mut spine_items_scanned = 0usize;
+	let mut next_cursor: Option<EpubSearchCursor> = None;
+
+	for meta in spine_meta.iter().skip(start_spine) {
```
aaronleopold: If I read things correctly, `enumerate_spine_position_meta` filters out items with no resource which might throw this off, right?

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,941 @@
...
+}
+
+fn from_hex(b: u8) -> Option<u8> {
+	match b {
+		b'0'..=b'9' => Some(b - b'0'),
+		b'a'..=b'f' => Some(b - b'a' + 10),
+		b'A'..=b'F' => Some(b - b'A' + 10),
+		_ => None,
+	}
+}
```
aaronleopold: Definitely need comments on these ones

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,941 @@
...
+				build_excerpt(&plain, match_start, match_end, EPUB_SEARCH_EXCERPT_RADIUS);
+			let excerpt = format!("{before}{highlight}{after}");
+			let progression = if plain.is_empty() {
+				0.0
+			} else {
+				(match_start as f64 / plain.len() as f64).clamp(0.0, 1.0)
+			};
+			let weight = if spine_meta.iter().map(|s| s.size).sum::<usize>().max(1) == 0 {
+				0.0
+			} else {
```
aaronleopold: `num.max(1)` will never == 0

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,941 @@
...
+				meta.size as f64
+					/ spine_meta.iter().map(|s| s.size).sum::<usize>().max(1) as f64
```
aaronleopold: We should capture `spine_meta.iter().map(|s| s.size).sum::<usize>().max(1)` so we can avoid recomputing it

aaronleopold (commented):
Thank you for getting the ball rolling here! I had some findings that should be addressed, a disclaimer that I have not had time to actually try out the changes on my computer yet (but will aim to really give it a try over the weekend, hopefully)

sandmor:
Thank you for checking the PR! I have a couple of things going on right now, but will address your review in a couple days most as it becomes possible

core/src/filesystem/media/epub_cfi.rs
sandmor: The whole reason of this is KoReader, that is why I didn't dropped epub cfi, it would have broke the current koreader sync server integration that I thought was a feature (not sure if it is working, I saw it in the code). The thing is that KoReader doesn't speak Readium, it uses epubcfi. I tried to keep this to maintain whatever compatibility there is with koreader right now but if is not required it could be dropped

aaronleopold: > The thing is that KoReader doesn't speak Readium, it uses epubcfi

KoReader uses x-pointers, though, so I don't think it would actually break the feature. I would want to double check since it's been a while since looking at the integration but a quick search yielded https://github.com/koreader/koreader/discussions/9099#discussioncomment-2751292

sandmor: KoReader x-pointers are untranslatable we don't currently handle those (doing so would require to simulare crengine internal state). The discussion you link says so too. What stump currently does is what is state in apps/server/src/routers/koreader/sync.rs parse_progress comment, a best effort parse that hopes that koreader sent a CFI or a page number.

Basically in all cases we store koop's progress in `koreader_progress` column, so we can keep interacting with koop correctly. But if we can translate it (meaning it is not a x-pointer) then we update our epubcfi column so we can have the progress matching.

Importantly, this aims to conserve the current behavior where if we cannot parse the new progress value, then our stored progress is not affected, though epubcfi is still a fallback.

TL;DR x-pointers aren't supported, we pray koreader gives us something different, we keep the original value and parse what we can, epubcfi translation exist to salvage this case that I'm not sure how common is but the code handles.

core/src/filesystem/media/epub_cfi.rs
sandmor: Agreed

apps/server/src/routers/api/v2/epub.rs
sandmor: Right, I'll remove those legacy routes, nothing seems to be using them

apps/server/src/routers/api/v2/epub.rs
sandmor: I see, I thought `public_url` was actually use currently and I missed it when reviewing the code, but you are right this forces a defined domain that while would be fine for regular server service (I was thinking along the line of certain servers that allow me to define a custom domain precisely because they're mean to be behind a reverse proxy) for a server mean to be accessed sometimes through the local network (or VPN, or tailscale) that wouldn't make that much of a sense.

If `public_url` is not use them I will follow the pattern of the rest of the codebase and just use `format!("{}/api/v2/epub/{}", host_details.url(), id)` here

core/src/filesystem/media/epub_cfi.rs
aaronleopold: Yeah we can't do anything with x-pointers really, and 

> What stump currently does is what is state in apps/server/src/routers/koreader/sync.rs parse_progress comment, a best effort parse that hopes that koreader sent a CFI or a page number

Exactly, and if that discussion is still true today (it is a few years dated) then I think we are effectively guaranteed to not get anything actually parseable as a CFI string. I do recall (vaguely, I wrote that quite some time ago now) that I added it as a hopeful shim until I could confirm more confidently.

> TL;DR x-pointers aren't supported, we pray koreader gives us something different, we keep the original value and parse what we can, epubcfi translation exist to salvage this case that I'm not sure how common is but the code handles

Let's just remove the prayers and kill the CFI support, if that sounds alright? I think folks using integrations like KoReader or Kobo likely aren't using Stump as a reader anyways, so keeping this will just mean more work with little actual gain from my perspective

crates/models/src/services/reading_progress.rs
sandmor: Alright, I'll remove it and drop CFI

core/src/filesystem/media/epub_search.rs
sandmor: `= len` would just do an empty search that I thought was consistent, but I'm reconsidering changing it

core/src/filesystem/media/epub_search.rs
sandmor: Yes, it is filtered. This is a genuine problem, I'll just make it use the full spine index as in OPF for consistency (so unfiltered)

core/src/filesystem/media/epub_search.rs
sandmor: AI, sorry. I should have reviewed that better

sandmor:
@aaronleopold This should address the requested changes. Mostly was about removing epub cfi, and the other changes requested, I found that yarn no longer resolves multiple react versions (something that I encountered during development) and I fixed some glaring mistakes in epub search that addressed what you flawed and my own concern (reimplementing url encoding, unescaping and hex parsing).

For internationalization I added the hooks needed and update `en-US`. I figured that keeping it minimal for this PR was best, and most other languages would have required me to machine translate.

aaronleopold:
Thank you, @sandmor! I'll try to review the updates sometime before Sunday, I still haven't had a moment to actually pull it down and give it a go, so will do that as well.

> For internationalization I added the hooks needed and update en-US. I figured that keeping it minimal for this PR was best, and most other languages would have required me to machine translate

Yeah that's exactly right, sorry I wasn't overly clear but we only ever need to manually update `en-US` since that is the base file which the others are automatically updated against via Crowdin

apps/server/src/routers/api/v2/epub.rs
```diff
@@ -1,90 +1,207 @@
...
+		.filter(media::Column::Id.eq(id.to_string()))
+		.into_model::<media::MediaIdentSelect>()
+		.one(conn)
+		.await?
+		.ok_or_else(|| APIError::NotFound("Book not found".to_string()))
+}
+
+/// Resolve the request-scoped base URL used in RWPM absolute resource links.
+fn epub_service_base_url(host_details: &HostDetails, id: &str) -> String {
+	host_details.url() + "/api/v2/epub/" + id
```
aaronleopold: There should exist url-specific concat tools somewhere I've added in the past, and if so can we use it? I'd worry about `//` here if `url()` ends in a trailing slash, otherwise. I _think_ OPDS V2 has a good amount of this kinda thing

apps/server/src/routers/api/v2/epub.rs
```diff
@@ -1,90 +1,207 @@
...
+fn search_error_to_api(error: EpubSearchError) -> APIError {
+	match error {
+		EpubSearchError::InvalidQueryLength { .. }
+		| EpubSearchError::InvalidCursor
+		| EpubSearchError::InvalidLimit { .. } => APIError::BadRequest(error.to_string()),
+		EpubSearchError::Cancelled => {
+			APIError::BadRequest("Search cancelled".to_string())
+		},
+		EpubSearchError::File(err) => APIError::from(err),
+	}
```
aaronleopold: This can probably be a `From` impl, instead. That would remove all the `map_err` usages in favor of try operators

Also, I don't think a canceled request is necessarily a bad request semantically. I don't imagine it would get too much use, but maybe a dedicated `CanceledRequest` variant on the api error?

apps/web/vite.config.ts
aaronleopold, line 1: I know you reverted some of the vite-specific changes, so think these might just be left over for clean up?

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,940 @@
...
+	#[error(transparent)]
+	File(#[from] FileError),
+}
+
+#[derive(Debug, Clone)]
+pub struct EpubSearchOptions {
+	pub query: String,
+	pub limit: usize,
+	pub cursor: Option<EpubSearchCursor>,
+	/// Overrideable caps for tests.
```
aaronleopold: Is this only for tests? If we won't be using it outside them then I'd say just remove them. I don't necessarily want to get into the habit of adding functionality just for the sake of tests

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,940 @@
...
+
+	pub fn decode_cursor(raw: &str) -> Result<EpubSearchCursor, EpubSearchError> {
+		let bytes = URL_SAFE_NO_PAD
+			.decode(raw.trim())
+			.map_err(|_| EpubSearchError::InvalidCursor)?;
+		let cursor: EpubSearchCursor =
+			serde_json::from_slice(&bytes).map_err(|_| EpubSearchError::InvalidCursor)?;
+		if cursor.v != CURSOR_VERSION {
+			return Err(EpubSearchError::InvalidCursor);
+		}
```
aaronleopold: Just to make sure I understand, `v` is just a flag we would have incremented in a scenario where a client is holding a now-outdated cursor when the server updated and the repr changed (where that increment came from)? I think we can do away with it, if that is the case:

1. We just return `InvalidCursor` if the versions are mismatched, which is the same if we, e.g., updated the repr of `EpubSearchCursor` which no longer serializes because of the change
2. The scenario itself feels unlikely

If I'm misunderstanding, though, let me know!

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,940 @@
...
+	pub v: u8,
+	pub spine_index: u32,
+	pub text_offset: u32,
+	/// Fingerprint of the normalized query so a cursor cannot be reused across searches.
+	pub query_fp: String,
+}
+
+impl EpubSearchCursor {
+	pub fn encode(&self) -> String {
+		let json = serde_json::to_vec(self).expect("cursor serialization");
```
aaronleopold: We should avoid panics in any runtime areas beyond startup

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,940 @@
...
+		URL_SAFE_NO_PAD.encode(json)
+	}
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
aaronleopold: Do we need `Deserialize` for a response-type?

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,940 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct EpubSearchResponse {
+	pub query: String,
+	pub results: Vec<EpubSearchResult>,
+	#[serde(skip_serializing_if = "Option::is_none")]
+	pub next_cursor: Option<String>,
+	pub truncated: bool,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
aaronleopold: Same as above I guess

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,940 @@
...
+		.map(|c| c.spine_index as usize)
+		.unwrap_or(0);
+	let mut start_text_offset = options
+		.cursor
+		.as_ref()
+		.map(|c| c.text_offset as usize)
+		.unwrap_or(0);
+
+	if options.cursor.is_some() && start_spine >= raw_spine_len {
+		return Err(EpubSearchError::InvalidCursor);
```
aaronleopold, line 242: I think a log here might be good to signal at least in one spot what went wrong

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,940 @@
...
+			Some(idx) => idx,
+			None => {
+				tracing::warn!(path = %zip_path, "spine resource missing from zip; skipping");
+				start_text_offset = 0;
+				continue;
+			},
+		};
+
+		let uncompressed = {
+			let file = archive.by_index(entry_index).map_err(FileError::from)?;
```
aaronleopold: This short-circuits the loop and errors, right? Do we feel it is a fatal scenario?

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,940 @@
...
+			break;
+		}
+
+		start_text_offset = 0;
+	}
+
+	// If we stopped due to scan budgets without filling the page, keep the cursor.
+	if next_cursor.is_none() && truncated {
+		// already set when budget hit
+	}
```
aaronleopold: I think we can just keep the note and remove the empty conditional block

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,940 @@
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct EpubSearchCursor {
+	pub v: u8,
+	pub spine_index: u32,
+	pub text_offset: u32,
+	/// Fingerprint of the normalized query so a cursor cannot be reused across searches.
+	pub query_fp: String,
```
aaronleopold: Coming back to this after reviewing most of the file, I'm still not 100% sure I understand what problem this aims to solve. I imagine what folks will be using search for has little chance for actual collision across cursors? And a client should trash any cursors if the search changes? Let me know if I'm totally missing something, though.

core/src/filesystem/media/epub_search.rs
```diff
@@ -0,0 +1,940 @@
...
+
+fn read_zip_entry_bounded<R: Read + std::io::Seek>(
+	archive: &mut ZipArchive<R>,
+	index: usize,
+	expected_size: usize,
+	cancel: &CancellationToken,
+) -> Result<Vec<u8>, EpubSearchError> {
+	let mut file = archive.by_index(index).map_err(FileError::from)?;
+	let mut buf = Vec::with_capacity(expected_size.min(EPUB_SEARCH_MAX_ITEM_BYTES));
+	let mut chunk = [0u8; 64 * 1024];
```
aaronleopold: Any reason we want to allocate this specific amount?

core/src/filesystem/media/readium.rs
aaronleopold, line 1: Not important right now, but I think we can find a better home for this than in `filesystem`

core/src/filesystem/media/readium.rs
```diff
@@ -0,0 +1,658 @@
...
+
+	pub fn with_rel(mut self, rel: impl Into<String>) -> Self {
+		self.rel = Some(vec![rel.into()]);
+		self
+	}
+
+	pub fn with_children(mut self, children: Vec<RWPMLink>) -> Self {
+		self.children = children;
+		self
+	}
```
aaronleopold: One thing I've used in Stump that I quite enjoyed was `derive_builder`. E.g., [here](https://github.com/stumpapp/stump/blob/main/core/src/opds/v2_0/progression.rs#L18)

core/src/opds/v2_0/progression.rs
```diff
@@ -113,6 +65,84 @@ struct OPDSProgressionLocator {
...
+			position: Some(current_page),
+			total_progression: percentage_completed
+				.or_else(|| Some(current_page as f64 / data.book.pages as f64)),
+			..Default::default()
+		};
+
+		OPDSProgressionLocatorBuilder::default()
+			.title(Some(format!("Page {current_page}")))
+			.href(Some(href))
+			._type(Some(OPDSLinkType::ImageJpeg))
```
aaronleopold, line 141: Can we retain the `TODO` that was removed?:

```
// TODO: Don't assume JPEG, use analysis to determine this
```

core/Cargo.toml
```diff
@@ -7,6 +7,7 @@ edition = "2021"
 alphanumeric-sort.workspace = true
 apalis.workspace = true
 apalis-cron.workspace = true
+axum.workspace = true
```
aaronleopold: We should avoid this one, it's been omitted intentionally up to this point. I assume it was added because the readium structs were defined in core

crates/models/src/entity/bookmark.rs
```diff
@@ -16,6 +16,7 @@ pub struct Model {
 	#[sea_orm(column_type = "Json", nullable)]
 	pub locator: Option<ReadiumLocator>,
 	#[sea_orm(column_type = "Text", nullable)]
+	#[graphql(skip)]
 	pub epubcfi: Option<String>,
```
aaronleopold: We can remove it I think

crates/models/src/entity/reading_session.rs
```diff
@@ -30,9 +30,7 @@ pub struct Model {
 	/// the "logical" date of this session, based on user prefs and start time
 	pub session_date: Date,
 
-	#[graphql(
-		deprecation = "future releases of Stump will remove support for this field"
-	)]
+	#[graphql(skip)]
 	#[sea_orm(column_type = "Text", nullable)]
 	pub epubcfi: Option<String>,
```
aaronleopold: Same as above

packages/browser/src/components/readers/epub/annotations/locator.ts
```diff
@@ -0,0 +1,155 @@
...
+}
+
+const TEXT_CONTEXT_LENGTH = 32
+
+/**
+ * Readium's web navigator does not populate `text.before` / `text.after` on the
+ * `Locator` returned by the `textSelected` event — only `text.highlight` is set.
+ * Without surrounding context, Readium's `TextQuoteAnchor.matchQuote()` cannot
+ * disambiguate repeated text and always anchors to the first occurrence in the
+ * chapter.
```
aaronleopold, line 98: Oh this is super interesting, I wonder why

packages/browser/src/components/readers/epub/readium/ReadiumWebReader.tsx
```diff
@@ -0,0 +1,922 @@
...
+		],
+	)
+
+	// Open the publication once per book id
+	useEffect(() => {
+		const abort = new AbortController()
+		setOpenState({ status: 'loading' })
+
+		void (async () => {
+			try {
```
aaronleopold, line 277: I generally avoid iife's

packages/browser/src/components/readers/epub/readium/ReadiumWebReader.tsx
```diff
@@ -0,0 +1,922 @@
...
+						: null,
+					text: input.text
+						? {
+								after: input.text.after,
+								before: input.text.before,
+								highlight: input.text.highlight,
+							}
+						: null,
+				}
+			: null
```
aaronleopold: I'm not sure I'm following the reason for this round trip

packages/browser/src/components/readers/epub/readium/ReadiumWebReader.tsx
```diff
@@ -0,0 +1,922 @@
...
+		(section: number) => {
+			const item = opened?.publication.manifest.readingOrder?.items?.[section]
+			if (item) {
+				api.goLink(item)
+			}
+		},
+		[api, opened],
+	)
+
+	const getLocatorPreviewText = useCallback(async (locator: ReaderLocator) => {
```
aaronleopold: I think this doesn't need async:


```suggestion
	const getLocatorPreviewText = useCallback((locator: ReaderLocator) => {
```

packages/client/src/hooks/useGraphQL.ts
aaronleopold, line 1: Nice, thank you!

packages/browser/package.json
```diff
@@ -21,6 +21,9 @@
 		"@dnd-kit/sortable": "^8.0.0",
 		"@hookform/resolvers": "^3.9.0",
 		"@panzoom/panzoom": "^4.6.0",
+		"@readium/navigator": "2.6.1",
+		"@readium/navigator-html-injectables": "2.5.0",
+		"@readium/shared": "2.2.3",
```
aaronleopold: While we are here, the latest looks to be:

```suggestion
		"@readium/navigator": "2.8.2",
		"@readium/navigator-html-injectables": "2.6.3",
		"@readium/shared": "2.4.0",
```

aaronleopold (commented):
Okay I finally got around to another review after your changes (thank you!) and actually ran it locally to give it a try (I also updated with main and pushed while I was there).

Nothing really major that I saw, but made some comments. A few things that are noteworthy from trying out the UI:

- Links don't seem to work (e.g., the TOC pages in the actual book). I can see it is a link, but clicking doesn't open it. It also doesn't account for local dev (i.e., port is 3000)
- Moving between chapters is notably slow, but can't quite sus out why
- I can't add annotations directly, I have to highlight then click into it then add the annotation and click save. I don't see a network call for it
- Clicking a bookmark did not take me to the spot I bookmarked, it didn't seem to do anything unless I was in a different chapter. I suspect some larger issue with bookmarks and generally location detection because the bookmark icon stayed filled the entire chapter
- The full screen button does not seem to do anything anymore
- Changing font and going back to default does not unset the font, refreshing the page fixes it

Overall I'm personally really excited to get this in. Thanks for all you've done so far!

core/src/filesystem/media/epub_search.rs
sandmor: No it is not for possible collisions between cursors. The thing is that property encodes a hash of the query string (something like 'alice' or 'susurrus') so this cursor is rejected if is used for a different query. This is because theoretically a cursor for a client is just an opaque base64 token and the epub search whole state so this guarantees a client doesn't mix internal states. Rejecting the states from foreign searches.

This is not necessary if we trust the client will not mix those states. And given that is our own client and there is no server crashing risk apparent in case of a corrupted query this may not be needed. I suppose it is just my mindset, to me the cursor was a implementation detail that should be hidden from the api contract, but it was also a state that I didn't wanted to get mixed by error so I put it for completeness, same as the `v` field.

To keep it simple we can drop both. Or perhaps even expose spine index and text offset? It would give more control to the client but would bind that coordinate format to the protocol should it change in the future

core/src/filesystem/media/epub_search.rs
sandmor: No, it felt nice. I can put it in a constant

core/src/filesystem/media/readium.rs
sandmor: Perhaps a top-level `core/src/readium`?

aaronleopold: I think maybe somewhere in `crates/models`? There is already `crates/models/shared/readium`. Otherwise core at root is probably fine for now

core/Cargo.toml
sandmor: It was added to implement `IntoResponse`, avoiding adding `axum` to the core was also the reason I didn't originally did that. I don't think we can have both options. What should we do?

core/src/filesystem/media/epub_search.rs
aaronleopold: Ah okay, thank you for explaining it. It's already here so I don't mind if it stays, maybe if you don't mind just add a little more doc comments to explain its purpose for when I inevitably forget N months down the road?

core/Cargo.toml
aaronleopold: Hm, yeah I understand. So in practice it shouldn't matter too much, I've been a little strict about it to try and keep `core` separate from `server` since it is meant to function strictly as a library and could hypothetically be embedded into a different Rust binary that isn't a server. That said, to my knowledge, nobody is doing that so the only Rust binary shipped with the core is the server lol

If we wanted to keep that strict separation, we could feature flag axum in the core and implement `IntoResponse` when set. You could probably also wrap them in wrapper structs and impl the trait on that in `apps/server`, but that feels a little off.

If keeping that boundary is bit of a hassle then don't worry about it, it's probably fine as-is.

packages/browser/src/components/readers/epub/annotations/locator.ts
sandmor: No idea, this is one of the workarounds that I had to do because without the surrounding context we cannot effectively store a highlight or note. You can see it here, it just discards the context:

https://github.com/readium/ts-toolkit/blob/navigator/2.6.1/navigator/src/webpub/WebPubNavigator.ts#L283

https://github.com/readium/ts-toolkit/blob/navigator/2.6.1/navigator/src/epub/EpubNavigator.ts#L468

core/Cargo.toml
sandmor: Tying to moving a part of `readium` into `models`, what if we drop the `IntoResponse` impls in `core`? We can just use a wrapper structure, like `OPDSAuthDocWrapper`seems to be doing and you suggest?

core/src/filesystem/media/readium.rs
sandmor: Sure, types can go to `crates/models/shared/readium` but filesystem then would still retain `ReadiumManifestGenerator` and others rather than creating a new module?

core/Cargo.toml
aaronleopold: Yeah that works

core/src/filesystem/media/epub_search.rs
sandmor: I'll make encode() return a Result, though this conversion is practically infallible

core/src/filesystem/media/readium.rs
aaronleopold: Sounds good!

sandmor:
> Moving between chapters is notably slow, but can't quite sus out why

I can only reproduce this in Firefox, Chrome and Brave (that uses the chromium engine) work fine. Which browser are you using?

> I can't add annotations directly, I have to highlight then click into it then add the annotation and click save. I don't see a network call for it

Not sure if I get the network call part, do you mean adding a note without highlighting? That is currently not supported because I needed a way to mark a part that has a note, and I'm not sure how it would work without it. The network call that creates it is `createAnnotation`, an annotation is a highlight with an optional note attached.

As for the click issue, I cannot reproduce it either in any browser but I have an idea what is happening: perhaps your browser is removing the selection when the dialog opens, which would cause the note to fail as it cannot determine the selection. I can fix it by capturing the selection before the dialog opens

> Clicking a bookmark did not take me to the spot I bookmarked, it didn't seem to do anything unless I was in a different chapter. I suspect some larger issue with bookmarks and generally location detection because the bookmark icon stayed filled the entire chapter

Hm, again, it works for me, in firefox too. The bookmark icon staying filled seems an isolate issue of me ignoring `progression`. For navigation, it passes through readium but readium seems to not be applying it reliablily then, I'll investigate the issue

> The full screen button does not seem to do anything anymore

Never did anything, it and the location manager button were broken for me when I tried them, but I'll fix the full screen too.

aaronleopold:
> > Moving between chapters is notably slow, but can't quite sus out why
> 
> I can only reproduce this in Firefox, Chrome and Brave (that uses the chromium engine) work fine. Which browser are you using?

I use Firefox

> > I can't add annotations directly, I have to highlight then click into it then add the annotation and click save. I don't see a network call for it
> 
> Not sure if I get the network call part, do you mean adding a note without highlighting? That is currently not supported because I needed a way to mark a part that has a note, and I'm not sure how it would work without it. The network call that creates it is `createAnnotation`, an annotation is a highlight with an optional note attached.

For the network call part, I really just meant I assume it wasn't wired up since hitting highlight did not ping the server to persist it. I don't mean a note without highlight, just a normal highlight is all I mean.

> > Clicking a bookmark did not take me to the spot I bookmarked, it didn't seem to do anything unless I was in a different chapter
> 
> Hm, again, it works for me, in firefox too. The bookmark icon staying filled seems an isolate issue of me ignoring `progression`. For navigation, it passes through readium but readium seems to not be applying it reliablily then, I'll investigate the issue

Oh interesting, yeah not sure what is happening then. In fairness I did not try any other browsers

> > The full screen button does not seem to do anything anymore
> 
> Never did anything, it and the location manager button were broken for me when I tried them, but I'll fix the full screen too

No worries if unrelated to this diff, the web reader has definitely lagged so isn't surprising if there are preexisting issues

sandmor:
That should fix all the problems, if it doesn't work please tell me. The most complex ones where the bookmarks and the firefox slow chapter transition.

For the bookmarks there were multiple possible issues. Readium doesn't accept navigation if there is one already in flight, so I made navigation async and queued. The bookmark handler then wait for the navigation result and only then closes the panel. Hopefully this fixes the problem in your end, tell me if it still doesn't work!

For the subpar performance on firefox I profiled what was happening and found that Readium creates the new iframe as hidden. Its `ColumnSnapper` handles both focus and go_progression using requestAnimationFrame, and Firefox throttles animation frames inside an `iframe` whose visibility is hidden to approximately one frame per second (that chromium doesn't) and as readium waits for the first frame before revealing the `iframe`, we have:

hidden iframe ->
  focus waits ~1 second
  progression waits ~1 second
  `iframe` becomes visible

I fixed it by adding another patch to readium in our codebase to make the `iframe` technically visible but fully transparent, non-interactive and aria-hidden. That fixed it.

EDIT: Also, I couldn't find a suitable utility for handling URLs in the codebase as you wanted so I added a small method to `HostDetails` that addresses the double `//` problem

aaronleopold:
I'm going to try to give your changes another look before the end of the weekend, thank you!! And also thank you for figuring out and explaining that iframe bug

> I fixed it by adding another patch to readium in our codebase to make the iframe technically visible but fully transparent, non-interactive and aria-hidden. That fixed it

At some point it would probably also be good to get that upstream to them, too, so eventually we can do away with the patch (if accepted)

aaronleopold (approved):
I reviewed this during my lunch hour today and gave it another spin locally. Once the conflicts are resolved, I think this is in a good spot to get into the hands of `nightly` testers :)

Thanks for all your time and effort!

sandmor:
It is nothing, thanks to you for this awesome project

PR #1295, opened 2026-07-20
:memo: Add justRead to OPDS 1.2 tested clients

Jachym-Jahoda:
justRead is a native iOS/iPadOS EPUB, PDF, and audiobook reader with a custom OPDS server field, so it can browse and download from a Stump server's OPDS 1.2 catalog. I haven't run it against page-streaming specifically (it's a general reader, not a comic-focused client), so I've marked that column No rather than claim untested behavior. Happy to update with real findings if I get a chance to test more thoroughly.

App Store: https://apps.apple.com/us/app/epub-reader-justread-app/id6757948605
Website: https://justread.app


aaronleopold (approved):
Thanks for sharing

PR #1324, opened 2026-08-13
:passport_control: Invert library access control model (whitelist)

aaronleopold:
Relates to #1151

## Description

Does not close linked issue, although I am working towards initial support for claims-based sync of permissions which should fully resolve it

- Replace `library_exclusions` with `library_access` table
- Invert all filters relating to library access accordingly
- Fix tests which previously relied on libraries being accessible by default (a lot)
- Add CLI subcommand for basic management of library access
- Misc changes to make base branch compile after previous merge with nightly


crates/cli/src/commands/library.rs
```diff
@@ -0,0 +1,156 @@
...
+		return Err(CliError::OperationFailed(
+			"Library does not exist".to_string(),
+		));
+	};
+
+	let users = user::Entity::find().all(&conn).await?;
+	let user_username_to_id = users
+		.iter()
+		.map(|user| (user.username.clone(), user.id.clone()))
+		.collect::<std::collections::HashMap<String, String>>();
```
Kernald: Tiny nit: there should probably be a `use` statement for `HashMap` rather than the fully qualified name here and below

crates/cli/src/commands/library.rs
```diff
@@ -0,0 +1,156 @@
...
+	progress.set_message("Updating access for library...");
+
+	let txn = conn.begin().await?;
+
+	// easier to just rm all access then re-add the selected ones
+	let current_access = library_access::Entity::find()
+		.filter(library_access::Column::LibraryId.eq(library_id.clone()))
+		.all(&conn)
+		.await?;
+	for access in current_access {
```
Kernald: I don't have the `library_access` schema in mind but if there's any timestamp in there logging when a user got granted access, this makes it less reliable, the logic to avoid deleting users who keep access is pretty trivial. If there's no timestamp anywhere though, agreed, it's a bit easier to just rewrite everything.

crates/graphql/src/mutation/library.rs
```diff
@@ -564,16 +573,18 @@ impl LibraryMutation {
...
-	/// removed from the exclusion list if they were previously excluded.
+	// TODO(permissions): should probably have non-full update mutations for this too eventually
+
+	/// Grant users access to a library. This operates as a full replacement of the
+	/// granted users list, so any users not included in the provided list will have
+	/// their access revoked if they were previously granted access
 	///
-	/// The server owner cannot be excluded from a library, nor can the user performing the action exclude
-	/// themselves.
+	/// The server owner cannot have their access modified, nor can the user performing
```
Kernald: I'm assuming the second part of this is to avoid ending up with libraries with no  users? Would it maybe make more sense preventing removing the last user instead? That way, someone can set up a library for some users, and then remove themselves if they're not interested (and an orphan library is always recoverable through the CLI anyway)

Kernald (commented):
Just a couple notes, looks great overall!

crates/graphql/src/mutation/library.rs
aaronleopold: I think the main reason is that it felt to me like an awkward thing to support on the UI (e.g., I deselect myself and lose immediate access). I think changing it to your suggestion could work if I make the UI not update automatically on a debounce and maybe add a confirm if you are removing access for yourself

aaronleopold: Done :) You can now revoke your own access, it will just show a confirmation modal before saving to confirm that was intentional. While I was there I also just ripped out the super admin guards, I know there will be a final sweet after fully removing the field but think it's fine to do now also where possible

PR #1318, opened 2026-08-08
🌐 Expand Korean localization coverage across web and mobile

devy1540:
## Description

This PR expands Korean localization coverage across the browser and Expo applications.

- Replaces remaining hardcoded user-facing strings with locale keys across navigation, libraries, books, readers, smart lists, book clubs, server settings, user management, OPDS, and shared controls.
- Adds the corresponding `en-US` source strings and complete `ko-KR` translations while preserving 1:1 locale key parity.
- Localizes validation messages, status text, accessibility labels, notifications, and count-dependent copy.
- Moves `LocaleProvider` above the browser error boundary so startup failures render translated fallback copy instead of raw keys such as `errorScene.criticalHeading`.
- Adds or updates regression coverage for the translated error fallback, reader controls, and API key inspection.

The motivation is to remove mixed English/Korean UI and untranslated raw keys from the Korean experience. The reported error fallback occurred because the locale provider was mounted inside the router while the error boundary was outside it; an error thrown during router startup therefore rendered without locale context.

This affects visible UI copy and localization wiring in the browser and Expo clients. It does not change backend APIs or data models.

This contribution was developed with assistance from OpenAI Codex. I reviewed the changes and validated them locally before submission.

How to contribute: https://github.com/stumpapp/stump/blob/main/.github/CONTRIBUTING.md

Validation performed:

- Locale parity: `en-US` 2,770 keys, `ko-KR` 2,770 keys, 0 missing and 0 extra keys.
- Browser tests: 5 suites and 12 tests passed.
- `corepack yarn workspace @stump/browser check-types`
- `corepack yarn workspace @stump/i18n check-types`
- ESLint and Prettier checks for the error-boundary regression fix.
- `git diff --check`

## Screenshots

Not included in this draft. The localized error fallback behavior is covered by a regression test that selects `ko-KR`, triggers the app error boundary, and verifies the Korean heading and home link.

## Ready?

Please read each item and check the boxes:

- [x] I read the [contributing guidelines](https://github.com/stumpapp/stump/blob/main/.github/CONTRIBUTING.md)
- [x] I searched for existing issues or pull requests that may be related to my contribution
- [x] This PR is based into `nightly` and not `main`
- [x] I added tests and/or documentation for my changes if applicable
- [x] I disclosed any use of LLMs in the creation of this PR (if applicable)

## Stump Contributor License Agreement

By contributing to Stump, you agree that your contributions will be licensed under the following licenses (where applicable):

[...]

PR #1342, opened 2026-08-24
fix(i18n): localize the error fallback

devy1540:
## Description

This is the first focused replacement PR extracted from #1318 after the maintainer requested smaller, independently reviewable changes.

## 목적

Make the browser error fallback honor the selected locale, including errors thrown before the application router renders.

## 내용(의도 포함)

- Move LocaleProvider above ErrorBoundary so the fallback remains inside the locale context.
- Replace the hard-coded error fallback copy with existing and newly added errorScene source keys.
- Add only en-US source strings so translations can continue through Crowdin.
- Add a regression test that selects de-DE and verifies translated fallback content is rendered.
- Keep this PR limited to the browser error fallback; it does not include broader web, mobile, book club, or desktop localization work.

## 성공기준

- The selected locale is available when ErrorFallback renders.
- Browser tests: 23 suites and 171 tests passed.
- Browser and i18n type checks passed.
- Browser ESLint completed with 0 errors; the existing 16 unrelated warnings remain.
- Prettier and git diff --check passed.

## Ready?

- [x] I read the contributing guidelines.
- [x] I searched for related issues and pull requests; this work is extracted from #1318.
- [x] This PR targets nightly rather than main.
- [x] I added a regression test for the change.

By contributing to Stump, I agree that these changes will be licensed under the applicable repository license.

### DIFF
--- a/packages/browser/src/App.tsx
+++ b/packages/browser/src/App.tsx
@@ -2,6 +2,7 @@ import './styles/index.css'
 import '@stump/components/styles/overrides.css'
 
 import { SDKProvider, StumpClientContextProvider, StumpClientProps } from '@stump/client'
+import { type AllowedLocale, LocaleProvider } from '@stump/i18n'
 import { ReactQueryDevtools } from '@tanstack/react-query-devtools'
 import { useEffect, useState } from 'react'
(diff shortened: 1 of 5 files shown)


aaronleopold (approved):
thank you!!

PR #1347, opened 2026-08-27
feat(i18n): localize thumbnail selectors

devy1540:
## 목적
Localize the shared thumbnail upload flow and the book and series thumbnail selectors as a focused follow-up to closed PR #1318.

## 내용(의도 포함)
- Add focused `thumbnailSelector` source keys while keeping existing common and dropdown keys where they are semantically appropriate.
- Localize upload validation, selector descriptions, actions, and mutation failure messages.
- Remove the mutable `20MB max` value from translatable copy without changing the actual 20 MiB upload limit.
- Add regression tests for the oversized-file path and locale-context usage in the series selector.
- Keep library thumbnail management, translated locale files, mobile, desktop, book club, and `StumpRouter` changes out of scope.

This follows the namespace and review-scope feedback from closed PR #1318 in a smaller, independently reviewable change. English source strings can be translated through Weblate after merge.

## 성공기준
- Browser tests: 38 files and 246 tests passed.
- Browser and i18n type checks passed.
- Browser lint completed with 0 errors; 17 existing warnings remain outside the changed files.
- `git diff --check` passed.
- Target hard-coded strings no longer remain in the three changed production components.
- Layout, DOM order, state transitions, GraphQL mutations, and upload behavior remain unchanged by source inspection.
- Authenticated manual browser rendering was not performed.

### DIFF
--- a/packages/browser/src/components/thumbnail/__tests__/UploadImageModal.test.tsx
+++ b/packages/browser/src/components/thumbnail/__tests__/UploadImageModal.test.tsx
@@ -1,0 +1,51 @@
+import '@/__mocks__/resizeObserver'
+
+import { LocaleProvider } from '@stump/i18n'
+import { fireEvent, render, screen, waitFor } from '@testing-library/react'
+import { toast } from 'sonner'
+
+import UploadImageModal from '../UploadImageModal'
+
+vi.mock('sonner', () => ({
+	toast: { error: vi.fn() },
+}))
+
+describe('UploadImageModal', () => {
+	beforeEach(() => {
+		vi.clearAllMocks()
+	})
+
+	it('shows the localized error when a dropped image exceeds the maximum size', async () => {
+		render(
+			<LocaleProvider locale="en-US">
+				<UploadImageModal isOpen onClose={vi.fn()} onUploadImage={vi.fn()} />
+			</LocaleProvider>,
+		)
+
+		const dropTarget = screen.getByText('Drop image here or click to select').parentElement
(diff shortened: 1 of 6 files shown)


packages/browser/src/components/thumbnail/__tests__/UploadImageModal.test.tsx
aaronleopold, line 1: nice, thank you for adding this

packages/browser/src/scenes/series/tabs/settings/__tests__/SeriesThumbnailSelector.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+vi.mock('../../../../book/settings/BookPageGrid', () => ({
+	default: () => <div />,
+}))
+
+vi.mock('../SeriesBookGrid', () => ({
+	default: () => <div />,
+}))
+
+describe('SeriesThumbnailSelector', () => {
+	it('renders localized selector content after opening the dialog', () => {
```
aaronleopold, line 50: this is totally fine for now, but i think in the future tests can mostly assert functional behavior instead. if we invoke the translate function in a component, asserting it gets the locale key isn't overly helpful since it is pretty much a guarantee. however, testing things like "i clicked X and expect Y behavior" is more along the lines of what we want to test (like your upload modal test that adds asserts to the validation of size, etc).

aaronleopold (approved):
thank you!

Issue #1263, opened 2026-06-22
[FEATURE] Remember browsing preferences (sorting/image alignment) in the browser


qinghon:
### Is your feature request related to a problem?

hello, this app is great and very lightweight

I found some customary problems when using it: 

the library cannot remember the selected sorting method, even temporarily; 

reading preferences cannot be remembered, such as the padding alignment of pictures.

### Describe the solution you'd like

I think it is possible to remember the currently selected sorting scheme and reading preference in the browser's local storage, while remembering the current screen ratio, example:

```json
{
  "hash(library name)": {
       "<screen_radio>": {
          "sort": {"filed": "create_at", "dsc": true},
          "view": {"preferences key": "value"}
       }
    } 
}
```

### Where would this feature be implemented?

- [x] Web App
- [ ] Mobile App
- [ ] Server
- [ ] Documentation

ajselzilic:
Hey, I can take this one

aaronleopold:
@ajselzilic Yeah go for it if you'd like. A few notes after looking back at the original requester's example JSON block:

- I wouldn't hash the library name for a key, we have stable idents already (i.e., the `id`)
- We already have structures for modeling filters and ordering, I don't see a reason to change the shape for storage only to convert it
- I don't know what would go in that proposed `view` honestly that isn't already persisted somewhere, but could be missing something there
- A lot of stores already exist and use [zustand](https://www.npmjs.com/package/zustand), so I imagine we can keep that going

PR #1376, opened 2026-09-03
✨ Persist browser ordering preferences

ajselzilic:
Closes https://github.com/stumpapp/stump/issues/1263

Remembers sorting for book and series views using the existing Zustand layout stores. Sorting specified in the URL still takes precedence.

Tested with typechecks, linting, formatting checks, and a local browser smoke test.


aaronleopold:
Just dropping a note that the failing test isn't from your changes, a [fix](https://github.com/stumpapp/stump/pull/1375/changes/3422534b431fd805ff7726fd1e7af672db34e10e) will go to nightly soon. I'll try to review your PR sometime in the next day or so, thank you!

aaronleopold (approved):
Thank you for knocking this out! I updated with `nightly` and tested a bit locally, things look good to me. I had a thought that URL syncing feels less important now that the primary driver is persisted (with URL as override) but think that can be a future problem (if at all).

Issue #619, opened 2025-02-26
[BUG?] Changing thumbnail for series does not stick


Cantello:
## Description of the Bug

I change the thumbnail of a PDF series (RPG books) to a thumbnail of one of the books, thumbnail reverts to previous version.

## Expected Behavior

Thumbnail remains the one last selected.

## To Reproduce

1. Open library
2. Open series
3. Select series settings
4. Choose thumbnail via 'select thumbnail'
5. Choose book
6. Choose page, watch as the thumbnail in the blurred background changes
7. Press 'confirm selection' and watch as thumbnail reverts to the previous one

## Screenshots

![Image](https://github.com/user-attachments/assets/a5f90447-ebfc-46b8-8cbe-10d7d1a70b19)

## Debugging Information

Last few log lines with debugging switched on: https://pub.microbin.eu/upload/snail-wolf-mole

**Environment:**

- OS: linux, oracle VPS
- Device: accessing via firefox browser on a windows computer

**Build Details:**

- Version: 0.0.10, latest
- Docker: yes

**Additional context**

aaronleopold:
I believe this is a caching issue, see also https://github.com/stumpapp/stump/issues/546. Can you confirm by disabling cache in your network tab and refreshing?

This is definitely still an issue, to be clear, even if it is just caching. Ideally Stump would tell the browser to update the cache for that image

Cantello:
I tried to disable caching but the behaviour still persists, see recording...

![Image](https://github.com/user-attachments/assets/1983674c-09c0-4cd7-bfb5-9ec53c1fea15)

aaronleopold:
@Cantello I believe a fix should have landed in the `nightly` image. I'll keep this issue open until it goes to a versioned release, but if you're able to verify before let me know

Cantello:
> [@Cantello](https://github.com/Cantello) I believe a fix should have landed in the `nightly` image. I'll keep this issue open until it goes to a versioned release, but if you're able to verify before let me know

I just tried with the nightly docker version and it works. I had to hit reload and clear the cache several times but after that it displayed the correct thumbnail. Thanks!

Issue #546, opened 2024-12-29
[BUG] Changing library or series thumbnail to a contained book's page fails


miasma13:
## Description of the Bug

When changing library or series thumbnail by picking a page from one of the contained issues fails. Picker shows selection of all pages but when confirming the selection the thumbnail updates for a split second then reverts to the initial one.
Uploading a custom thumbnail image updates it immediately.

## Expected Behavior

Selecting page properly updates library or series thumbnail.

## To Reproduce

1. Edit library or series thumbnail 
2. Select "From books"
3. Pick any of the pages from the picker

**Build Details:**

- Version: 0.0.9
- Docker: yes

aaronleopold:
Have you confirmed the issue isn't related to caching? E.g. purge your browser cache after changing the thumbnail?

aaronleopold:
I cannot replicate this, the thumbnail changes after busting the cache, so I'm closing this issue. Feel free to reach back out if that does not work for you.

PR #1380, opened 2026-09-04
 🐛 Fix stale thumbnail caching

### LINKED ISSUE #913: [BUG] Cover change on the Filesystem Level not reflected in WebUI
### Description of the Bug

When the embedded cover of an EPUB is changed and then the previously present files replaced, the old cover will stay in the WebUI. Those will be retained through any rebuild scans till the browser cache is cleared.

### Expected Behavior

Having the custom scan option "Force rebuild" also clear the cache. And displaying the new covers.

Maybe implementing a new custom scan option, which specifically deletes the cache.

### Steps To Reproduce

1. Have a library build with covers
2. Change the cover in the EPUB
3. Replace EPUB
4. Rescan and see no change

- [x] Web App
- [ ] Mobile App
- [ ] Server

#### Server Details:
  - Release Channel: experimental
  - Server OS: Arch Linux Kernel 6.18.5-arch1-1
  - Using Docker: Yes
  - Server Version: [7a24fdb](https://github.com/stumpapp/stump/commit/7a24fdb)

#### Client Details:
  - Device: Asus Zenfone 10 
  - OS: [Not Applicable
  - Browser: Floorp

### Additional context

The mobile application has the option to clear the cache, so the new covers are displayed after clearing.

ajselzilic:
Closes #913

Related to #546 and #619

Regenerates Stump-generated thumbnails during forced library rebuilds and adds modification timestamps to media, series, and library thumbnail URLs. When a thumbnail changes, the related timestamp changes with it, giving the browser a new URL and bypassing the existing long-lived cache without disabling caching.

Relevant browser queries are refreshed after thumbnail uploads, selections, deletions, scans, and generation jobs. Thumbnail consumers, including tables and the File Explorer, now use the versioned URLs returned by GraphQL instead of rebuilding unversioned URLs locally.

New generated thumbnails include a .generated filename marker so Stump can distinguish them from custom uploads. This allows forced rebuilds to replace generated thumbnails without overwriting user-provided covers. A filename marker was chosen instead of a database migration because the distinction is only needed while managing the files, and existing thumbnails cannot be reliably classified after the fact. Existing unmarked thumbnails are therefore treated as custom or legacy files, while newly generated thumbnails can be safely replaced going forward.

While tracing the full thumbnail flow, this also fixes:

- Library and series thumbnail selections updating the source book instead of the intended entity.
- Thumbnail mutations returning stale entity data after an update.
- Library thumbnail deletion leaving stale paths and metadata in the database.
- Authenticated images displaying an outdated result when their source changes during an active request.
- Failed authenticated image requests leaving the image without a fallback.

Library thumbnail cleanup now clears its database state transactionally. It also uses relational queries instead of passing every series and media ID as a separate SQL parameter, avoiding SQLite bind limits on large libraries.

### DIFF
--- a/core/src/filesystem/image/thumbnail/generate.rs
+++ b/core/src/filesystem/image/thumbnail/generate.rs
@@ -1,5 +1,6 @@
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
+use chrono::Utc;
 use futures::{stream::FuturesUnordered, StreamExt};
 use models::{
 	entity::{library, media, series},
@@ -9,8 +10,8 @@ use models::{
 	},
 };
 use sea_orm::{
-	prelude::*, sea_query::Query, EntityTrait, Order, QueryFilter, QueryOrder,
-	QuerySelect,
+	prelude::*, sea_query::Query, ConnectionTrait, EntityTrait, Order, QueryFilter,
+	QueryOrder, QuerySelect,
 };
 use tokio::{fs, sync::oneshot, task::spawn_blocking};
 
@@ -39,7 +40,7 @@ pub enum ThumbnailGenerateError {
 	ProcessorError(#[from] ProcessorError),
 	#[error("Did not receive thumbnail generation result")]
 	ResultNeverReceived,
-	#[error("Failed to update media entity")]
+	#[error("Failed to update thumbnail entity")]
 	UpdateFailed,
 	#[error("A candidate source for thumbnail generation could not be found")]
 	NothingToGenerate,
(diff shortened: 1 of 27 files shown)


aaronleopold:
Hey, just wanted to say this has not slipped my radar. I'm hoping to have time closer to or over the weekend to give this a proper look

core/src/filesystem/image/thumbnail/generate.rs
```diff
@@ -39,7 +40,7 @@ pub enum ThumbnailGenerateError {
 	ProcessorError(#[from] ProcessorError),
 	#[error("Did not receive thumbnail generation result")]
 	ResultNeverReceived,
-	#[error("Failed to update media entity")]
+	#[error("Failed to update thumbnail entity")]
```
aaronleopold, line 43: thank you

core/src/filesystem/image/thumbnail/generate.rs
```diff
@@ -59,7 +60,87 @@ pub struct GenerateThumbnailOptions {
...
+	pub is_custom: bool,
+	pub target: ThumbnailTarget,
+}
+
+#[derive(Debug, Clone)]
+pub enum ThumbnailTarget {
+	Media,
+	Series(String),
+	Library(String),
+}
```
aaronleopold, line 72: I'm not sure I understand the move away from `filename` and towards an enum, rather than just making the callers set `filename` accordingly (e.g., `mediaIdaryid.generated` for a generated one or `mediaId` for custom).

I guess coming back to this after a little more review, it's intention is for the update clause. I think we can get away without the `is_custom` flag, though. I don't see why a custom thumbnail would even pass through the generate flow.

Separately, I think my decision to include database updates in the `generate_book_thumbnail` is one I am currently questioning. If we had e.g. `generate_library_thumbnail` that calls (renamed) `generate_thumbnail_from_book` then that library-specific one can handle the update call and be a bit cleaner to follow. Then the same would follow for `generate_series_thumbnail`/`generate_book_thumbnail` calls `generate_thumbnail_from_book` and handles their own update.

core/src/filesystem/image/thumbnail/generate.rs
```diff
@@ -188,26 +281,74 @@ pub async fn generate_book_thumbnail(
...
+	if is_custom {
+		match &target {
+			ThumbnailTarget::Media => {
+				bump_media_thumbnail_fallbacks(conn, Some(&book.series_id)).await?
+			},
+			ThumbnailTarget::Series(id) => {
+				bump_series_thumbnail_fallbacks(conn, std::slice::from_ref(id)).await?
+			},
+			ThumbnailTarget::Library(_) => {},
+		}
```
aaronleopold, line 348: Same as above re: why would a custom thumb flow through to the generate flow?

core/src/filesystem/image/thumbnail/generate.rs
```diff
@@ -59,7 +60,87 @@ pub struct GenerateThumbnailOptions {
...
+	Library(String),
+}
+
+fn is_generated_thumbnail(path: &Path) -> bool {
+	path.file_stem()
+		.and_then(|stem| stem.to_str())
+		.is_some_and(|stem| stem.ends_with(".generated"))
+}
+
+pub async fn bump_media_thumbnail_fallbacks<C>(
```
aaronleopold, line 80: I understand why these bumps were added, but I don't _love_ how it works around an awkward point (image fallbacks and cache-friendly URLs). I think it's fine for now, just being a bit pedantic

packages/browser/src/components/entity/AuthImage.tsx
```diff
@@ -21,32 +27,48 @@ export const AuthImage = forwardRef<HTMLImageElement, Props>(({ token, src, ...p
...
+					if (active) {
+						setLoadedImage({ src, token, url: URL.createObjectURL(data) })
+					}
+				})
+				.catch((error) => {
+					if (active) {
+						setLoadedImage({ src, token, url: src })
+						console.error('Failed to load authenticated image:', error)
+					}
+				})
```
aaronleopold, line 54: I generally prefer async/await, which is why originally I handled the local state in the `fetchImage` callback. If it was refactored for the `active` I think react-query has utilities to cancel in-flight queries instead

aaronleopold (commented):
I had a few comments, but otherwise I think this is on the right track. Thanks for working on this! Let me know if you have any questions

ajselzilic:
Thanks for the feedback. I separated thumbnail generation from the entity updates and removed ThumbnailTarget and is_custom; callers now choose the filename. I also switched AuthImage back to async/await and kept the local guard, since cancelling a query shared by URL could affect other images.

aaronleopold (approved):
Thank you!

Issue #1388, opened 2026-09-04
Weblate: enable language-specific plural forms


ajselzilic:
While translating Bosnian, I noticed that `epubReader.controls.columnCount_one` and `columnCount_other` appear as separate strings, with no way to translate the `few` form.

Bosnian needs:

- `one`: `{{count}} kolona`
- `few`: `{{count}} kolone`
- `other`: `{{count}} kolona`

The Weblate component currently uses `json-nested`. Could it be switched to [i18next JSON v4](https://docs.weblate.org/en/latest/formats/i18next.html), which supports language-specific plurals?

This should let translators provide the missing forms directly in Weblate.

aaronleopold:
I've updated the setting in Weblate to use `i18next JSON v4`, no code changes should be required afaik. I did add the new language in https://github.com/stumpapp/stump/pull/1390 so it can be selected in the apps

PR #1390, opened 2026-09-05
:globe_with_meridians: Add Bosnian locale as option

aaronleopold:
Relates(ish) to https://github.com/stumpapp/stump/issues/1388

## Description

A new language was added via Weblate and this PR makes it actually registered in the app

### DIFF
--- a/packages/i18n/src/dateFnsLocale.ts
+++ b/packages/i18n/src/dateFnsLocale.ts
@@ -5,6 +5,7 @@ import type { AllowedLocale } from './config'
 
 // Note: lazy loading as to not bloat initial bundle with all the locales
 const dateFnsLocaleLoaders: Record<AllowedLocale, () => Promise<Locale>> = {
+	bs: () => import('date-fns/locale/bs').then((m) => m.bs),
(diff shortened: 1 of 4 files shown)


packages/i18n/src/index.ts
```diff
@@ -12,6 +12,7 @@ export { default as LocaleProvider } from './LocaleProvider'
 export type { AllowedLocale }
 
 export const localeNames: Record<AllowedLocale, string> = {
+	bs: 'Bosanski',
```
aaronleopold, line 15: A cursory search showed that this would be the spelling for the native word but please correct me if the preference is something else

packages/i18n/src/dateFnsLocale.ts
```diff
@@ -227,6 +229,7 @@ export function formatNarrowDuration(
 }
 
 const LOCALE_NARROW_UNITS = {
+	bs: { h: ' h', m: ' min', s: ' s' },
```
aaronleopold, line 232: Same here if the preference would be different units

PR #1391, opened 2026-09-06
:sparkles: Improve epub thumbnail selection algorithm

hal8174:
## Description

Improve the thumbnail generation for EPUB files by finding better covers.
Parse the chapters of the EPUB file in reading order to find the first image in the book and select it as the cover.

This is my first contribution in this repository, and I am open to feedback and suggestions especially regarding the two following open questions.

#### 1. order of selection algorithms
The order of the cover extraction methods might not be optimal.
The current order prioritizes the internal filename over the first image in the book.

I have an EPUB file in my library that is a book series with multiple internal images with cover in the name. As a result the cover is different based on the order of the Hash Map and therefore different every time the server is started.

On the other hand parsing XHTML is inherently slower than selecting the image from the filename alone.

#### 2. testing

I'm not sure exactly how to write tests for this code other than run it against my library of EPUB files.
Either it requires either real EPUB files (impossible to include in the repo), or constructed EPUB files (would require creating a lot of custom EPUB files to achieve coverage) for the tests.

## Ready?

- [x] I read the [[contributing guidelines](https://github.com/stumpapp/stump/blob/main/.github/CONTRIBUTING.md)](https://github.com/stumpapp/stump/blob/main/.github/CONTRIBUTING.md)
- [x] I disclosed any use of LLMs in the creation of this PR (if applicable)

- The [[expo application](https://github.com/stumpapp/stump/blob/main/apps/expo/LICENSE)](https://github.com/stumpapp/stump/blob/main/apps/expo/LICENSE) is licensed under [[GPL-3.0](https://www.gnu.org/licenses/gpl-3.0.html)](https://www.gnu.org/licenses/gpl-3.0.html)
- All other code in the repository is licensed under [[MIT License](https://www.tldrlegal.com/license/mit-license)](https://www.tldrlegal.com/license/mit-license)


core/src/filesystem/media/format/epub.rs
```diff
@@ -307,31 +324,17 @@ impl EpubProcessor {
...
-			return sorted.first().map(|(id, _)| id.to_string());
-		}
-
-		if let Some((_, id)) = search_result {
-			return Some(id.to_string());
-		}
-
-		None
+			// ignore images that do no contain cover in their name
+			.filter(|&(wheight, _)| wheight > 0)
```
aaronleopold, line 335:
```suggestion
			.filter(|&(weight, _)| weight > 0)
```

core/src/filesystem/media/format/epub.rs
```diff
@@ -363,7 +366,19 @@ impl EpubProcessor {
...
+		let id = Self::get_cover_id_by_resource_name(&resources_map);
+		if let Some(id) = id {
+			if let Some((buf, mime)) = epub_file.get_resource(id.as_str()) {
+				return Ok((ContentType::from(mime.as_str()), buf));
+			}
+		}
+
+		if let Some((mime, buf)) = Self::get_cover_by_reading_order(epub_file) {
+			return Ok((ContentType::from(mime.as_str()), buf));
+		}
```
aaronleopold, line 379: I think this is basically addressing your first question, but I'd say cover via reading order first since it is more deterministic. The performance implications are acceptable since it would really only be noticeable for folks not generating thumbs querying for the cover on a browser/device that has yet to cache the covers locally

core/src/filesystem/media/format/epub.rs
aaronleopold, line 394:
```suggestion
	/// metadata, it will go through four rounds of fallback methods:
```

core/src/filesystem/media/format/epub.rs
```diff
@@ -514,6 +531,92 @@ impl EpubProcessor {
...
+							["png", "jpg", "jpeg"]
+								.iter()
+								.any(|file_ending| a.ends_with(file_ending))
+						}) {
+						// Assamble full path to access resource.
+						//
+						// Manually normalize the path
+						//
+						// TODO: Replace with normalize_lexically
+						// https://doc.rust-lang.org/std/path/struct.Path.html#method.normalize_lexically
```
aaronleopold, line 590: thank you!

core/src/filesystem/media/format/epub.rs
```diff
@@ -514,6 +531,92 @@ impl EpubProcessor {
...
+			epub_file
+				.get_resource(&spine_item.idref)
+				.and_then(|(data, _mime)| {
+					if let Some(dir_path) = epub_file
+						.resources
+						.get(&spine_item.idref)
+						.and_then(|r| r.path.parent())
+						.map(|p| p.to_owned())
+					{
+						Self::find_image_in_xhtml(epub_file, dir_path, &data, &mut buf)
```
aaronleopold, line 551: I think `find_image_in_xhtml` should just have its own buffer and we can drop the one in this fn, otherwise if I am reading correctly it will just accumulate items from the previous spine item

core/src/filesystem/media/format/epub.rs
```diff
@@ -514,6 +531,92 @@ impl EpubProcessor {
...
+				quick_xml::events::Event::Empty(e) if e.name().as_ref() == b"img" => {
+					if let Some(img_path) = e
+						.try_get_attribute("src")
+						.ok()
+						.flatten()
+						.and_then(|a| String::from_utf8(a.value.to_vec()).ok())
+						.filter(|a| {
+							["png", "jpg", "jpeg"]
+								.iter()
+								.any(|file_ending| a.ends_with(file_ending))
```
aaronleopold, line 583: Maybe a `to_lowercase().ends_with()` here in case the ext on file are not lowercase?

aaronleopold (commented):
Hey, thanks for these improvements! I left a few comments, some of which should answer your questions, but I will directly address them here for clarity:

> 1. order of selection algorithms

I think prioritizing via the xhtml is my preference here, it should be negligible during scans and for folks that do not enable thumbnail generation it would impact initial loads but the browser/mobile app should cache things after that point.

> [2. testing] I'm not sure exactly how to write tests for this code other than run it against my library of EPUB files

It's not the most convenient module for testing. What exists today are mostly constructed files, and I would say a minimal one to cover a few scenarios would probably give a decent enough amount of coverage. I also imagine the upstream crate might have some testing strategies we can look at if needed

core/src/filesystem/media/format/epub.rs
hal8174: I was missing the buf.clear() at the end of the loop.
Now it is used the same way as in the example from quick-xml.

core/src/filesystem/media/format/epub.rs
hal8174: I noticed that this functionality is already implemented in the normalize_resource_path function which I am now using.

hal8174:
I think I now have resolved all the comments and remaining bugs.

aaronleopold (approved):
lgtm, thank you!

Issue #245, opened 2024-01-20
[FEATURE] UI customization options


aaronleopold:
**Is your feature request related to a problem? Please describe.**

A suite of options for customizing the UI on a per-user basis (i.e. manage my own preferences for what is displayed)

**Describe the solution you'd like**

The following are things I personally want for my own hosted server, if anyone else has any ideas/suggestions let me know. Otherwise, I'm satisfying my own needs only 😛 

- [ ] Rearrange order sidebar/topbar items
- [ ] Hide sidebar/topbar items
    - [ ] Include the ability to hide the shortcuts, e.g. create new library button
- [ ] Rearrange home page lists (e.g. recently read, recently added, etc.) and add additional content
    - [ ] I'd also like to add more here, this is one of the more dated parts of Stump and there are a lot more options available. I also think varying the type of content (i.e. list vs grid group vs table, etc) would be really visually good
- [ ] Support a paged, table layout for the explore books scene
    - [ ] After working with Stump's smart lists, I really enjoyed using a table for browsing my stuff. This is a large effort and will likely be separated to its own ticket, though.
- [ ] Compact display mode to reduce overall spacing
- [ ] Prefer colors vs prefer monochrome

devy1540:
I'd like to help with the home page customization mentioned here. In particular, I'd like to be able to hide sections such as “Recently added books” and “Recently added series,” and change the order of the sections I keep.

Would a first PR covering the following be welcome?

- Show or hide individual home page sections.
- Reorder the existing sections.
- Save these preferences per user, keeping the current layout as the default.

If this would be useful, I'd be happy to work on the implementation. Is anyone already working on this, or is there a preferred approach I should follow before starting?

aaronleopold:
> Would a first PR covering the following be welcome?
> 
> - Show or hide individual home page sections.
> - Reorder the existing sections.
> - Save these preferences per user, keeping the current layout as the default

Yes and nobody else is working on it to my knowledge. There are already some [models relevant to this](https://github.com/stumpapp/stump/blob/nightly/crates/models/src/shared/arrangement.rs), so you have something to work off of.


> is there a preferred approach I should follow before starting?

One thing I'll say is in the arrangements I had soft plans to remove that "locked" status, so if you can avoid that in the mutations that'd be preferred. I would also eventually want to hook this up to the mobile app, as well

PR #1392, opened 2026-09-06
✨ Add home page section preferences

devy1540:
## Description

Home page sections currently have a fixed order and cannot be hidden. This adds an editor under **Preferences → Home page** to show or hide the four existing sections, reorder them by dragging or keyboard, and save the layout per user. Reset restores the default order and visibility. When every section is hidden, the home page provides a link back to preferences.

This implements the home-section portion of #245.

- Match the navigation arrangement editor's styling, with eye-icon visibility controls and small Reset, Cancel, and Save buttons in the header. Save keeps its label while pending; failures use a short toast title and the error message as its description.
- Add `updateHomeArrangement` and apply the same normalization when reading saved home sections: keep the first supported occurrence of each section, preserve its order and visibility, and append missing sections as hidden.
- Use ordinary tagged Serde serialization. Unreadable arrangement JSON falls back to the corresponding default configuration; the next save writes the current format. Decoding happens at the arrangement-field boundary so a bad configuration does not prevent loading the user's preferences.
- Remove unused `links` fields from the built-in home section configurations.
- Render and prefetch enabled sections in the saved order. Successful saves update the query cache and cancel older in-flight reads; failed saves retain the draft for retry.

The shared server API can support a future mobile implementation. This PR adds the web editor and updates its documentation.

LLM tooling assisted with this contribution.

## Validation

- Browser tests: **274 passed**, including keyboard ordering, visibility, reset and invalid-configuration recovery, save failure/retry, rendering, prefetching, and stale-read handling.
- Model tests: **78 passed**, including tagged JSON round trips and section normalization.
- Server integration tests: **60 passed**. Home-specific cases cover saved order/visibility, normalization on reads and writes, and unreadable-config fallback followed by a successful save.
- Browser TypeScript checks, production web build, Rust formatting, Clippy for `models` and `graphql`, GraphQL schema consistency, and client generation passed.
- Browser ESLint: **0 errors**; 17 warnings remain in existing files outside this change.
- Revalidated in Chrome using a temporary local database and empty library: keyboard reordering, visibility, cancellation, saving and reload, all-hidden recovery, and Reset. Stopping the test server confirmed the error toast and draft retention; restarting it allowed the same draft to be saved.
- Rechecked after merging `nightly`: aligned the home test with `react-router` and added explicit hook return types to fix declaration generation.

## Ready?

- [x] This PR targets `nightly` and not `main`
- [x] I added tests and documentation for my changes
- [x] I disclosed the use of LLMs in this PR


apps/server/tests/graphql/home_arrangement.rs
```diff
@@ -0,0 +1,141 @@
...
+		.unwrap()
+		.unwrap();
+	// Exercise a real stored legacy preference whose lock status was true.
+	let legacy = json!({"locked": true, "sections": [
+		{"visible": true, "config": {"name": null, "links": []}},
+		{"visible": true, "config": {"entity": "BOOKS", "name": null, "links": []}},
+		{"visible": true, "config": {"entity": "SERIES", "name": null, "links": []}}
+	]});
+	// Set the raw JSON to test the persisted legacy representation, before decoding.
+	use sea_orm::{ConnectionTrait, DatabaseBackend, Statement};
```
aaronleopold, line 43: I prefer all imports to be at the top of the file

crates/graphql/src/mutation/user.rs
```diff
@@ -441,6 +443,29 @@ impl UserMutation {
 		Ok(User::from(updated_user))
 	}
 
+	/// Replace the authenticated user's home sections without a lock/unlock workflow.
```
aaronleopold, line 446:
```suggestion
	/// Replace the authenticated user's home sections
```

crates/models/src/shared/arrangement.rs
```diff
@@ -168,16 +168,76 @@ pub struct RecentlyAdded {
...
+pub struct OnDeckBooks {
+	name: Option<String>,
+	#[graphql(default)]
+	links: Vec<FilterableArrangementEntityLink>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Union, OneofObject)]
 #[graphql(input_name = "ArrangementConfigInput")]
-#[serde(untagged)]
+#[serde(tag = "type")]
```
aaronleopold, line 191:
```suggestion
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Union, OneofObject)]
#[graphql(input_name = "ArrangementConfigInput")]
#[serde(tag = "type")]
```

I'll explain more below

crates/models/src/shared/arrangement.rs
```diff
@@ -168,16 +168,76 @@ pub struct RecentlyAdded {
...
+		} else if value.get("filter").is_some() || value.get("order_by").is_some() {
+			serde_json::from_value(value).map(Self::Custom)
+		} else if value.get("entity").is_some() {
+			serde_json::from_value(value).map(Self::RecentlyAdded)
+		} else {
+			serde_json::from_value(value).map(Self::InProgressBooks)
+		}
+		.map_err(D::Error::custom)
+	}
+}
```
aaronleopold, line 239: This is not something I want to maintain honestly, mostly in the sense that we are talking about a feature that was partly developed and never used and so I don't think we need to consider "legacy" quite so seriously here.

If a stored config cannot be deserialized into what we are expected, I am inclined to instead just discard it and return the default config instead. I think for JSON columns this is mostly acceptable, could do some kind of versioned enum with Rust but that is way more headache than what this feature calls for imo. This way, if I ever make updates that break, it would be as simple as redefining what you want on the homepage and the next write fixes the bad data.

That would also mean we can delete:

- `legacy_home_order_and_visibility_survive_upgrade`
- `navigation_and_custom_legacy_configs_keep_their_identity`
- `invalid_config_tags_are_not_reinterpreted_as_legacy` can turn into e.g. `unknown_config_tag_errors` or something

crates/models/src/shared/arrangement.rs
```diff
@@ -168,16 +168,76 @@ pub struct RecentlyAdded {
...
+	Serialize,
+	Deserialize,
+	SimpleObject,
+	InputObject,
+)]
+#[graphql(input_name = "OnDeckBooksInput")]
+pub struct OnDeckBooks {
+	name: Option<String>,
+	#[graphql(default)]
+	links: Vec<FilterableArrangementEntityLink>,
```
aaronleopold, line 186: OnDeck does not need links, there are a few others that make similar mistake (e.g., preexisting `InProgressBooks`). I do recall, although it was really some time ago, adding them for a future where there is a `See all` button but I think until that point it is simpler to remove until needed

crates/models/src/shared/arrangement.rs
```diff
@@ -189,8 +249,54 @@ pub struct ArrangementSection {
...
+		if sections.len() != 4 {
+			return Err("Include each of the four home sections exactly once; use visible to hide a section");
+		}
+		let mut keys = std::collections::HashSet::new();
+		for section in &sections {
+			let key = section.home_key().ok_or("Unsupported home section")?;
+			if !keys.insert(key) {
+				return Err("Duplicate home section");
+			}
+		}
```
aaronleopold, line 285: If we are validating as part of new, I'd rather validate and fix instead of erroring. E.g., if we are missing sections then just add the missing ones as not visible?

Same would go for the duplicate or unsupported section, I'd say just filter them out

crates/models/src/shared/arrangement.rs
```diff
@@ -189,8 +249,54 @@ pub struct ArrangementSection {
...
+		Ok(Self { sections })
+	}
+}
+
+impl From<HomeArrangement> for Arrangement {
+	fn from(home: HomeArrangement) -> Self {
+		Self {
+			// Legacy storage/query field only; it is not part of the home
+			// mutation contract and does not gate editing.
```
aaronleopold, line 294:
```suggestion
			// TODO(arrangement): delete this field and usages
```

crates/models/src/shared/arrangement.rs
```diff
@@ -201,14 +307,39 @@ pub struct Arrangement {
...
+						.position(|item| item.home_key() == Some("continueReading"))
+						.map_or(self.sections.len(), |index| index + 1);
+					self.sections.insert(index, section);
+				} else {
+					self.sections.push(section);
+				}
+			}
+		}
+		self
+	}
```
aaronleopold, line 329: I think if you make the suggested changes to `HomeArrangement::new` then this would not be needed, right?

packages/browser/src/scenes/settings/app/preferences/HomeArrangementPreference.tsx
aaronleopold, line 1: The look and feel departs a bit from the existing navigation arrangement section. I'd rather align on the existing style:

| Your UI | Existing Example |
| ---------------- | ------- |
|    Your UI              |  Existing       |

You can change the lock icon to a small save/cancel button, instead (reset just saying `Reset` on the left in same header, maybe?). I'd also say maybe all the buttons should be `ghost` variant, though, I'll clean up the other section in a separate branch for that.

packages/i18n/src/locales/en-US.json
```diff
@@ -1704,6 +1706,19 @@
...
+				"homeArrangement": {
+					"title": "Home page",
+					"description": "Choose which sections appear on your home page and their order.",
+					"label": "Home sections",
+					"hint": "Drag sections or use the arrow buttons to reorder them. Changes apply to your account after saving.",
+					"reorder": "Reorder {{section}}",
+					"moveUp": "Move {{section}} up",
+					"moveDown": "Move {{section}} down",
+					"reset": "Restore defaults",
+					"saving": "Saving…",
```
aaronleopold, line 1719: I think we can avoid changing the button text to `Saving...` and instead either just disable it (with the `Save` text) or add a loader

packages/i18n/src/locales/en-US.json
```diff
@@ -1704,6 +1706,19 @@
 				"readingSessionGracePeriod": {
 					"label": "Reading session grace period",
 					"description": "If no activity is detected within this time frame, in seconds, the next update will start a new session"
+				},
+				"homeArrangement": {
+					"title": "Home page",
+					"description": "Choose which sections appear on your home page and their order.",
```
aaronleopold, line 1712:
```suggestion
					"description": "Adjust the sections on your home page",
```

packages/i18n/src/locales/en-US.json
```diff
@@ -1704,6 +1706,19 @@
...
+					"label": "Home sections",
+					"hint": "Drag sections or use the arrow buttons to reorder them. Changes apply to your account after saving.",
```
aaronleopold, line 1714:
```suggestion
					"hint": "Drag and rearrange sections or hide them entirely",
```

packages/i18n/src/locales/en-US.json
```diff
@@ -1704,6 +1706,19 @@
...
+					"reorder": "Reorder {{section}}",
+					"moveUp": "Move {{section}} up",
+					"moveDown": "Move {{section}} down",
```
aaronleopold, line 1717: I think these go away per my suggestions earlier

packages/i18n/src/locales/en-US.json
```diff
@@ -1704,6 +1706,19 @@
...
+					"reset": "Restore defaults",
+					"saving": "Saving…",
+					"saveFailed": "Could not save your home page. Your changes are still here; please try again.",
+					"unsupported": "This home configuration contains sections that this editor does not support."
```
aaronleopold, line 1721:
```suggestion
					"saveFailed": "Failed to save your changes",
					"unsupported": "Invalid configuration",
					"unsupportedDescription": "Please reset to defaults and try again",
```

A few notes here:

1. `saveFailed` imo is a title and should be brief, then the error message should be the toast description
2. Same for `unsupported`, but added a semantic description instead of using an error message

crates/graphql/src/object/user_preferences.rs
```diff
@@ -22,6 +22,7 @@ impl UserPreferences {
 			.home_arrangement
 			.clone()
 			.unwrap_or(Arrangement::default_home())
+			.with_missing_home_sections()
```
aaronleopold, line 25: I mentioned removing this earlier, so perhaps all existing ones still need to flow through `new()` or something in case they need cleaning

apps/server/tests/graphql/home_arrangement.rs
```diff
@@ -0,0 +1,141 @@
...
+		.one(app.conn())
+		.await
+		.unwrap()
+		.unwrap();
+	// Exercise a real stored legacy preference whose lock status was true.
+	let legacy = json!({"locked": true, "sections": [
+		{"visible": true, "config": {"name": null, "links": []}},
+		{"visible": true, "config": {"entity": "BOOKS", "name": null, "links": []}},
+		{"visible": true, "config": {"entity": "SERIES", "name": null, "links": []}}
+	]});
```
aaronleopold, line 41: We can remove this and any other legacy tests per previous feedback, as well

apps/server/tests/graphql/home_arrangement.rs
```diff
@@ -0,0 +1,141 @@
...
+		.one(app.conn())
+		.await
+		.unwrap()
+		.unwrap();
+	assert_eq!(stored.locale, preferences.locale);
+	assert_eq!(stored.app_theme, preferences.app_theme);
+}
+
+#[tokio::test]
+async fn home_arrangement_rejects_invalid_and_unauthenticated_writes() {
```
aaronleopold, line 80: This is really two tests, but also basic auth vs unauth in general should be handled in a separate test module imo. We can omit it from these, at the very least.

apps/server/tests/graphql/home_arrangement.rs
```diff
@@ -0,0 +1,141 @@
...
+		.await;
+	assert!(
+		!response.status_code().is_success()
+			|| response.json::<Value>().get("errors").is_some()
+	);
+	assert_eq!(app.execute_gql(QUERY, None).await, before);
+}
+
+#[tokio::test]
+async fn home_arrangement_is_scoped_to_the_authenticated_user_and_can_hide_every_section()
```
aaronleopold, line 106: This test was a little hard to follow, why did we create a secondary user here? I think it would be enough to assert the default user created as part of `TestApp` can update the sections. If you're aiming to add assertions that the `me` node in GraphQL only returns for the logged-in user, that would not be the domain of this test imo

apps/server/tests/graphql/home_arrangement.rs
```diff
@@ -0,0 +1,141 @@
...
+	])
+}
+
+async fn update(app: &TestApp, sections: Value) -> Value {
+	app.execute_gql(MUTATION, Some(json!({"input": {"sections": sections}})))
+		.await
+}
+
+#[tokio::test]
+async fn home_arrangement_persists_without_unlocking_and_leaves_other_preferences_alone()
```
aaronleopold, line 24: I don't think this test is needed for two reasons:

1. We don't need to assert locking/unlocking because it is explicitly not developed for it, and it is being removed for others eventually
2. We already have guarantees it will leave the other preferences alone via sea-orm and don't think we need to assert behavior owned by that library

aaronleopold (commented):
Thank you for working on this, I reviewed the changes this morning and have some feedback

devy1540:
Thanks for the detailed review, @aaronleopold! I've pushed updates addressing your feedback and merged the latest nightly changes. Whenever you have a chance, I'd appreciate another look. Happy to make any further adjustments!

crates/models/src/entity/user_preferences.rs
```diff
@@ -53,11 +53,11 @@ pub struct Model {
 	#[graphql(skip)]
 	#[sea_orm(column_type = "Json", nullable)]
 	#[serde(default = "Model::default_navigation_arrangement")]
-	pub navigation_arrangement: Option<Arrangement>,
+	pub navigation_arrangement: Option<Json>,
 	#[sea_orm(column_type = "Json", nullable)]
 	#[graphql(skip)]
 	#[serde(default = "Model::default_home_arrangement")]
-	pub home_arrangement: Option<Arrangement>,
+	pub home_arrangement: Option<Json>,
```
aaronleopold: Not sure I like this one, instead we could maybe manually implement `FromJsonQueryResult` with a safer parse? Otherwise this cuts a bit from a dx-perspective. I think a lot of the changes in this file go away if we do that, as well

aaronleopold (commented):
Thanks for making those updates! I have one remaining point, but otherwise I think this looks good

crates/models/src/entity/user_preferences.rs
devy1540: Thanks for the suggestion! Addressed in 9274ad72.

Both fields are back to `Option<Arrangement>`. The manual SeaORM JSON conversion treats unreadable stored configurations as absent, allowing home and navigation to use their respective defaults without losing other preferences. This removes the explicit JSON conversions from callers while preserving home section normalization.

SQLite tests cover valid and invalid configurations, preservation of other preferences, and recovery on the next save without modifying stored data during reads. Model, GraphQL, and server tests passed, along with formatting and schema checks.

The broader all-target Clippy check reports four existing lint errors on local Rust 1.98; I confirmed the same errors on the unchanged PR head.

aaronleopold (approved):
Thank you

Issue #682, opened 2025-07-26
[BUG] Reader text size & linespacing button


Silther:
## Description of the Bug

The buttons don't register a single short press but a continues press, this may be a good idea, but currently it is way to sensitive, clicking on a touchpad registers at least as 2-3 presses, so you have no granular control over it.

## Expected Behavior

Register one single press independent of the duration you keep it pressed or at least wait 1-2 seconds before registering new presses.

## To Reproduce

1. Open A book
2. Click on the customise icon
3. try changing the line spacing or text font size.

## Screenshots

https://github.com/user-attachments/assets/ff4d3c9e-1dff-416a-b6f0-76fd9493bfb2

## Debugging Information

Please provide any relevant debugging information that you have, e.g. logs, error messages, etc, and any additional context in the following sections.

**Environment:**
- Brower: vivaldi on windows

**Build Details:**

- Version: [0.11, stable]
- Docker: [yes]

**Additional context**

PR #1393, opened 2026-09-06
:adhesive_bandage: Fix repeated EPUB font size and line height adjustments

ajselzilic:
Closes https://github.com/stumpapp/stump/issues/682

## Description

A 200 ms press on the EPUB font-size or line-height buttons could apply three changes because the hold handler repeated every 100 ms. Use native `onClick` handlers so each completed mouse or touch press applies one step, with normal Enter and Space activation.

Remove the unused `usePressAndHold` hook and add accessible names to all four buttons. Existing limits, rounding, preview behavior and saved preferences are unchanged.

LLM disclosure: Codex was used to implement the fix, review the code and draft this description.

## Validation

- `yarn workspace @stump/browser check-types`: passed.
- ESLint on both controls: passed.
- Prettier check on both controls: passed.
- `yarn workspace @stump/web build`: passed.

Manual verification with a live EPUB and the reporter's Windows/Vivaldi touchpad remains pending.

## Ready?

- [x] I disclosed any use of LLMs in the creation of this PR (if applicable)


packages/browser/src/components/readers/epub/controls/FontSizeControl.tsx
```diff
@@ -65,12 +59,10 @@ export default function FontSizeControl() {
 					{localFontSize}px
 				</span>
```
aaronleopold, line 60: Eventually I think the `style={{ fontSize: `${displayedFontSize}px` }}` should also be removed so the controls don't shift away from the cursor, not part of this change but just making a note

aaronleopold (approved):
Thank you!

PR #1404, opened 2026-09-10
:recycle: Improve web and mobile readium interop

aaronleopold:
## Description

This PR aims to improve some of the issues reported when moving between web and mobile ebook readers. It definitely doesn't remove all the rough edges, and I have some more testing I'd like to get done, but this should improve things a bit: 

- Store package-relative urls for locator `href`s instead of server urls
   - The web app was storing full urls while mobile was storing relative urls, which caused the locators to not route correctly between platforms (e.g., reading on web, jump to mobile, it kicks you to the start of the book)
   - This should also improve issues with highlights/annotations not "working" between web and mobile platforms
- Refactor the positions generation logic to chunk pages per the RMSDK / Readium standard (a resource has roughly `compressed_size / 1024` positions)
   - This brings page counts into alignment between platforms, but does introduce what I found to be a slightly awkward UX on web footer that would ideally be corrected by showing the page number for each rendered "page" instead of one `X/Y` in the bottom-right
- Add a `position_aware_page` resolver to resume reading cursors, with a note to persist positions as part of ingestion at some point down the road

### DIFF
--- a/apps/expo/app/stump/[serverId]/books/[bookId]/index.tsx
+++ b/apps/expo/app/stump/[serverId]/books/[bookId]/index.tsx
@@ -180,6 +180,7 @@ export default function Screen() {
 		})
 	}, [downloadBook, book])
 
+	const insets = useSafeAreaInsets()
 	const router = useRouter()
 	const thumbnailRatio = usePreferencesStore((state) => state.thumbnailRatio)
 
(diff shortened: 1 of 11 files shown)


aaronleopold:
I've tested this on Android and while it is far from perfect, it does not throw me back to the beginning when switching between web/mobile so I think net improved-enough to merge. I still think there are UX problems for web in the footer, and you don't _always_ land where you'd want when switching to mobile (because the "end" of the current location on web is the anchor-point for re-entry, when probably we want the start?) but I think this just needs more time to be improved.

PR #1419, opened 2026-09-14
🐛 Fix ReadMore description flicker/re-render on expand

balazs-szucs:
## Description

Minor UI thing: books with a Summary flicker/re-render when you hit Read more. Generally more noticeable on long summaries, but can happen on smaller ones. (see examples below.). Not all summaries are effected (in my experience), so these examples were chosen based on the fact the bug more visible there. Nothing major though.

It was rendering the whole description again on toggle while also animating the box open, so I just left the text alone and let it open fully without the animation.

This seems to be a good solutions, though, i am assuming removing the MAX_EXPANDED_HEIGHT may backfire in some cases, so if you have edge case to test for that lmk. But for me even on longer summaries this seemed OK.

## Screenshots

Before:

https://github.com/user-attachments/assets/b9e3c792-f808-4d41-b9a2-7f58bf64c76d

https://github.com/user-attachments/assets/a329ffc3-642e-4634-883e-89d2a92745c1

After:

https://github.com/user-attachments/assets/f8ae3e42-2381-4e41-b267-99b18f29c19e

Screenshots or a screen recording of the visual changes associated with this PR.

(Feel free to delete this section for non-visual changes, or changes which do not benefit from some kind of visual demonstration)

## Ready?

- [X] I read the [contributing guidelines](https://github.com/stumpapp/stump/blob/main/.github/CONTRIBUTING.md)
- [X] I searched for existing issues or pull requests that may be related to my contribution
- [X] This PR is based into `nightly` and not `main`
- [X] I added tests and/or documentation for my changes if applicable
- [X] I disclosed any use of LLMs in the creation of this PR (if applicable)


packages/browser/src/components/ReadMore.tsx
```diff
@@ -9,43 +10,60 @@ type Props = {
...
+		return content
 	}
 
 	return (
 		<div>
 			<div
-				className={showingAll ? 'overflow-y-auto' : 'overflow-hidden'}
-				style={{
-					maxHeight: showingAll ? MAX_EXPANDED_HEIGHT : COLLAPSED_HEIGHT,
-					transition: 'max-height 0.3s cubic-bezier(0.4, 0, 0.2, 1)',
```
aaronleopold, line 34: Did you try to re-add the transition with this fix? Without it, it just snaps into place instead of transitioning to the `maxHeight` change

balazs-szucs: Yes, but sadly it is that transition that gives it the jittery feeling. So, re-adding it, essentially revert it back to the original state. Perhaps, i am out of my depth, but for me this works (as obviously it removes the whole animation) but re-adding even if i tweak it, just leads back to the original problem. Essentially the way it went, is i tried tweak original animation, no dice, then just removed it.

aaronleopold: Ah, well that's okay then

aaronleopold (approved):
Thank you!!

